PR #7, opened 2025-02-14
fix: resolve cli usage with grpc stubs

ramizpolic:
Overrides simple instead of autogenerated string values for enums on locators and digest types

Signed-off-by: Ramiz Polic <rpolic@cisco.com>


paralta (approved):
LGTM 👍

PR #12, opened 2025-02-17
refactor: digest value type to string

paralta:
* Simplify digest usage by changing the value from `byte` to `string`


api/core/v1alpha1/digest.proto
```diff
@@ -8,7 +8,7 @@ package core.v1alpha1;
 // A digest of content or globally unique content identifier
 message Digest {
   DigestType type = 1;
-  bytes value = 2;
+  string value = 2;
```
ramizpolic, line 15: please add a comment here how the value is constructed and an example (e.g. we expect a hex string here for the digest value)

api/core/v1alpha1/digest.go
```diff
@@ -22,7 +21,7 @@ func init() {
 }
 
 func (d *Digest) ToString() string {
```
ramizpolic: maybe we rename `ToString -> Encode`, and `FromString -> Decode` which will construct the digest into `<digest-type>:<digest-value>` string and read from it

server/store/oci/oci.go
```diff
@@ -54,7 +52,7 @@ func New(config Config) (types.StoreService, error) {
 }
 
 func (s *store) Lookup(ctx context.Context, digest *coretypes.Digest) (*coretypes.ObjectMeta, error) {
-	_, metaReader, err := s.repository.Blobs().FetchReference(ctx, ocidigest.NewDigestFromBytes(ocidigest.SHA256, digest.Value).String())
+	_, metaReader, err := s.repository.Blobs().FetchReference(ctx, digest.ToString())
```
ramizpolic: i would like to prevent accidental changes to what OCI expects here, so id suggest we create another method that will return oci specific digest, and we can call `String()` method on it, for example

```go
// converts directory-specific digest into OCI digest
func ociDigest(ourDigest *core.Digest) *digest.Digest {
 // implement me
}
```
we can even move this method to the original object, but i think it may be better to keep it close to the storage implementation as not to blow up digest conversion

```go
func (digest *Digest) ToOCI() *oci.digest {
 // implement me
}
```

then on the consumer side
```suggestion
	_, metaReader, err := s.repository.Blobs().FetchReference(ctx, ociDigest(ourDigest).String())
```

ramizpolic: note that there are instances where we cannot convert our own digest to the digest allowed by OCI (e.g. multihash), so we may want to also return an error during conversion

ramizpolic (changes requested):

api/core/v1alpha1/digest.go
paralta: completely agree!

api/core/v1alpha1/digest.go
```diff
@@ -21,27 +20,21 @@ func init() {
...
 	}
 
-	// extract signature
-	digestValue, err := hex.DecodeString(parts[1])
-	if err != nil {
-		return fmt.Errorf("failed to decode digest: %w", err)
-	}
-
 	// update digest
 	*d = Digest{
```
ramizpolic, line 42: i would still validate the digest value itself with the following:
- is the value read a valid HEX string
- is the value length valid for a given digest type (e.g. sha256 should have 64 chars)

ramizpolic (approved):
lgtm

PR #13, opened 2025-02-18
refactor: small fixes on cli

adamtagscherer:
This PR adds the output format flags to the pull cli command and changes the `fmt.Fprint()` calls to `cmd.Print` calls.

### DIFF
--- a/cli/cmd/pull/options.go
+++ b/cli/cmd/pull/options.go
@@ -7,11 +7,11 @@ var opts = &options{}
 
 type options struct {
 	AgentDigest string
-	OutputFormat      string
+	JSON        bool
 }
 
 func init() {
 	flags := Command.Flags()
 	flags.StringVar(&opts.AgentDigest, "digest", "", "Digest of the agent to pull")
-	Command.Flags().StringVarP(&opts.OutputFormat, "output", "o", "json", "Output format (json|yaml)")
+	flags.BoolVar(&opts.JSON, "json", false, "Output in JSON format")
 }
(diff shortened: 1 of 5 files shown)


paralta (approved):

ramizpolic:
@adamtagscherer can you remove the other fmt.Fprint usages from CLI, there are still some not cleared up

cli/cmd/pull/options.go
```diff
@@ -7,9 +7,11 @@ var opts = &options{}
 
 type options struct {
 	AgentDigest string
+	OutputFormat      string
 }
 
 func init() {
 	flags := Command.Flags()
 	flags.StringVar(&opts.AgentDigest, "digest", "", "Digest of the agent to pull")
+	Command.Flags().StringVarP(&opts.OutputFormat, "output", "o", "json", "Output format (json|yaml)")
```
ramizpolic: lets not allow yaml outputs because we have to support then the marshallign from yaml on input (will break).
lets just use `--json` binary flag that will only print out json result when used `dirctl pull --digest <digest> --json`

cli/cmd/push/push.go
```diff
@@ -28,7 +28,7 @@ var Command = &cobra.Command{
 	dirctl build <args> | dirctl push
 
 	# From pull
-	dirctl pull --id agent-id | dirctl push
+	dirctl pull --digest sha256:1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef | dirctl push
```
ramizpolic:
```suggestion
	dirctl pull --digest <digest-string> --json | dirctl push
```

PR #20, opened 2025-02-19
feat: allow multiple agent artifacts during build

paralta:
* Allow multiple artifacts during build with the format `type:url`
* Tested with e2e test ✅


README.md
```diff
@@ -117,8 +117,7 @@ dirctl build [options] <path-to-agent>
 Options:
 - `--name` : Name of the agent.
 - `--version` : Version of the agent.
-- `--artifact-url` : URL of the agent artifact.
-- `--artifact-type` : Type of the agent artifact.
+- `--artifact` : Artifacts to set for the agent. Each artifact should be in the format `type:url`. Example usage: `--artifact type1:url1 --artifact type2:url2`.
```
ramizpolic: Can we just have a description of allowed types, for example:
```suggestion
- `--artifact` : Artifacts to set for the agent. Each artifact should be in the format `type:url`. Example usage: `--artifact type1:url1 --artifact type2:url2`. Supported types: `docker-image`, `python-package`.
```

paralta: neat 👍  will just add `helm-chart`

ramizpolic: we also need one more locator type to be defined `source-code` for pointing to the source code

ramizpolic:
```proto
enum LocatorType {
  LOCATOR_TYPE_UNSPECIFIED = 0;
  LOCATOR_TYPE_HELM_CHART = 1;
  LOCATOR_TYPE_DOCKER_IMAGE = 2;
  LOCATOR_TYPE_PYTHON_PACKAGE = 3;
  LOCATOR_TYPE_SOURCE_CODE = 4;
}
```

also update the mapping here with `source-code`: https://github.com/agntcy/dir/blob/1419ecc3ef4f497a59bd77ea95575076a6ee0412/api/core/v1alpha1/locator.go#L6

paralta: @ramizpolic should we also add `binary`?

cli/cmd/build/options.go
```diff
@@ -47,4 +41,12 @@ func init() {
 	// Creation time (only for dev purposes)
 	flags.StringVar(&opts.CreatedAt, "created-at", "", "Agent creation time in RFC3339 format")
 	_ = flags.MarkHidden("created-at")
+
+	// Artifacts
+	flags.StringSliceVar(
+		&opts.Artifacts,
+		"artifact",
+		[]string{},
+		"Artifacts to set for the agent. Each artifact should be in the format 'type:url'. Example usage: --artifact type1:url1 --artifact type2:url2",
```
ramizpolic: the docs for allowed types should also be available here

paralta: nice catch, updated

ramizpolic (approved):
lgtm

PR #23, opened 2025-02-20
feat: custom extensions from build config file

paralta:
* Improve e2e test to check build output
* Refactor build config logic and remove `created-at` field
* Implement custom extensions in agent model from build config file
* E2E test extended to cover the latter ✅

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,6 @@ Options:
 - `--artifact` : Artifacts to set for the agent. Each artifact should be in the format `type:url`. Example usage: `--artifact type1:url1 --artifact type2:url2`. Supported types: `docker-image`, `python-package`, `helm-chart`, `source-code` and `binary`.
 - `--author` : Authors to set for the agent. Overrides builder defaults. Example usage: `--author author1 --author author2`.
 - `--category` : Categories to set for the agent. Overrides builder defaults. Example usage: `--category category1 --category category2`.
-- `--created-at` : Agent creation time in RFC3339 format.
 - `--llmanalyzer` : Enable LLMAnalyzer extension.
 - `--config-file` : Path to the agent build configuration file. Please note that other flags will override the build configuration from the file. Supported formats: YAML. Example template: cli/cmd/build/config/build.config.yaml.
 
--- a/cli/cmd/build/config/config.go
+++ b/cli/cmd/build/config/config.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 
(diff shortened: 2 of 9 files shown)


cli/cmd/build/options.go
```diff
@@ -39,10 +39,6 @@ func init() {
 		"Categories to set for the agent. Overrides builder defaults. Example usage: --category category1 --category category2",
 	)
 
-	// Creation time (only for dev purposes)
-	flags.StringVar(&opts.CreatedAt, "created-at", "", "Agent creation time in RFC3339 format")
-	_ = flags.MarkHidden("created-at")
-
```
ramizpolic: this needs to keep for now as CSIT is using this flag to avoid creation time settings

cli/cmd/build/config/config.go
```diff
@@ -15,13 +16,49 @@ type Artifact struct {
 	URL  string `yaml:"url"`
 }
 
+type Extension struct {
+	Name    string            `yaml:"name"`
+	Version string            `yaml:"version"`
+	Specs   map[string]string `yaml:"specs"`
+}
+
 type Config struct {
```
ramizpolic, line 26: createAt can be specified here but it doesnt have to be documented in the example build config

cli/cmd/build/config/config.go
```diff
@@ -15,13 +16,49 @@ type Artifact struct {
 	URL  string `yaml:"url"`
 }
 
+type Extension struct {
+	Name    string            `yaml:"name"`
+	Version string            `yaml:"version"`
+	Specs   map[string]string `yaml:"specs"`
```
ramizpolic: we load in a generic struct

```suggestion
	Specs   any `yaml:"specs"`
```

README.md
```diff
@@ -120,7 +120,6 @@ Options:
 - `--artifact` : Artifacts to set for the agent. Each artifact should be in the format `type:url`. Example usage: `--artifact type1:url1 --artifact type2:url2`. Supported types: `docker-image`, `python-package`, `helm-chart`, `source-code` and `binary`.
 - `--author` : Authors to set for the agent. Overrides builder defaults. Example usage: `--author author1 --author author2`.
 - `--category` : Categories to set for the agent. Overrides builder defaults. Example usage: `--category category1 --category category2`.
-- `--created-at` : Agent creation time in RFC3339 format.
```
ramizpolic, line 123: this is okay to remove from documentation, we dont have to revert it

cli/cmd/build/config/config.go
```diff
@@ -15,13 +16,49 @@ type Artifact struct {
...
-	Authors     []string   `yaml:"authors"`
-	Categories  []string   `yaml:"categories"`
-	Artifacts   []Artifact `yaml:"artifacts"`
+	Source      string      `yaml:"source"`
+	Name        string      `yaml:"name"`
+	Version     string      `yaml:"version"`
+	LLMAnalyzer bool        `yaml:"llmanalyzer"`
+	Authors     []string    `yaml:"authors"`
+	Categories  []string    `yaml:"categories"`
+	Artifacts   []Artifact  `yaml:"artifacts"`
```
ramizpolic, line 34: we should build config artifacts type object to include metadata + digest string

ramizpolic: digest should be a string we decode on our end into the object

ramizpolic (changes requested):

cli/cmd/build/config/config.go
paralta: let's update this separately as it falls outside the scope of this PR. also we need to define if this `annotations`/`metadata` field will be allowed when we load config from flags

paralta:
@ramizpolic thanks for reviewing! just pushed some changes, please review when you have time

PR #26, opened 2025-02-24
fix: cobra command print

adamtagscherer:
This PR fixes to print the result of commands to the `os.Stdout` instead of the `os.Stderr.`

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -6,7 +6,6 @@ package cmd
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/agntcy/dir/cli/cmd/build"
 	"github.com/agntcy/dir/cli/cmd/pull"


cli/cmd/root.go
```diff
@@ -51,6 +55,6 @@ func Run(ctx context.Context) error {
 		return err
 	}
 
-	_, _ = fmt.Fprintf(RootCmd.OutOrStdout(), "\n")
+	RootCmd.Print("\n")
```
ramizpolic: we can remove this

cli/cmd/root.go
```diff
@@ -31,6 +33,8 @@ var RootCmd = &cobra.Command{
 		ctx := util.SetClientForContext(cmd.Context(), c)
 		cmd.SetContext(ctx)
 
+		cmd.SetOut(os.Stdout)
```
ramizpolic: this will break the piping. i suggest we introduce a presenter for the CLI to help with printing options

```go
// defined in cli/presenter/printer.go
func Print(cmd cobra.Cmd, args...interface{}) {
	_, _ = fmt.Fprint(cmd.OutOrDefault(), args...)
}

func Printf(cmd cobra.Cmd, format string, args...) {
	_, _ = fmt.Fprintf(cmd.OutOrDefault(), format, args...)
}

func Error(cmd cobra.Cmd, args...interface{}) {
	_, _ = fmt.Fprint(cmd.ErrOrStderr(), args...)
}

func Errorf(cmd cobra.Cmd, format string, args...) {
	_, _ = fmt.Fprintf(cmd.ErrOrStderr(), format, args...)
}
```
please check the method signature and align. this package becomes a central point once we decide to introduce a more complete presentation logic

ramizpolic: we also avoid the overrides in e2e tests with pipe manipulation

Issue #21, opened 2025-02-19
[Issue]: The Build Command Incorrectly Parses Files from Internal Directories


shanchunyang0919:
### Bug Description

go run cli.go build --name=alfred --version="v1.0.0" --author="alfred-team" --artifact-url="http://ghcr.io/cisco-agents/alfred" --artifact-type="docker-image" --llmanalyzer ../../alfred > alfred.json
(I added debug logs)
```
2025/02/19 13:38:40 Parsing file: ../../alfred/.github/workflows/cd.yaml
2025/02/19 13:38:40 Parsing file: ../../alfred/.github/workflows/ci.yaml
2025/02/19 13:38:40 Parsing file: ../../alfred/.venv/lib/python3.12/site-packages/torch/_export/serde/schema.yaml
2025/02/19 13:38:40 Parsing file: ../../alfred/.venv/lib/python3.12/site-packages/torchgen/packaged/ATen/native/native_functions.yaml
Error: failed to build agent: failed to process file tree: failed to parse file: yaml: unmarshal errors:
  line 9: cannot unmarshal !!seq into map[string]interface {}
```

### Expected Behavior

The parser should skip yaml files in internal directories like `/.venv`.

### Affected Version

v1.0.3

### Steps to Reproduce

Set the source path to repo/dir that contains internal directories

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

ramizpolic:
Thanks for reporting @shanchunyang0919! We will add skip checks

PR #27, opened 2025-02-24
feat: add ignore paths config in build cmd 

paralta:
* Rename `artifact` to `locator` in `build` cmd
* Remove `created at` from configuration
* Remove `category` extension. Category information should be "injected" as an extension in the configuration file
* Add new configuration to ignore certain paths when running the builder. Fixes https://github.com/agntcy/dir/issues/21

### DIFF
--- a/cli/builder/builder.go
+++ b/cli/builder/builder.go
@@ -7,21 +7,22 @@ import (
 	"context"
 
 	apicore "github.com/agntcy/dir/api/core/v1alpha1"
+	"github.com/agntcy/dir/cli/builder/config"
 	"github.com/agntcy/dir/cli/builder/extensions/crewai"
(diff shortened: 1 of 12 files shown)


cli/builder/builder.go
```diff
@@ -34,21 +31,5 @@ func Build(ctx context.Context, cfg *config.Config) ([]*apicore.Extension, error
 		return nil, err
 	}
 
-	// Append config extensions
-	for _, ext := range cfg.Extensions {
```
ramizpolic, line 38: why was this removed?

paralta: it was moved to `build.go`, so this `Extensions ` object belongs to the build config

PR #33, opened 2025-03-12
refactor: runtime extension

adamtagscherer:
This PR refactors the framework and language runtime extensions.

### DIFF
--- a/cli/builder/plugins/runtime/framework/framework.go
+++ b/cli/builder/plugins/runtime/framework/framework.go
@@ -20,10 +20,10 @@ const (
 type Type string
 
 const (
-	CrewAI    Type = "crewai"
-	Autogen   Type = "autogen"
+	CrewAI     Type = "crewai"
+	Autogen    Type = "autogen"
 	Llamaindex Type = "llama-index"
-	Langchain Type = "langchain"
+	Langchain  Type = "langchain"
 )
 
 type ExtensionSpecs struct {


cli/types/types.go
```diff
@@ -47,5 +47,5 @@ func (a *AgentExtension) ToAPIExtension() (apicore.Extension, error) {
 }
 
 type ExtensionBuilder interface {
```
paralta:
```suggestion
type Builder interface {
```
what do you think about having just `Builder` here? to avoid any `plugin` vs `extension` confusion

e2e/testdata/build.config.yaml
```diff
@@ -50,5 +50,23 @@ builder:
   # Enable or disable the CrewAI extension
   crewai: false
 
-  # Enable or disable the runtime extension
-  runtime: false
+  # Runtime plugin configuration
+  runtime:
+    # Enable or disable the runtime plugin
+    enabled: false
```
paralta: would be good to enable this and remove the `language ` config, then we could verify if the `language` in the agent model is retrieved from the `runtime` plugin

cli/cmd/build/build.go
```diff
@@ -42,19 +42,24 @@ func runCommand(cmd *cobra.Command) error {
...
+	err = builderInstance.RegisterPlugins()
 	if err != nil {
-		return fmt.Errorf("failed to register extensions: %w", err)
+		return fmt.Errorf("failed to register plugins: %w", err)
 	}
-	extensions, err := manager.Run(cmd.Context())
+	extensions, err := builderInstance.Run(cmd.Context())
 	if err != nil {
-		return fmt.Errorf("failed to run extension manager: %w", err)
+		return fmt.Errorf("failed to build extensions: %w", err)
```
paralta:
```suggestion
		return fmt.Errorf("failed to build plugins: %w", err)
```

cli/builder/config/config.go
```diff
@@ -84,3 +92,17 @@ func (c *Config) GetAPILocators() ([]*apicore.Locator, error) {
...
+		return fmt.Errorf("failed to validate framework: %w", err)
+	}
+
+	err = c.Builder.Runtime.Language.Validate()
+	if err != nil {
+		return fmt.Errorf("failed to validate language: %w", err)
+	}
+
+	return nil
+}
```
paralta: We can only do this check when the `runtime` plugin is disabled. otherwise we will not be able to get the `language` and `framework` through the `runtime` plugin

adamtagscherer: it should be enabled, rather than disabled, right?

paralta: solved by agent merge ✅

cli/builder/plugins/runtime/language/language.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, fmt.Errorf("failed to get SBOM: %w", err)
+	}
+
+	return &types.AgentExtension{
+		Name:    ExtensionName,
+		Version: ExtensionVersion,
+		Specs: ExtensionSpecs{
+			Type:    runtimeInfo.Language,
+			Version: runtimeInfo.Version,
+			SBOM:    sbom,
```
paralta: I think SBOM should not be included in the language extension

cli/builder/plugins/runtime/framework/framework.go
```diff
@@ -0,0 +1,46 @@
...
+	return &Framework{}
+}
+
+func (fw *Framework) Build(_ context.Context) (*types.AgentExtension, error) {
+	return &types.AgentExtension{
+		Name:    ExtensionName,
+		Version: ExtensionVersion,
+		Specs: ExtensionSpecs{
+			Type:    CrewAI,
+			Version: "0.102.0",
```
paralta: why is this hardcoded? it should come from the SBOM scanning that finds the `crewai` package

cli/builder/plugins/llmanalyzer/llmanalyzer.go
```diff
@@ -18,10 +18,8 @@ import (
 )
 
 const (
-	// ExtensionName is the name of the extension.
 	ExtensionName = "llmanalyzer"
```
paralta:
```suggestion
	PluginName = "llmanalyzer"
```

cli/builder/builder.go
```diff
@@ -6,79 +6,104 @@ package builder
...
 	}
 
-	for _, i := range em.cfg.Model.Extensions {
+	return &apicore.Agent{
+		Name:       b.cfg.Model.Name,
+		Version:    b.cfg.Model.Version,
+		Authors:    b.cfg.Model.Authors,
+		CreatedAt:  timestamppb.New(time.Now()),
+		Skills:     b.cfg.Model.Skills,
+		Locators:   locators,
```
paralta, line 82: As we discussed previously, these details can be empty because this is the "Plugin/Builder Agent Model" not the "User Agent Model"

cli/cmd/build/build.go
```diff
@@ -43,33 +40,24 @@ func runCommand(cmd *cobra.Command) error {
...
-	agent := &apicore.Agent{
-		Name:       cfg.Model.Name,
-		Version:    cfg.Model.Version,
-		Authors:    cfg.Model.Authors,
-		CreatedAt:  timestamppb.New(time.Now()),
-		Skills:     cfg.Model.Skills,
-		Locators:   locators,
-		Extensions: extensions,
-	}
+	agent.Merge(userAgent)
```
paralta: I think the convention here should be `mainAgent.Merge(enricherAgent)` aka `userAgent.Merge(builderAgent)`, why did you went the this way?

api/core/v1alpha1/agent.go
```diff
@@ -0,0 +1,95 @@
...
+			keys[item] = struct{}{}
+			result = append(result, item)
+		}
+	}
+
+	return result
+}
+
+//nolint:gocognit
+func (a *Agent) Merge(other *Agent) {
```
paralta: A simple unit test for this would be useful

cli/builder/builder.go
paralta: these details should be filled at `BuildUserAgent()`

cli/build.config.yaml
```diff
@@ -56,5 +74,5 @@ builder:
   # Enable or disable the CrewAI extension
   crewai: false
 
-  # Enable or disable the runtime extension
+  # Enable or disable the runtime
```
paralta: replace all with `# Enable or disable the ** plugin` just because this is the example config file

cli/cmd/build/build.go
```diff
@@ -43,33 +40,24 @@ func runCommand(cmd *cobra.Command) error {
...
-	agent := &apicore.Agent{
-		Name:       cfg.Model.Name,
-		Version:    cfg.Model.Version,
-		Authors:    cfg.Model.Authors,
-		CreatedAt:  timestamppb.New(time.Now()),
-		Skills:     cfg.Model.Skills,
-		Locators:   locators,
-		Extensions: extensions,
-	}
+	agent.Merge(builderAgent)
```
paralta, line 62:
```suggestion
	// Merge Agent Model from user config with Agent Model from plugins
	// User model will override plugin model
	agent.Merge(builderAgent)
```

api/core/v1alpha1/agent.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	return result
+}
+
+//nolint:gocognit,cyclop
+func (x *Agent) Merge(other *Agent) {
+	if x == nil {
+		return
+	}
```
paralta: return other if nil

api/core/v1alpha1/agent.go
```diff
@@ -0,0 +1,115 @@
...
+
+	// Merge annotations, keeping receiver's values when keys conflict
+	if x.GetAnnotations() == nil {
+		x.Annotations = make(map[string]string)
+	}
```
paralta: maybe move this closer to annotations' logic below

api/core/v1alpha1/agent.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if other == nil {
+		return
+	}
+
+	// Only use other's scalar fields if receiver doesn't have them set
```
paralta: only a suggestion but I think this func would be cleaner if we used a firstNonEmpty  like at https://github.com/agntcy/dir/pull/23/files#diff-b4a482fbbe5d5a637248ae9b746b338fa92ceaf485f5f762fc21c9511178571dR115

api/core/v1alpha1/agent.go
```diff
@@ -0,0 +1,115 @@
...
+			if ext != nil {
+				extensionMap[ext.GetName()] = ext
+			}
+		}
+
+		x.Extensions = make([]*Extension, 0, len(extensionMap))
+		for _, ext := range extensionMap {
+			x.Extensions = append(x.Extensions, ext)
+		}
+	}
```
paralta: same as above but if we had a `mergeSlices` func we would not have to repeat this code

PR #45, opened 2025-03-17
feat: interconnect arch components for API serving

ramizpolic:
This PR addresses:
- Public API: adding typedefs for missing models and APIs
- Architecture: initial draft, common internal datastore for local data mgmt
- Networking: initial libp2p node to support routing/networking APIs
- Config fixes: wrap/inherit config objects
- OCI store improvements: auth, config, local/remote mode, content tagging
- Store improvements: benchmarking, flow testing
- Caching

Breaking changes:
- model objects
- server config keys

Next up:
- Announcement APIs (ie. update routing table on request received)
- Discovery APIs (ie. querying the routing tables)
- Networking (ie. operational modes, testing, benchmarks)
- Update CSIT before the next release


api/core/v1alpha1/locator.go
ramizpolic, line 1: we need a better way to manage string types between golang and protobuf

api/core/v1alpha1/skill.go
```diff
@@ -0,0 +1,10 @@
+// SPDX-FileCopyrightText: Copyright (c) 2025 Cisco and/or its affiliates.
+// SPDX-License-Identifier: Apache-2.0
+
+package corev1alpha1
+
+import "fmt"
+
+func (skill *Skill) FQDN() string {
```
ramizpolic, line 8:
```suggestion
func (skill *Skill) FQDN() (string, error) {
```
we may not be able to construct FQDN for a skill to a string

api/core/v1alpha1/locator.proto
```diff
@@ -5,41 +5,33 @@ syntax = "proto3";
...
+  // Specs: https://datatracker.ietf.org/doc/html/rfc1738
+  string url = 2;
+  // Metadata associated with this locator.
+  map<string, string> annotations = 3;
+  // Size of the source in bytes pointed by the {url} property.
+  optional uint64 size = 4;
+  // Digest of the source pointed by the {url} property.
+  // Specs: https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests
+  optional string digest = 5;
 }
```
paralta, line 27:
```suggestion
optional ObjectRef data_ref = 5;
}
```
missing?

api/core/v1alpha1/object.proto
```diff
@@ -5,51 +5,34 @@ syntax = "proto3";
...
-  // This value should be used for custom objects to describe their schema.
-  // For existing objects such as agent data models, this value is ignored.
-  string kind = 2;
-
-  // Arbitrary name specifier of this object
-  string name = 3;
-
-  // Annotations contains arbitrary metadata
+// Reference to a typed object in the content store.
+message ObjectRef {
```
paralta, line 9: you mentioned this today and I agree that `ObjectMeta` would be more appropriate here

api/core/v1alpha1/skill.go
paralta: why? when skills are not from OASF? is this possible?

api/core/v1alpha1/skill.proto
```diff
@@ -0,0 +1,31 @@
...
+  // UID of the class.
+  string class_uid = 2;
+  // Name of the category.
+  optional string category_name = 3;
+  // Name of the class.
+  optional string class_name = 4;
+  // Additional metadata for this skill.
+  map<string, string> annotations = 5;
+  // Optional schema version.
+  optional string version = 6;
```
paralta: should this be optional? we might need to add OASF validation later and this will be needed

api/core/v1alpha1/extension.proto
```diff
@@ -5,22 +5,29 @@ syntax = "proto3";
...
+// Key := {name}/{version}
 //
 // This is an immutable object.
 message Extension {
-  // Name of the extension attached to this agent
+  // Name of the extension attached to an agent.
   string name = 1;
-  // Version of the extension attached to this agent
+  // Version of the extension attached to an agent.
   string version = 2;
```
paralta, line 21: will this be OASF version for cases where the extension is from OASF?

api/routing/v1alpha1/routing.proto
```diff
@@ -0,0 +1,31 @@
...
+message Key {
+  // Full path of the key that holds some data.
+  repeated string path = 1;
+}
+
+message Peer {
+  // ID of a given peer.
+  bytes id = 1;
+  // Multiaddrs for a given peer.
+  repeated bytes addrs = 2;
```
paralta: why `bytes` here? can we just have `string` here? it overcomplicated stuff with the digest before

api/core/v1alpha1/locator.proto
ramizpolic: its not missing, we want to prevent using the storage layer as a potential LFS store. our store API is local only (with a possible remote sync), and its not intended to be used as a storage serving service externally. actual locator data should be stored elsewhere

api/core/v1alpha1/object.proto
ramizpolic: i decided to keep it as is because meta is optional, ref itself isnt. we can come back to this API later on and decouple meta.

api/core/v1alpha1/skill.go
ramizpolic: this was just a note to self that we cant construct FQDN if we have missing data on struct, so we need validation of some kind

api/core/v1alpha1/extension.proto
ramizpolic: yes, thats the idea

api/routing/v1alpha1/routing.proto
ramizpolic: will update

api/core/v1alpha1/skill.proto
ramizpolic: will update, nice catch

api/routing/v1alpha1/routing.proto
ramizpolic: i used libp2p API defs because interconnecting the data with DHT with bytes wont require an intermediate transform to bytes step which will be the case with strings. but these are details, we can change

api/core/v1alpha1/locator.proto
paralta: so following this logic, should the `ObjectRef ` (currently commented out) be removed from api/core/v1alpha1/extension.proto?

ramizpolic: it is partially true, yes. i am still unsure what constraints we should add on extensions, so it's just there as a comment for now

client/store.go
```diff
@@ -47,15 +49,13 @@ func (c *Client) Push(ctx context.Context, meta *coretypes.ObjectMeta, reader io
...
 }
 
-func (c *Client) Pull(ctx context.Context, dig *coretypes.Digest) (io.Reader, error) {
-	req := &coretypes.ObjectRef{
-		Digest: dig,
-	}
-
-	stream, err := c.StoreServiceClient.Pull(ctx, req)
+// Pull
+// TODO: needs to pull each extension data individually and send back.
```
paralta, line 56: maybe remove these `extension` TODOs for now? this might be confusing since only agent models will have `ObjectRef `, ie for now we cannot pull extensions only models

PR #46, opened 2025-03-18
feat: routing publish and list

paralta:
* Query datastore for `skills`, `locators` and `extensions`
* Test invalid queries

### DIFF
--- a/server/routing/routing.go
+++ b/server/routing/routing.go
@@ -57,16 +57,6 @@ func (r *routing) Publish(ctx context.Context, ref *coretypes.ObjectRef, a *core
 		}
 	}
 
-	// Extensions
-	for _, e := range a.GetExtensions() {
-		k := datastore.NewKey(fmt.Sprintf("extensions/%s/%s", e.Key(), ref.GetDigest()))
-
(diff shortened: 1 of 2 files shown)


api/core/v1alpha1/skill.go
```diff
@@ -8,3 +8,7 @@ import "fmt"
 func (skill *Skill) FQDN() string {
 	return fmt.Sprintf("%s/%s", skill.GetCategoryName(), skill.GetClassName())
 }
+
+func (skill *Skill) Key() string {
+	return skill.FQDN()
```
ramizpolic: feel free to remove the FQDN

api/routing/v1alpha1/routing_service.proto
```diff
@@ -46,11 +46,8 @@ message PublishRequest {
 }
 
 message ListRequest {
-  // Key to be listed
-  Key key = 1;
-
   // Regexp query for sub-keys
-  string query = 2;
```
ramizpolic, line 53: since we only allow skill-based listing over the DHT, we could potentially on the API accept skill-only queries

PR #55, opened 2025-03-20
feat: routing client and client e2e test

paralta:
* Add routing client
* Add e2e test for client


e2e/utils.go
```diff
@@ -0,0 +1,72 @@
...
+	"fmt"
+	"reflect"
+	"sort"
+
+	coretypes "github.com/agntcy/dir/api/core/v1alpha1"
+)
+
+func Ptr[T any](v T) *T {
+	return &v
+}
```
ramizpolic, line 17: todo for later: utils package to reuse across dir packages with common things (logging, libs, etc)

paralta: completely agree 👍

api/routing/v1alpha1/routing_service.proto
```diff
@@ -43,6 +44,9 @@ message PublishRequest {
 
   // Published record reference.
   core.v1alpha1.ObjectRef record = 2;
+
+  // Published agent.
+  core.v1alpha1.Agent agent = 3;
```
ramizpolic, line 49: not needed here, we can look it up from the storage service on publish controller

ramizpolic: we need to pass storeAPI interface to the routing controller

paralta: this was required on the client because the server is requiring it for now. the changes need to happen on the server then

ramizpolic: oh ok, lets keep it, just realized its only here for the sake of the client test

PR #57, opened 2025-03-21
feat: routing publish without agent

paralta:
* Agent should not be required to publish, routing API should use the store API to get the agent.
* Add lint cache

### DIFF
--- a/server/controller/routing.go
+++ b/server/controller/routing.go
@@ -13,6 +13,7 @@ import (
 	coretypes "github.com/agntcy/dir/api/core/v1alpha1"
 	routingtypes "github.com/agntcy/dir/api/routing/v1alpha1"
 	"github.com/agntcy/dir/server/types"
+	"github.com/opencontainers/go-digest"
 	"google.golang.org/protobuf/types/known/emptypb"
 )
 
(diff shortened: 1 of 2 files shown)


server/controller/routing.go
```diff
@@ -35,11 +39,22 @@ func (c *routingCtlr) Publish(ctx context.Context, req *routingtypes.PublishRequ
 		return nil, errors.New("digest is required")
 	}
 
-	if req.GetAgent() == nil {
-		return nil, errors.New("agent is required")
+	reader, err := c.store.Pull(ctx, req.GetRecord())
```
ramizpolic: lets add a sanity check before this to lookup via digest and validate if its an agent and if the size is smaller than 4MB

server/controller/routing.go
```diff
@@ -35,11 +40,40 @@ func (c *routingCtlr) Publish(ctx context.Context, req *routingtypes.PublishRequ
...
-		return nil, errors.New("agent is required")
+	// Check is digest is valid
+	_, err := digest.Parse(req.GetRecord().GetDigest())
+	if err != nil {
+		return nil, fmt.Errorf("invalid digest: %w", err)
+	}
+
+	if req.GetRecord().GetSize() == 0 {
+		return nil, errors.New("size is required")
+	}
```
ramizpolic: we don't need this, we only need the digest from the request to be able to lookup. we can do `store.Lookup` instead of this

PR #67, opened 2025-03-26
feat: add inital config file support

pbalogh-sa:


install/charts/dir/apiserver/templates/deployment.yaml
```diff
@@ -38,10 +38,18 @@ spec:
...
+              {{- end }}
               protocol: TCP
             - name: health
-              containerPort: {{ .Values.config.health_port }}
+              {{- if .Values.config.healthcheck_address }}
+              containerPort: {{ (split ":" .Values.config.healthcheck_address)._1 }}
+              {{- else }}
+              containerPort: 8889
+              {{- end }}
               protocol: TCP
```
ramizpolic, line 59: can you please attach one more port here? we are going to add the networking, and we just need 1 additional port. you can reserve a value like `8999`

server/server.config.yml
ramizpolic, line 1: feel free to move this to `server/config/future/server.config.yml` path to make it transparent that this version and current version of configs dont match

PR #70, opened 2025-03-26
feat: version cmd

paralta:
* Add version to api `<release version>-<commit hash>`
* Add version command


docker-bake.hcl
```diff
@@ -27,6 +28,9 @@ target "_common" {
     "linux/arm64",
     "linux/amd64",
   ]
+  args = {
+    EXTRA_LDFLAGS = "${EXTRA_LDFLAGS}"
```
ramizpolic, line 32: not sure this will work, i dont see who is setting this value (taskfile, git actions). can you confirm if docker build image gets  the version injected

paralta: this is set on the taskfile, I tested by checking the api server logs while running the e2e tests

PR #79, opened 2025-03-31
feat: delete agent model from store

paralta:
* Small refactor to avoid repeated code to load agent model from file and from reader
* Add method to delete agent model from store (needs implementation for both `oci.Store` and `remote.Repository`)
* Unit and E2E test added


muscariello (approved):
LGTM

PR #81, opened 2025-04-01
feat: unpublish agent

paralta:
* Implement `unpublish` method in routing service
* Unit and E2E test added


api/routing/v1alpha1/routing_service.proto
```diff
@@ -79,3 +82,14 @@ message ListResponse {
...
+
+message UnpublishRequest {
+  // Reference to the record to be unpublished.
+  // On unpublish, we read from the local store and remove the associated labels.
+  core.v1alpha1.ObjectRef record = 1;
+
+  // Notify the network about the unpublication.
+  // This will remove the item from the network
+  // and make it no longer searchable.
+  optional bool network = 2;
```
adamtagscherer, line 94: do we need this value if the unpublish only works on local nodes?

adamtagscherer (changes requested):

adamtagscherer (approved):
LGTM! :shipit:

PR #82, opened 2025-04-02
feat: add CLI support for Agent Hub

matewolf:


api/hub/buf.gen.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - proto_file: saas/v1alpha1/category.proto
+  - proto_file: saas/v1alpha1/locator.proto
+  - proto_file: saas/v1alpha1/repo_version_id.proto
+plugins:
+  - remote: buf.build/grpc/go:v1.5.1
+    out: .
+    opt: paths=source_relative
+  - remote: buf.build/protocolbuffers/go:v1.36.5
+    out: .
+    opt: paths=source_relative
```
paralta: here you are using the same plugins that we use for the core/store/routing APIs, so we could generate hub code with the same `api/buf.gen.yaml` file. this would also avoid simplify the changes in `Taskfile.yml`

Taskfile.yml
```diff
@@ -39,6 +39,8 @@ vars:
   GOLANGCI_LINT_BIN: '{{ .BIN_DIR }}/golangci-lint-{{.GOLANGCI_LINT_VERSION}}'
   LICENSEI_VERSION: '0.9.0'
   LICENSEI_BIN: '{{ .BIN_DIR }}/licensei-{{.LICENSEI_VERSION}}'
+  HUB_API_VERSION: 'main'
+  HUB_REPO_URL: 'https://github.com/cisco-eti/phoenix-saas-be.git'
```
paralta, line 45: should we keep this reference to an internal repo in the open repo? the `git clone` below will fail for users without access

api/hub/buf.gen.yaml
```diff
@@ -0,0 +1,27 @@
...
+    - file_option: go_package
+      path: core/v1alpha1
+      value: github.com/agntcy/dir/api/core/v1alpha1
+inputs:
+  - proto_file: saas/v1alpha1/agent_service.proto
+  - proto_file: saas/v1alpha1/agent_id.proto
+  - proto_file: saas/v1alpha1/agent_id_response.proto
+  - proto_file: saas/v1alpha1/agent.proto
+  - proto_file: saas/v1alpha1/category.proto
+  - proto_file: saas/v1alpha1/locator.proto
```
tkircsi: There is no `locator.proto` in the `saas/v1alpha1` package. We're using the `core/v1alpha1/locator.prot

api/hub/buf.gen.yaml
matewolf: I wouldn't like to use the same buf file and integrate `deps:api:gen` task because running this task would need a permission to reach the private hub repo. Maybe it is possible to exclude some module from `buf generate` command.

cli/cmd/root.go
```diff
@@ -8,63 +8,106 @@ import (
...
+		push.NewCommand(baseOption),
+		del.NewCommand(),
 		// routing commands
-		publish.Command,
-		list.Command,
-		unpublish.Command,
+		publish.NewCommand(baseOption),
+		list.NewCommand(baseOption),
+		unpublish.NewCommand(baseOption),
 		network.Command,
```
paralta, line 64: this new command needs to be refactored as well

cli/cli.go
```diff
@@ -17,7 +16,6 @@ func main() {
 	ctx, cancel := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGHUP, syscall.SIGTERM)
 
 	if err := cmd.Run(ctx); err != nil {
```
paralta, line 18: was this print removed by mistake?

cli/cmd/build/build.go
```diff
@@ -14,47 +14,55 @@ import (
...
+		Use:   "build",
+		Short: "Build agent model to prepare for pushing",
+		Long: `
 This command handles the build process for agent data models
 from source code. It generates a JSON object that 
 describes an agent and satisfies the **Open Agent Schema Framework** specification.
 
 Usage examples:
 
-1. When build config is present under the agent source code:
```
paralta: let's keep the formatting here, since the same numbered formatting is used in the push and pull commands

cli/cmd/list/info/info.go
```diff
@@ -9,36 +9,48 @@ import (
...
+   	dir list info
 	
-2. List summary about published data across the network:
-
+	# List summary about published data by a specific peer.
+   	dir list info --peer <peer-id>
+	
+	# List summary about published data by the whole network.
+	# NOTE: This starts a DHT walk, so it may take a while.
+	# NOTE: Results are not guaranteed to be complete and up-to-date.
```
paralta: there might have been a merge conflict issue here, please check

cli/cmd/hub/push/push.go
```diff
@@ -0,0 +1,129 @@
...
+			return fmt.Errorf("failed to get agent bytes: %w", err)
+		}
+
+		// TODO: Push based on repoName and version misleading
+		repoID, _, err := parseRepoTagID(args[0])
+		if err != nil {
+			return fmt.Errorf("failed to parse repo id: %w", err)
+		}
+
+		ctx := metadata.NewOutgoingContext(context.Background(), metadata.Pairs("authorization", "Bearer "+session.Tokens[session.CurrentTenant].AccessToken))
```
tkircsi:
```suggestion
        if session.Tokens == nil {
            return nil, errors.New("session tokens not initialized")
        }

        tokens, exists := session.Tokens[session.CurrentTenant]
        if !exists {
            return nil, errors.New("no tokens found for current tenant")
        }

        if tokens == nil {
            return nil, errors.New("tokens for current tenant are nil")
        }

        if tokens.AccessToken == "" {
            return nil, errors.New("access token is empty")
        }
        ctx := metadata.NewOutgoingContext(context.Background(), metadata.Pairs("authorization", "Bearer "+tokens.AccessToken))
```

tkircsi (commented):
This PR is big, so it makes sense to add my reviews step by step. I continue my review.

cli/hub/config/config.go
```diff
@@ -0,0 +1,79 @@
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrFetchingConfig, err)
+	}
+
+	var authConfig *AuthConfig
+	if err = json.Unmarshal(body, &authConfig); err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrParsingConfig, err)
+	}
```
tkircsi: Possible nil pointer

```suggestion
	if err = json.Unmarshal(body, &authConfig); err != nil {
		return nil, fmt.Errorf("%w: %w", ErrParsingConfig, err)
	}
	
	if authConfig == nil {
		 return nil, fmt.Errorf("%w: config is nil", ErrParsingConfig)
	}
```

cli/cmd/hub/logout/logout.go
```diff
@@ -0,0 +1,69 @@
...
+			return runCmd(cmd.OutOrStdout(), opts, session, sessionStore, oktaClient)
+		},
+		TraverseChildren: true,
+	}
+
+	return cmd
+}
+
+func runCmd(outStream io.Writer, opts *options.HubOptions, session *sessionstore.HubSession, secretStore sessionstore.SessionStore, oktaClient okta.Client) error {
+	resp, err := oktaClient.Logout(&okta.LogoutRequest{IDToken: session.Tokens[session.CurrentTenant].IDToken})
```
tkircsi:
```suggestion
	if session == nil {
		return errors.New("session is nil")
	}

	if session.Tokens == nil {
		return errors.New("session tokens are nil")
	}

	tokens, ok := session.Tokens[session.CurrentTenant]
	if !ok {
		return fmt.Errorf("no tokens found for tenant: %s", session.CurrentTenant)
	}

	if tokens == nil {
		return errors.New("tokens for current tenant are nil")
	}

	if tokens.IDToken == "" {
		return errors.New("ID token is empty")
	}

	resp, err := oktaClient.Logout(&okta.LogoutRequest{IDToken: tokens.IDToken})
```

cli/cmd/hub/pull/pull.go
```diff
@@ -0,0 +1,123 @@
...
+		},
+	}
+
+	return cmd
+}
+
+func runCmd(ctx context.Context, hc hubClient.Client, agentID *v1alpha1.AgentIdentifier, session *sessionstore.HubSession) error {
+	if session.Tokens != nil && session.Tokens[session.CurrentTenant].AccessToken != "" {
+		ctx = metadata.NewOutgoingContext(ctx, metadata.Pairs("authorization", "Bearer "+session.Tokens[session.CurrentTenant].AccessToken))
+	}
```
tkircsi:
```suggestion
	if session.Tokens != nil {
		tokens, exists := session.Tokens[session.CurrentTenant]
		if exists && tokens != nil && tokens.AccessToken != "" {
			ctx = metadata.NewOutgoingContext(ctx, metadata.Pairs("authorization", "Bearer "+tokens.AccessToken))
		}
	}
```

cli/hub/config/config.go
```diff
@@ -0,0 +1,79 @@
...
+	if !urlutil.IsURL(frontedURL) {
+		return nil, ErrInvalidFrontendURL
+	}
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, frontedURL+"/config.json", nil)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrFetchingConfig, err)
+	}
+
+	resp, err := http.DefaultClient.Do(req)
```
tkircsi: We could use later something more secure http client as we did in panoptica/policy-framework. For example

```go
func createSecureHTTPClient() *http.Client {
    return &http.Client{
        Timeout: 30 * time.Second,
        Transport: &http.Transport{
            TLSClientConfig: &tls.Config{
                MinVersion: tls.VersionTLS12,
                InsecureSkipVerify: false,
            },
            MaxIdleConns:        100,
            MaxIdleConnsPerHost: 10,
            IdleConnTimeout:     90 * time.Second,
        },
    }
}
```

cli/hub/config/config.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrFetchingConfig, err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("%w: %w: (%d) %s", ErrFetchingConfig, errors.New("unexpected status code"), resp.StatusCode, resp.Body)
```
tkircsi: Maybe this error messages expose internal details that could be useful for attackers.

cli/hub/config/config.go
```diff
@@ -0,0 +1,79 @@
...
+	IdpProductID       string `json:"IAM_PRODUCT_ID"`
+	ClientID           string `json:"IAM_OIDC_CLIENT_ID"`
+	IdpIssuerAddress   string `json:"IAM_OIDC_ISSUER"`
+	IdpBackendAddress  string `json:"IAM_API"`
+	IdpFrontendAddress string `json:"IAM_UI"`
+	HubBackendAddress  string `json:"HUB_API"`
+}
+
+func FetchAuthConfig(frontedURL string) (*AuthConfig, error) {
+	if !urlutil.IsURL(frontedURL) {
```
tkircsi: As I see, `IsURL` only checks the Schema. You can make the same with Go-s url package url.Scheme, 
```go
parsedURL, err := url.Parse(url)
if err != nil {
    return "", ErrInvalidFrontendURL
}

if parsedURL.Scheme != "https" {
    return "", fmt.Errorf("URL scheme must be HTTPS")
}
```
but it would make sense to create a method `validateAndSanitizeURL` to validate and sanitize URLs

cli/cli.go
matewolf: No. It was intended. It resulted that the error message was printed twice. Once on stderr and once on stdout.

cli/hub/api/buf.gen.yaml
```diff
@@ -0,0 +1,24 @@
+version: v2
```
paralta: why was the API code moved from the `api` directory to `cli/hub/api`? I think it would be better to maintain all API code under the same folder

cli/hub/api/v1alpha1/agent_dir_service.proto
```diff
@@ -0,0 +1,33 @@
...
+message PullAgentRequest {
+  AgentIdentifier id = 1;
+}
+
+message PullAgentResponse {
+  core.v1alpha1.Object model = 1;
+}
+
+// This API is manily for CLIs and the implementation of these APIs should communicate with the Agent Directory
+service AgentDirService {
```
paralta: just a suggestion, but I think the naming here should be something like `HubService` or `SaasService` to make a distinction between the hub and the directory

cli/cmd/list/info/info.go
```diff
@@ -10,7 +10,7 @@ import (
 
 	routetypes "github.com/agntcy/dir/api/routing/v1alpha1"
 	"github.com/agntcy/dir/cli/presenter"
-	"github.com/agntcy/dir/cli/util"
+	ctxUtils "github.com/agntcy/dir/cli/util/context"
```
paralta, line 13:
```suggestion
	util "github.com/agntcy/dir/cli/util/context"
```
this `util` -> `ctxUtils` change has not been very consistent across the PR, it would be good to either change to `ctxUtils` or to `util` not both

cli/cmd/options.go
```diff
@@ -1,18 +0,0 @@
...
-//nolint:unused
-type options struct {
-	Query string
-}
-
-func init() {
-	flags := RootCmd.PersistentFlags()
-	flags.StringVar(&clientConfig.ServerAddress, "server-addr", clientConfig.ServerAddress, "Directory Server API address")
-
-	RootCmd.MarkFlagRequired("server-addr") //nolint:errcheck
```
paralta: where was this code moved to?

cli/hub/client/hub/client.go
```diff
@@ -0,0 +1,135 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package hub
```
paralta: similar to the api code comment, maybe the client code should be moved to the `client` directory?

We can initiate a discussion on this topic, because if the hub code should be independent from the rest, it probably needs a new `hub` folder at root instead of a folder inside the `cli` folder

Taskfile.yml
matewolf: I haven't got any answer in the webex space whether I can keep these or not and I think that these are not sensitive information so I would like to keep them because they are needed by api stub generating.

cli/cmd/hub/push/push.go
matewolf: Access token validation has been introduced in `PreRun` function of push command.

cli/hub/config/config.go
matewolf: I have added extra check.

cli/cmd/hub/logout/logout.go
matewolf: Extra check has been added.

cli/cmd/hub/pull/pull.go
matewolf: I resolved the missing nil checks.

cli/hub/config/config.go
matewolf: I replaced all the `DefaultClient` with a more secure one.

cli/hub/config/config.go
matewolf: Message has been changed.

cli/hub/config/config.go
matewolf: New function has been introduced to validate the url.

cli/hub/api/buf.gen.yaml
matewolf: I moved everything under the hub folder because later all of the code related to hub cli should be moved to a separated github repository and it should be used as one hub implementation.

cli/hub/api/v1alpha1/agent_dir_service.proto
matewolf: The name is come from hub side and it is synced from there. From the hub side these interfaces are the functionalities that dir provides and hub just extends them. The name should be considered as a name from the hub side.

matewolf: Btw, the other set of functionalities that only hub provides are called `HubService` on hub side.

cli/cmd/list/info/info.go
matewolf: I changed everywhere.

cli/hub/client/hub/client.go
matewolf: The reason is the same I mentioned above.

cli/cmd/options.go
matewolf: I deleted this because it was not used according to my ide.

matewolf: I put it back.

PR #83, opened 2025-04-02
ci: add local demo workflow 

paralta:
* Fix zot service name in helm values
* Add demo workflow that runs agent directory build/push/publish/pull commands with one dir server node
* Example run https://github.com/paralta/dir/actions/runs/14226199344/job/39866198246

### DIFF
--- a/.github/workflows/demo-local.yaml
+++ b/.github/workflows/demo-local.yaml
@@ -1,11 +1,11 @@
 # Copyright AGNTCY Contributors (https://github.com/agntcy)
 # SPDX-License-Identifier: Apache-2.0
 
-name: Demo
+name: Dir - Local Demo
 
 description: |
   This workflow demonstrates how to use the Dir CLI with a Kind cluster and Helm chart.
-  It includes steps for building, pushing, publishing, and listing images.
+  It includes steps for building, pushing, publishing, and listing agents.
 
 on:
   workflow_dispatch:
@@ -47,10 +47,6 @@ jobs:
       - name: Load image into Kind
         run: |
           kind load docker-image ${{ inputs.dir-apiserver-image }}:${{ inputs.dir-version }} --name dir-demo
-
-      - name: Login to OCI Registry
-        run: |
-          helm registry login ghcr.io -u notused -p ${{ secrets.GITHUB_TOKEN }}
 
       # TODO Remove values file once release with fix is available
       - name: Deploy Helm Chart
--- a/.github/workflows/reusable-dir-demo.yaml
+++ b/.github/workflows/reusable-dir-demo.yaml
@@ -1,0 +1,116 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
+name: Demo
+
+description: |
+  This workflow demonstrates how to use the Dir CLI with a Kind cluster and Helm chart.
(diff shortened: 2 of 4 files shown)


.github/workflows/reusable-dir-demo.yaml
```diff
@@ -0,0 +1,116 @@
...
+        run: |
+          docker pull ${{ inputs.dir-apiserver-image }}:${{ inputs.dir-version }}
+
+      - name: Load image into Kind
+        run: |
+          kind load docker-image ${{ inputs.dir-apiserver-image }}:${{ inputs.dir-version }} --name dir-demo
+
+      - name: Login to OCI Registry
+        run: |
+          helm registry login ghcr.io -u notused -p ${{ secrets.GITHUB_TOKEN }}
```
pbalogh-sa: I think we don't need to login to the OCI registry because it is open.

.github/workflows/reusable-dir-demo.yaml
```diff
@@ -0,0 +1,116 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+name: Demo
+
+description: |
+  This workflow demonstrates how to use the Dir CLI with a Kind cluster and Helm chart.
+  It includes steps for building, pushing, publishing, and listing images.
+
+on:
+  workflow_dispatch:
```
pbalogh-sa: I think, if it is as a reusable workflow, it should be called from an other workflow. So it should be `workflow_call` instead of `workflow_dispatch`

.github/workflows/reusable-dir-demo.yaml
paralta: it failed when I tested on the forked repo, let me test again

.github/workflows/reusable-dir-demo.yaml
paralta: I will change the `reusable-*` naming then because I think this is something we want to run ad-hoc to demo the agent flow

.github/workflows/reusable-dir-demo.yaml
paralta: https://github.com/paralta/dir/actions/runs/14239794678/job/39906837077 indeed :D

.github/workflows/demo.yaml
ramizpolic, line 1: `demo_local.yaml`

.github/workflows/demo.yaml
```diff
@@ -0,0 +1,112 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
+name: Demo
```
ramizpolic:
```suggestion
name: Dir - Local Example
```

ramizpolic (approved):

.github/workflows/demo-local.yaml
```diff
@@ -0,0 +1,112 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
+name: Dir - Local Demo
+
+description: |
+  This workflow demonstrates how to use the Dir CLI with a Kind cluster and Helm chart.
+  It includes steps for building, pushing, publishing, and listing images.
```
ramizpolic:
```suggestion
  It includes steps for building, pushing, publishing, and listing agents.
```

PR #85, opened 2025-04-03
feat(cli): add network cli command

adamtagscherer:
This PR adds a `network init` and `network info` commands to the `dirctl` that generates a peer id from a ED25519 private key.
This peer id can be used for the node to connect to the DHT network.


paralta (commented):
Please confirm but I understood that this command should generate the key pair (like https://github.com/agntcy/dir/blob/main/server/routing/internal/p2p/options.go#L146) and not read it from a file. It would then use this generated key to create the peer identity. 

Similar to this IPFS command https://docs.ipfs.tech/how-to/command-line-quick-start/#initialize-the-repository

pbalogh-sa:
> Please confirm but I understood that this command should generate the key pair (like https://github.com/agntcy/dir/blob/main/server/routing/internal/p2p/options.go#L146) and not read it from a file. It would then use this generated key to create the peer identity.
> 
> Similar to this IPFS command https://docs.ipfs.tech/how-to/command-line-quick-start/#initialize-the-repository

I think we have to let the possibility to generate peerID from an existing private key file.
In operations point of view it is a must.

adamtagscherer:
There will be 2 commands:
`dirctl network init` -> golang code generates a key and a peer id from it
`dirctl network info` -> golang code reads the key from the filesystem and generates a peer id from it

wdyt?

PR #87, opened 2025-04-04
feat: add persistance for storage and routing

ramizpolic:
Adds datastore persistance for storage metadata and peer/content routing data


server/types/api.go
```diff
@@ -23,24 +23,16 @@ type API interface {
 type APIOptions interface {
 	// Config returns the config data. Read only! Unsafe to edit.
 	Config() *config.Config
-
-	// Datastore holds access to local datastore.
-	// Used as a local data source to serve routing, storage, and caching.
-	Datastore() Datastore
```
ramizpolic, line 29: i should have probably handled it directly through this interface, but it required further extensions and handling different providers for storage and routing.

id like to bring it back here eventually once we have enough time to properly handle provider-specific flows and configs.

paralta (approved):
LGTM 👍

PR #89, opened 2025-04-07
feat(cli): change network commands

adamtagscherer:
With this PR the `dirctl network init` command will save the generated key to a file on the filesystem.


paralta (approved):
LGTM

PR #92, opened 2025-04-09
fix: rename zot-config.json to config.json

subbaksh:
Keep getting the following error when running `task server:start`
```
Error: statfs /tmp/config.json: no such file or directory
```


ramizpolic (approved):

paralta (approved):
LGTM 🚀

PR #94, opened 2025-04-10
feat: python client SDK

adamtagscherer:
This PR adds the python client SDK, that can be used to call the server's methods from python code.


Taskfile.yml
```diff
@@ -169,7 +169,7 @@ tasks:
   ##
   client:test:
     desc: Unit test Client code
-    dir: ./client
+    dir: ./client/go
```
ramizpolic: lets keep the client golanf code as is and move only the python code to 'sdk/python/client'. i would only suggest to keep the location of golang code as is

api/buf.gen.yaml
```diff
@@ -14,10 +14,22 @@ plugins:
...
+      - paths=source_relative
+      - require_unimplemented_servers=false
+      - use_generic_streams_experimental=false
+  # Stubs for Python
+  - remote: buf.build/protocolbuffers/python:v29.3
+    out: ../client/python/client/proto
+  - remote: buf.build/grpc/python:v1.71.0
+    out: ../client/python/client/proto
+  - remote: buf.build/protocolbuffers/pyi:v30.2
+    out: ../client/python/client/proto
```
paralta: why do we need this `protocolbuffers/pyi`?

paralta (commented):
As discussed, we should look into keeping the generated code under the `api` directory like we have for go

api/buf.gen.yaml
adamtagscherer: if you check the generated `.pyi` files there are the actual classes defined, without them we can't import nothing because there nothing to import

PR #97, opened 2025-04-14
feat: add description to agent model

paralta:
* Add description to agent model

Enhance the Agent Data Model by adding a description field to provide detailed context and information about each agent. This will come in handy for future semantic search capabilities, making it easier to find and understand agents based on their descriptions.


adamtagscherer (approved):
LGTM! :shipit:

adamtagscherer:
shouldn't we also add description to the `e2e/testdata/examples` models?

PR #98, opened 2025-04-14
feat: copy network deploy from csit

paralta:
Instructions/tasks for multi peer deployment should also be available in the DIR repository to allow users to quickly try-out this deployment type

* Copy network deployment code from CSIT
* Change it to deploy the local helm chart

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -310,12 +310,24 @@ tasks:
           --wait-for-jobs \
           --timeout "15m"
 
+  deploy:kubernetes:local:port-forward:
+    desc: Set up port-forwarding for the local deployment
+    vars:
+      # Helm args
+      HELM_NAMESPACE: '{{ .HELM_NAMESPACE | default "default" }}'
+    cmds:
       # Port-forward dependency services
       - |
         {{ .KUBECTL_BIN }} port-forward service/dir-apiserver 8888:8888 -n {{ .HELM_NAMESPACE }} &
 
       # Delay to ensure services are online
       - sleep 10
+
+  deploy:kubernetes:local:port-forward:cleanup:
+    desc: Cleanup port-forwarding processes
+    cmds:
+      # Kill any existing port-forward processes for the dir-apiserver service
+      - kill -9 $(ps aux | grep port-forward | grep dir-apiserver | awk '{print $2}') || true
 
   deploy:kubernetes:local:cleanup:
     aliases: [deploy:local:cleanup]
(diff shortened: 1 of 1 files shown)


Taskfile.yml
```diff
@@ -346,22 +312,159 @@ tasks:
...
+    aliases: [deploy:local:cleanup]
+    desc: Cleanup Kubernetes environment for local deployment
     deps:
       - deps:kind
     vars:
       # Kind args
-      KIND_CLUSTER_NAME:  '{{ .KIND_CLUSTER_NAME | default "test-agntcy-cluster" }}'
+      KIND_CLUSTER_NAME: '{{ .KIND_CLUSTER_NAME | default "agntcy-cluster" }}'
+    cmds:
+      - '{{ .KIND_BIN }} delete cluster --name {{ .KIND_CLUSTER_NAME }}'
```
pbalogh-sa, line 341: I think the port-forward cleanup is missing here. Maybe it would be better to create a teardown task for the port-forward and use it in the tests with defer.

Taskfile.yml
```diff
@@ -346,22 +312,159 @@ tasks:
...
+            --install \
+            --wait \
+            --wait-for-jobs \
+            --timeout "15m"
+
+      # Set up port-forwarding for the peers
+      - kubectl port-forward svc/agntcy-dir-apiserver -n peer1 8890:8888 &
+      - kubectl port-forward svc/agntcy-dir-apiserver -n peer2 8891:8888 &
+      - kubectl port-forward svc/agntcy-dir-apiserver -n peer3 8892:8888 &
+      - sleep 1
```
pbalogh-sa: Maybe it would be better to create a task for port-forwards, and use it before tests along with defer teardown task of the port-forward. (In this case we will be able to use the same directory infrastructure for other tests without destroying and recreating the whole kind cluster and we can manage it in a more granular way). WDYT?

Taskfile.yml
```diff
@@ -346,22 +312,159 @@ tasks:
 
       # Port-forward dependency services
       - |
-        {{ .KUBECTL_BIN }} port-forward service/dir-apiserver  8888:8888   -n {{ .HELM_NAMESPACE }} &
+        {{ .KUBECTL_BIN }} port-forward service/dir-apiserver 8888:8888 -n {{ .HELM_NAMESPACE }} &
```
pbalogh-sa, line 321: Maybe it would be better to create a task for port-forwards, and use it before tests along with defer teardown task of the port-forward. WDYT?

paralta:
@pbalogh-sa thanks for reviewing! I agree with the comments, currently working on it 🚧

pbalogh-sa (approved):

PR #101, opened 2025-04-16
feat: pyproject parser plugin

paralta:
* Implement new plugin to automatically generate an Agent Base Model from this file which will provide users a valuable starting point for building a Model directly from the source code path.
* The pyproject.toml file in the Agent source code contains essential data required by the Agent Data Model, such as the project name, version, description, and authors. For example:

```
[tool.poetry]
name = "simple"
version = "0.1.0"
description = "simple using crewAI"
authors = ["Your Name [you@example.com](mailto:you@example.com)"]

…
```

### DIFF
--- a/USAGE.md
+++ b/USAGE.md
@@ -50,6 +50,9 @@ builder:
  # Disable the runtime plugin
  runtime: false
 
+ # Disable the pyprojectparser plugin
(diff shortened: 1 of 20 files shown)


adamtagscherer (approved):
LGTM! :shipit:

PR #110, opened 2025-05-06
feat: add docker compose installation

paralta:
* Add docker compose file to deploy DIR with the `localfs` as store and `local` mode configuration (default)
* Add `oci` profile to deploy DIR with the `zot` as store and `local` mode configuration


ramizpolic (approved):
lgtm

PR #112, opened 2025-05-07
feat: synchronize data models with OASF

akijakya:


api/core/v1alpha1/extension.proto
```diff
@@ -18,7 +18,7 @@ message Extension {
   // Name of the extension attached to an agent.
   string name = 1;
   // Version of the extension attached to an agent.
-  string version = 2;
+  optional string version = 2;
```
ramizpolic: perhaps we keep this as it was, but we can treat empty value as unset (ie. to come from schema version)

ramizpolic (approved):
LGTM

ramizpolic:
@rappizs please confirm if this is safe to merge and wont impact Hub

rappizs:
Sorry for the late review @ramizpolic, if there are breaking changes then we will update the hub logic accordingly. Jus one comment: modifying the order (field number) of properties in proto objects is usually considered a bad practice and it's quite dangerous. It's ok for now, as the service is only used by the hub and we are before launch, but in the future I would try to avoid such changes if possible.

ramizpolic:
> Sorry for the late review @ramizpolic, if there are breaking changes then we will update the hub logic accordingly. Jus one comment: modifying the order (field number) of properties in proto objects is usually considered a bad practice and it's quite dangerous. It's ok for now, as the service is only used by the hub and we are before launch, but in the future I would try to avoid such changes if possible.

Thanks @rappizs. We are aware of the proto changes, AFAIK these do not matter for now as we deal with JSON at rest and gRPC in transit.

ramizpolic (approved):
lgtm :shipit:

PR #113, opened 2025-05-07
refactor: separate api bindings

adamtagscherer:
This PR separates the `python` and `golang` proto file bindings in their separate directory, so that later on we will be able to easily add new language bindings if needed.
This PR also removes the `HTTP` gateway bindings from the generated files because we no longer have to support a web frontend.


paralta (approved):
LGTM 👍 agree with the gateway removal for now, I am not aware of any service that might require HTTP request support

PR #114, opened 2025-05-08
fix: list command outputs with network flag

paralta:
Check https://github.com/agntcy/dir/actions/runs/14832685915/job/41637153235#step:12:15 for logs showing the issues described below

* When an agent was pushed and published on Peer 1, the output of `dirctl list "/skills/Natural Language Processing" --server-addr peer1 --network` was empty. This was happening because we skipped running list on self, thus no agent was found. Please check https://github.com/agntcy/dir/commit/c11a1da5bfbf5ac8320ce9b3a83b57149e394a8f for the fix.
* On a network with a bootstrap node and 3 other nodes, when an agent was pushed and published on Peer 1, the output of `dirctl list "/skills/Natural Language Processing" --server-addr peer2 --network` was 4 agents each on a different node. This happened because all "searched" peers are connected to the peer that holds the agent and output was printing the incorrect Peer ID for the agent. To avoid repeated output items, we need to skip already seen items. Please note that there is planned roadmap work to enhance how we traverse peers, check https://github.com/agntcy/dir/commit/71ccce21d689a69b587ce9e49b998a3200014817 for the fix
* Add a base e2e test for network deployments
* Fix linter

### DIFF
--- a/.github/workflows/reusable-test-e2e.yaml
+++ b/.github/workflows/reusable-test-e2e.yaml
@@ -1,0 +1,143 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
+name: Test
+
+on:
+  workflow_call:
+    inputs:
+      image_repo:
+        required: true
+        type: string
+        description: 'Image repo to use.'
+      image_tag:
+        required: true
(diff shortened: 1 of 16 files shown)


ramizpolic (approved):
LGTM

PR #116, opened 2025-05-09
refactor: distroless docker images

adamtagscherer:
This PR removes the `alpine` runner image from the `cli` and `server` dockerfiles and substitutes it with a `distroless` image.

### DIFF
--- a/cli/Dockerfile
+++ b/cli/Dockerfile
@@ -35,7 +35,7 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 
 RUN xx-verify /bin/dirctl
 
-FROM gcr.io/distroless/static:nonroot
+FROM gcr.io/distroless/static:nonroot@sha256:c0f429e16b13e583da7e5a6ec20dd656d325d88e6819cafe0adb0828976529dc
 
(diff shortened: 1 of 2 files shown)


rappizs (approved):
LGTM

muscariello (approved):
LGTM

cli/Dockerfile
```diff
@@ -35,12 +35,9 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 
 RUN xx-verify /bin/dirctl
 
-FROM alpine:3.20@sha256:beefdbd8a1da6d2915566fde36db9db0b524eb737fc57cd1367effd16dc0d06d
-
-RUN apk add --update --no-cache \
-    ca-certificates \
-    libc6-compat
+FROM gcr.io/distroless/static:nonroot
```
ramizpolic: can we pin it against a digest, in addition to the tag

ramizpolic: im not sure the dependency tools will pick these up later on

rappizs: Good idea, we will do that in the hub image too

ramizpolic (approved):
lgtm

PR #121, opened 2025-05-09
chore: update brew formula version

ramizpolic:


ramizpolic:
@arpad-csepi not sure if we missed something during release or not, but it seems that hub details were removed. on another note: hub is now packaged as part of dirctl, so im guessing its okay here. could you please review if this is okay?

arpad-csepi:
> @arpad-csepi not sure if we missed something during release or not, but it seems that hub details were removed. on another note: hub is now packaged as part of dirctl, so im guessing its okay here. could you please review if this is okay?

I will rework this brew formula to remove the hub option because there is no seperate hub binary. Otherwise it is looking good, I made a CI workflow to update the hashes of binaries in the brew formula upon release. (This CI workflow is only needed because the binary release does not play nicely with brew, forexample the binary hash is not released as a different file or not "bottled" (tar.xz-ed, or something).)

So if okay for you then I will update this brach with the fixed formula later.

arpad-csepi (approved):

PR #128, opened 2025-05-14
refactor: api bindings

adamtagscherer:
This PR moves the python API bindings from `api/python` to `tools/python`.


ramizpolic (approved):
LGTM

PR #133, opened 2025-05-15
fix: python sdk

adamtagscherer:
This PR fixes a bug where the python client converted the uint64 values to string before pushing to the server which later caused errors.
There is sadly no way to easily fix this without us implementing a function that converts strings to ints: https://github.com/protocolbuffers/protobuf/issues/2954

### DIFF
--- a/tools/python/client/example.py
+++ b/tools/python/client/example.py
@@ -6,6 +6,21 @@ from client.client import Client, Config
 from core.v1alpha1 import object_pb2, agent_pb2, skill_pb2, extension_pb2
 from routing.v1alpha1 import routing_service_pb2 as routingtypes
 
+def convert_uint64_fields_to_int(d):
+    """
+    Recursively convert specific fields known to be uint64 from string to int.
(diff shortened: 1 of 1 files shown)


tools/python/client/example.py
```diff
@@ -6,6 +6,21 @@
 from core.v1alpha1 import object_pb2, agent_pb2, skill_pb2, extension_pb2
 from routing.v1alpha1 import routing_service_pb2 as routingtypes
 
+def convert_uint64_fields_to_int(d):
```
ramizpolic, line 9: this is not a solution in the client code, but rather on the user side. can we make this a method on the object, ie. `Agent.dict`

tools/python/client/example.py
ramizpolic, line 1: wdyt if we move the api/client into the same python package, ie. `dirctl` package?
it would make everything easier and streamline into a shared package.

ramizpolic (approved):

PR #134, opened 2025-05-19
feat: add agent signing and verification

ramizpolic:
Adds commands for signing and verifying signatures of Agents using keyless OIDC identity-based flow. 
Uses Sigstore and relies on public-good Sigstore instance by default with optional flags to switch between instances.

Usage example: `dirctl pull <digest> | dirctl sign --stdin | dirctl verify --stdin`


muscariello (approved):
LGTM

PR #153, opened 2025-05-29
feat: search API

adamtagscherer:
This PR adds an index database to the server where we save some parts of the `Record` object. Based on the saved information we can do searching on them.


server/controller/store.go
```diff
@@ -128,6 +132,11 @@ func (s storeCtrl) Push(stream storetypes.StoreService_PushServer) error {
 		return status.Errorf(st.Code(), "failed to push object to store: %s", st.Message())
 	}
 
+	err = s.search.AddRecord(v1alpha1.NewAgentAdapter(agent))
```
ramizpolic, line 135: if the store pushes the object properly, but it fails creating an index on this step, we will not be able to index it again, at least not via push API.

we would need either a way to list things from the store so that we can submit it for indexing in the background, or a different, atomic way of indexing here. we can sort this out in another PR.

@adamtagscherer please create a followup ticket for this

server/search/config/config.go
```diff
@@ -0,0 +1,17 @@
...
+	DefaultDBType       = "sqlite"
+	DefaultSQLiteDBPath = "/data/search.db"
+)
+
+type Config struct {
+	// DBType is the type of the search database.
+	DBType string `json:"db_type,omitempty" mapstructure:"db_type"`
+
+	// SQLiteDBPath is the path to the SQLite database file.
+	SQLiteDBPath string `json:"sqlite_db_path,omitempty" mapstructure:"sqlite_db_path"`
```
ramizpolic: this can be registered here using provider-specific config struct, ie.

```go
// server/search/sqlite/config/config.go
type Config struct {
	DBPath string `json:"db_path,omitempty" mapstructure:"db_path"`
}

// server/search/config/config.go
type Config struct {
	// DBType is the type of the search database.
	DBType string `json:"db_type,omitempty" mapstructure:"db_type"`

	// Config for SQLite db provider
	SQLite sqliteconfig.Config `json:"sqlite,omitempty" mapstructure:"sqlite"`
}
```

ramizpolic: helps with extendability so that each provider can define their own configs independently

server/search/sqlite/extension.go
```diff
@@ -0,0 +1,42 @@
...
+		AgentID: agentID,
+		Name:    extensionObject.GetName(),
+		Version: extensionObject.GetVersion(),
+	}
+
+	if err := tx.Create(extension).Error; err != nil {
+		return 0, fmt.Errorf("failed to add extension to SQLite search database: %w", err)
+	}
+
+	logger.Info("Added extension to SQLite search database", "agent_id", agentID, "extension_id", extension.ID)
```
ramizpolic: i would switch this to debug level since we can have many extensions for a single record, and a lot of record. it may log too many items for each insertion

server/search/sqlite/locator.go
```diff
@@ -0,0 +1,42 @@
...
+		AgentID: agentID,
+		Type:    locatorObject.GetType(),
+		URL:     locatorObject.GetUrl(),
+	}
+
+	if err := tx.Create(locator).Error; err != nil {
+		return 0, fmt.Errorf("failed to add locator to SQLite search database: %w", err)
+	}
+
+	logger.Info("Added locator to SQLite search database", "agent_id", agentID, "locator_id", locator.ID)
```
ramizpolic: same note about logging as for extensions

server/search/sqlite/record.go
```diff
@@ -0,0 +1,192 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/agntcy/dir/server/types"
+	"gorm.io/gorm"
+)
+
+type Record struct {
```
ramizpolic, line 14: we miss two fields/methods here: `GetCID() string` and `GetCreatedAt() time.Time`.
these are important to for other ops such as for store and hub APIs

server/search/sqlite/skill.go
```diff
@@ -0,0 +1,42 @@
...
+		AgentID: agentID,
+		SkillID: skillObject.GetID(),
+		Name:    skillObject.GetName(),
+	}
+
+	if err := tx.Create(skill).Error; err != nil {
+		return 0, fmt.Errorf("failed to add skill to SQLite search database: %w", err)
+	}
+
+	logger.Info("Added skill to SQLite search database", "agent_id", agentID, "skill_id", skill.ID)
```
ramizpolic: same log level as extensions and locators

server/search/sqlite/record.go
ramizpolic: for the CID calculation, let it return an empty value for now. we need to first discuss how to leverage IDs from the internal converter to properly calculate the CID for a specific schema.

please add this to the backlog as well

we need to push first #157 to the codebase and have CID calculator defined for records first

server/types/search.go
```diff
@@ -0,0 +1,121 @@
...
+type RecordObject interface {
+	GetName() string
+	GetVersion() string
+
+	GetSkillObjects() []SkillObject
+	GetLocatorObjects() []LocatorObject
+	GetExtensionObjects() []ExtensionObject
+}
+
+type Filters struct {
```
ramizpolic:
```suggestion
type RecordFilters struct {
```

server/types/search.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+}
+
+type SearchAPI interface {
+	// AddRecord adds a new agent record to the search database.
+	AddRecord(record RecordObject) error
+
+	// GetRecords retrieves agent records based on the provided Filters.
+	GetRecords(opts ...FilterOption) ([]RecordObject, error)
+}
```
ramizpolic: nit: i would move this to the top of this file above everything else

server/controller/store.go
adamtagscherer: I've created an issue to track it: https://github.com/agntcy/dir/issues/175

server/search/sqlite/record.go
adamtagscherer: I've created an issue to track it: https://github.com/agntcy/dir/issues/176

ramizpolic (approved):
lgtm

PR #154, opened 2025-05-30
feat: update api specs to v1alpha2

ramizpolic:
The purpose of this PR is to address API limitations by:
- decoupling `core` objects to enable their management elsewhere (specifically OASF)
- standardizing the naming convention around agent data model into `AgentRecord`
- moving the responsibility of object mapping/conversion control to the storage layer rather than across the whole API
- making the routing service more easier to use by clients and moving the responsibility to only control data flow
- enabling synchronisation of data between two (storage) services by consuming data returned from the discovery directly
- providing a way to add middleware service to control storage/routing service policies for access control

In summary, the schema management can be decoupled from ADS. If and once the `core` schema is moved to OASF, we can have our own internal compact representation of the record that can be used for Directory purposes only (e.g. in routing). Compact schema allows us to pass it between services without fully pulling the object as only several properties are needed for Dir (ie. several KBs of data). For other services, the full record may be needed, so the clients can pull the full object.

Please note that the diagram is not fully up to date, but it tries to convey the overall flow. This will be updated in the upcoming days.


api/proto/routing/v1alpha2/record.proto
```diff
@@ -0,0 +1,60 @@
...
+  string cid = 1;
+
+  // The name of the AgentRecord.
+  string name = 2;
+
+  // The version of the AgentRecord.
+  string version = 3;
+
+  // List of skills the AgentRecord has.
+  repeated string skills = 4;
```
paralta:
```suggestion
  repeated core.v1alpha2.AgentSkill = 4;
```

api/proto/routing/v1alpha2/record.proto
```diff
@@ -0,0 +1,60 @@
...
+
+package routing.v1alpha2;
+
+import "core/v1alpha2/locator.proto";
+
+// Record defines a minimal representation of an AgentRecord
+// that can be used for routing and discovery.
+// Can be constructed from an object from StoreService,
+// or it can be served by an internal database.
+// Can be verified against the actual AgentRecord.
```
paralta: I think this should not be constructed from `StoreService` if we want to keep all conversion logic inside the store service. We should construct from `AgentRecord`.

api/proto/routing/v1alpha2/routing_service.proto
```diff
@@ -0,0 +1,102 @@
...
+  repeated RecordQuery match_queries = 3;
+
+  // The score of the search match.
+  uint32 match_score = 4;
+}
+
+message ListRequest {
+  // List of queries to match against the records.
+  // All queries must match for the record to be returned.
+  repeated RecordQuery queries = 1;
```
paralta: I think we do not need `queries` here given the description above (`// List all records that this peer is currently providing.`). If the user wants to get records matching queries in this peer, they should use `Search`.

paralta (approved):
LGTM 👍  I think I would only remove all the `Agent*` prefixes from the different message names like `AgentSkill` -> `Skill`

api/proto/core/v1alpha2/record.proto
```diff
@@ -0,0 +1,72 @@
...
+  // Descriptors are used to generically extend the agent's functionality.
+  repeated AgentDescriptor descriptors = 9;
+
+  // Security signature of the agent.
+  AgentSignature signature = 10;
+
+  // Reference to the previous agent record, if any.
+  // Used to link the agent record to its previous versions.
+  // Field number is explicitly reserved for extenability.
+  optional AgentRecordRef prev_ref = 99;
```
akijakya: In OASF it is [recommended to avoid abbreviations](https://github.com/agntcy/oasf/blob/4e15528e7acd32a9e678ecfbce9e35e5160ed95a/CONTRIBUTING.md?plain=1#L61), in this case it should be called `previous_reference`.

api/proto/store/v1alpha2/object.proto
```diff
@@ -0,0 +1,38 @@
...
+// objects in a content-addressable store.
+message Object {
+  // Globally-unique content identifier of the object.
+  // Specs: https://github.com/multiformats/cid
+  string cid = 1;
+
+  // Type of the object.
+  // Can be used to identify the format or schema of the data.
+  // For example, it can be a MIME type or a custom type identifier.
+  string type = 2;
```
akijakya: Shouldn't we make this into an enum? If it refers to the type of the data inside the same object, I'd call it `data_type`. Even better, I'd call `data` `agent_data` so it is more unique and clear similar to the other `data` fields, and call the type `agent_data_type`.

api/proto/core/v1alpha2/record.proto
akijakya: I'd also consider putting this into the Object object, and make its type an `ObjectRef`. I'm not sure an Agntcy Directory-specific identifier of an agent should be part of a general agent record schema. If we want to make the reference to a previous version of the same agent part of the agent record, we could put that information into an extension/feature/descriptor, what do you think?

ramizpolic: i am not sure that will work well as moving the previous ref to a descriptor means that it is not a standard behaviour nor use-case. we do want to be able to link records between versions as it can allow us to a) announce/discover based on the latest record, and b) able to track between versions (and sync back) once we discover one.
the second part is object, which imo should remain decoupled from the record schema. the object should not be extended in any other way except to store data. other services can be added for new things without the need to change the store ops.

PR #169, opened 2025-06-11
chore: resolve v1alpha2 storage API object handling

ramizpolic:
Cleans up v1alpha2 API details and removes the handling of record data from the stored object itself. The `RecordObject` can be defined on the client-side and converted to an `Object` before being pushed to the storage layer. This ensures that we can keep adjusting our `RecordObject` without changing things on the storage interface.


muscariello (approved):
LGTM

api/proto/store/v1alpha2/object.proto
```diff
@@ -20,49 +17,38 @@ message ObjectRef {
 // objects in a content-addressable store.
 message Object {
   // Globally-unique content identifier of the object.
+  // Encodes fully-qualified type of the object as part of "codec".
   // Specs: https://github.com/multiformats/cid
   string cid = 1;
 
+  // Type of the object.
+  // Can be extracted from CID.
+  ObjectType type = 2;
```
akijakya, line 26: We already have a string `type` in OASF, please rename to `object_type` for example.

api/proto/core/v1alpha2/record.proto
```diff
@@ -25,20 +25,20 @@ import "core/v1alpha2/signature.proto";
 // All records are referenced by a globally-unique content identifier (CID).
 // Specs: https://github.com/multiformats/cid
 message Record {
```
akijakya, line 27: If we keep the `signature` attribute in the Record, please rename it to `agent_signature` for example so there is no type collision in OASF with the other signature attribute in the Signature message!

api/proto/store/v1alpha2/object.proto
ramizpolic: image

understood

ramizpolic:
after consulting with @akijakya regarding OASF, we will merge this and prepare OASF so that it can handle schema here. there are some limitations on the OASF side the way we can specify props and types, which is not ideal. we will report in the upcoming days if this can be done and how.

api/proto/core/v1alpha2/record.proto
muscariello: I think the prefix agent is bad in general. It's a record signature.

PR #170, opened 2025-06-11
docs: add signing and verification docs

ramizpolic:
Adds documentation about the signing and verification flow used for agent records


ramizpolic:
todo: add full rfc with details about problem statement and requirements

PR #177, opened 2025-06-16
feat: add unified objectmanager for API objects

ramizpolic:
This PR moves the objectmanager package as a decoupled, standalone, non-API proto interface to handle versioning and conversion between different objects passed across the API for internal purposes with typehints.

This schema can be used to provide unified control of objects, regardless of their schemas and versions.

An example for object control that leverages this behaviour via CIDs only can be found here: https://go.dev/play/p/YQ7YVrtZ9FS


tkircsi (approved):
LGTM!

tkircsi (approved):

paralta (approved):

PR #181, opened 2025-06-18
feat: multi API and API object handling

tkircsi:
This PR is in very early stage.

- Added converter package (api/converter/) for type conversions between API versions and objectmanager.RecordObject
- Version-specific controllers now handle API type conversions and delegate to common controllers
- Common controllers work exclusively with objectmanager.RecordObject types


api/converter/converter.go
ramizpolic, line 1: i think using conversion interfaces would make these more clear to use and implement, while also making things more dynamic. something like

```go
type ObjectConverter[TypeA, TypeB any] interface {
	ConvertToA(B) A
	ConvertToB(A) B
}
```

then, we can define a specific implementation (e.g. in`api/converter/v1alpha1.go` and `api/converter/v1alpha2.go`)

```go
var (
	V1A1 = NewObjectConverter[v1storetypes.Object, objectmanager.RecordObject]
	V1A2 = NewObjectConverter[v2storetypes.Object, objectmanager.RecordObject]
)
```

tkircsi: I like this idea! Thanks!

ramizpolic:
@tkircsi can we close this one since we have https://github.com/agntcy/dir/pull/225 PR that takes care of this?

ramizpolic:
done in #255

PR #184, opened 2025-06-18
fix: add login check to avoid segfault

paralta:
* Fix segfault when running the hub command to list orgs when user is not logged in.
* Plus, add login checks wherever required

Kudos to @jubarbot-cisco for finding this bug 🚀

### DIFF
--- a/hub/auth/login.go
+++ b/hub/auth/login.go
@@ -99,7 +99,7 @@ func Login(
 	return currentSession, nil
 }
 
-func IsLoggedIn(currentSession *sessionstore.HubSession) bool {
+func HasLoginCreds(currentSession *sessionstore.HubSession) bool {
 	if currentSession == nil || currentSession.AuthConfig == nil {
 		return false
 	}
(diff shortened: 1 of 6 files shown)


hub/auth/login.go
```diff
@@ -98,3 +98,20 @@ func Login(
 
 	return currentSession, nil
 }
+
+func IsLoggedIn(currentSession *sessionstore.HubSession) bool {
```
ramizpolic, line 102: we dont check here if the user is logged in, but rather if they have login creds.
maybe we rename it to something like `HasLoginCreds` or something similar

ramizpolic: we dont check here if the user is logged in, but rather if they have login creds.
maybe we rename it to something like `HasLoginCreds` or something similar

ramizpolic (changes requested):

hub/auth/login.go
paralta: good point, that is more accurate 👍

PR #186, opened 2025-06-18
feat: add local search service

adamtagscherer:
This PR wires in the newly added index DB through the controller, client and the CLI.
There is a new command `dirctl search ...` that you can use to search for agents uploaded to the local node.


api/search/v1alpha2/record_query.go
```diff
@@ -0,0 +1,30 @@
...
+func init() {
+	// Override allowed names for RecordQueryType
+	RecordQueryType_name = map[int32]string{
+		0: "unspecified",
+		1: "agent-name",
+		2: "agent-version",
+		3: "skill-id",
+		4: "skill-name",
+		5: "locator",
+		7: "extension",
```
paralta:
```suggestion
		6: "extension",
```

api/search/v1alpha2/record_query.go
```diff
@@ -0,0 +1,30 @@
...
+package searchv1alpha2
+
+const ValidQueryTypes = "agent-name,agent-version,skill-id,skill-name,locator,extension"
+
+func init() {
+	// Override allowed names for RecordQueryType
+	RecordQueryType_name = map[int32]string{
+		0: "unspecified",
+		1: "agent-name",
+		2: "agent-version",
```
ramizpolic:
```suggestion
		1: "name",
		2: "version",
```

api/search/v1alpha2/record_query.go
```diff
@@ -0,0 +1,30 @@
...
+		3: "skill-id",
+		4: "skill-name",
+		5: "locator",
+		6: "extension",
+	}
+	RecordQueryType_value = map[string]int32{
+		"":              0,
+		"unspecified":   0,
+		"agent-name":    1,
+		"agent-version": 2,
```
ramizpolic:
```suggestion
		"name":    1,
		"version": 2,
```

cli/cmd/search/search.go
```diff
@@ -0,0 +1,96 @@
...
+1. Search for records with specific filters and limit:
+
+	dirctl search --limit 10 \
+		--offset 0 \
+		--query "agent-name:my-agent-name" \
+		--query "agent-version:v1.0.0" \
+		--query "skill-id:10201" \
+		--query "skill-name:Text Completion" \
+		--query "locator:docker-image:https://example.com/docker-image" \
+		--query "extension:my-custom-extension-name:v1.0.0"
```
ramizpolic: maybe we do equal sign so we can match against operations as well (equal, not equal, larger, etc).
also id update usage like this

```suggestion
		--query "name=my-agent-name" \
		--query "version=v1.0.0" \
		--query "skill-id=10201" \
		--query "skill-name=Text Completion" \
		--query "locator=docker-image:https://example.com/docker-image" \
		--query "extension=my-custom-extension-name:v1.0.0" 
```

cli/cmd/search/search.go
```diff
@@ -0,0 +1,96 @@
...
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		return runCommand(cmd)
+	},
+}
+
+func runCommand(cmd *cobra.Command) error {
+	if opts.Limit <= 0 {
+		return errors.New("limit must be greater than 0")
+	}
+
```
ramizpolic: this check is not needed, limit is unsigned integer type

cli/cmd/search/search.go
```diff
@@ -0,0 +1,96 @@
...
+		if query[1] == "" {
+			return fmt.Errorf("query value cannot be empty for type: %s", query[0])
+		}
+
+		queries = append(queries, &searchtypesv1alpha2.RecordQuery{
+			Type:  searchtypesv1alpha2.RecordQueryType(searchtypesv1alpha2.RecordQueryType_value[query[0]]),
+			Value: query[1],
+		})
+	}
```
ramizpolic: i would move this part to a separate method where we can extract a single query from a string, for example

```go
func StringToQuery(qs string) (query, error) {

}
```

ramizpolic: we can also make the query a custom cobra value type and implement a method on it `ToQuery()`. i think this would be a better approach

https://app.studyraid.com/en/read/11421/357750/custom-flag-types

we can then implement proper validation, pattern matching, etc

client/search_v1alpha2.go
```diff
@@ -0,0 +1,40 @@
...
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+
+	searchtypes "github.com/agntcy/dir/api/search/v1alpha2"
+)
+
+func (c *Client) SearchV1alpha2(ctx context.Context, req *searchtypes.SearchRequest) (io.Reader, error) {
```
ramizpolic: i think it would be better to simply use versionless naming here (because we can always put these under version specific module under the client). and on the second part, id consider returning chans here
```suggestion
func (c *Client) Search(ctx context.Context, req *searchtypes.SearchRequest) (<-chan string, error) {
```

server/controller/v1alpha2/search.go
```diff
@@ -0,0 +1,124 @@
...
+	for _, r := range records {
+		if err := srv.Send(&searchtypes.SearchResponse{RecordCid: r.GetCID()}); err != nil {
+			return fmt.Errorf("failed to send record: %w", err)
+		}
+	}
+
+	return nil
+}
+
+func createGetRecordsParams(req *searchtypes.SearchRequest) ([]types.FilterOption, error) { //nolint:gocognit,cyclop
```
ramizpolic:
```suggestion
func queryToFilters(req *searchtypes.SearchRequest) ([]types.FilterOption, error) { //nolint:gocognit,cyclop
```

ramizpolic (commented):
overall changes and the approach is good, well done @adamtagscherer, we can extend this further as needed.

ive attached some nits and overall improvement suggestions to help us with usage and extension options.

cli/cmd/search/options.go
```diff
@@ -0,0 +1,22 @@
...
+	Query Query
+}
+
+func init() {
+	flags := Command.Flags()
+
+	flags.Uint32Var(&opts.Limit, "limit", 100, "Maximum number of results to return (default: 100)") //nolint:mnd
+	flags.Uint32Var(&opts.Offset, "offset", 0, "Pagination offset (default: 0)")
+
+	flags.Var(&opts.Query, "query", "Search query terms")
```
ramizpolic: This will not work properly, we need to do this individually via

```suggestion
	flags.StringSliceVarP(&opts.Queries, "query", "q", []string{}, "Search query terms")
```

ramizpolic: the query cobra var implementation should also be readjusted to work on a single item

context: https://github.com/spf13/cobra/issues/661

cli/cmd/search/search.go
```diff
@@ -0,0 +1,68 @@
...
+
+	var data []byte
+
+	for recordCid := range ch {
+		if recordCid == "" {
+			continue
+		}
+
+		data = append(data, []byte(recordCid)...)
```
ramizpolic: we can print here rather than collect everything into a single value

PR #189, opened 2025-06-24
feat: custom type for error handling

arpad-csepi:


utils/componenterror/componenterror.go
ramizpolic, line 1: lets rename the package to `err` rather than `componenterror` so that we can import it as `import errutils "path-to-pkg"`

ramizpolic: suggestion: rename package and files to match
`utils/err/component_error.go`
`utils/err/component_error_test.go`

utils/componenterror/componenterror.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
+)
+
+type ComponentError struct {
+	Err error `json:"-"`
+
+	Component string            `json:"component"`
+	Metadata  map[string]string `json:"metadata,omitempty"`
```
ramizpolic:
```suggestion
	Annotations  map[string]string `json:"annotations,omitempty"`
```

utils/componenterror/componenterror.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	return cErr
+}
+
+func (e *ComponentError) Error() string {
+	return e.Message
+}
+
+func (e *ComponentError) WithMetadata(metadata map[string]string) *ComponentError {
```
ramizpolic:
```suggestion
func (e *ComponentError) WithAnnotations(annotations map[string]string) *ComponentError {
```

utils/componenterror/componenterror.go
```diff
@@ -0,0 +1,78 @@
...
+
+	if err != nil {
+		cErr.Message = err.Error()
+	}
+
+	return cErr
+}
+
+func (e *ComponentError) Error() string {
+	return e.Message
```
ramizpolic: i would opt to add a TODO note here

```suggestion
	// TODO: decide how we want to format the component error message in the future, e.g. do we want to include annotations or not
	return e.Message
```

ramizpolic (approved):
lgtm

PR #192, opened 2025-06-25
ci(csit-integration): update input parameters

arpad-csepi:
The CSIT integration workflow input parameters are changed to disable all tests by default. This way every dependent are needed to explicitly enable the necessarry tests.

https://github.com/agntcy/csit/pull/122

### DIFF
--- a/.github/actions/trigger-integrations/action.yaml
+++ b/.github/actions/trigger-integrations/action.yaml
@@ -38,7 +38,7 @@ runs:
               workflow_id: 'test-integrations.yaml',
               ref: 'main',
               inputs: {
-                  skip_slim_test: true,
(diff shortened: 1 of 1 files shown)


.github/actions/trigger-integrations/action.yaml
```diff
@@ -38,7 +38,7 @@ runs:
               workflow_id: 'test-integrations.yaml',
               ref: 'main',
               inputs: {
-                  skip_slim_test: true,
+                  skip_diretory_test: false,
```
paralta:
```suggestion
                  skip_directory_test: false,
```

arpad-csepi: Oops. Thanks! 🙏

PR #198, opened 2025-06-27
feat(api): only use record objects across APIs

ramizpolic:
This PR addresses the versioning hell introduced with moving and adjusting things. Simplifies interaction logic and usage of external packages into a standardized, more stable API schema, decoupled across components.

**IMPORTANT NOTE**: We do not care about generic objects. All our interfaces (storage, search, routing) will ONLY rely on Record objects. 

## TODO

- [x] Fix APIs to use only Records (apart from `storage.v1alpha1` which will be removed later)
- [ ] Adjust code to use `core.v1.Record` rather than previous Object type
- [ ] Implement bidirectional Storage API based on the new service


ramizpolic:
refactor and fixes are part of https://github.com/agntcy/dir/pull/225 PR

PR #201, opened 2025-07-02
feat: sign and verify grpc service

### LINKED ISSUE #204: Add gRPC Client service for Sign and Verify
Currently, our signing and verification functionality is implemented directly in the Golang client at client/sign.go and client/verify.go, which limits our ability to provide consistent signing capabilities across multiple programming languages. To support our planned Python and JavaScript clients with standardized signing operations, we need to extract this functionality into a dedicated gRPC service.

* The implementation will involve creating a new {{SignService}} protobuf definition that exposes methods like {{Sign()}}, and {{Verify()}}, mirroring the existing functionality in our Golang client. 
* Once the gRPC service is implemented, we'll refactor the Golang client to use the generated stubs, ensuring all clients (Golang, Python, JavaScript) use the same underlying signing infrastructure while maintaining the same developer experience.

arpad-csepi:
Resolves #204


api/proto/sign/v1alpha2/sign_service.proto
```diff
@@ -0,0 +1,94 @@
...
+
+    // Client identity for signing with OIDC
+    string oidc_client_id = 5;
+
+    // Client token for signing with OIDC
+    string oidc_token = 6;
+
+    // Private key for signing with OIDC
+    string key = 7;
+  }
```
paralta: These `SignOpts` fields are currently required fields (not `optional`) which means that we always need to specify them when using these client sign methods. However, the client has some default values for them at https://github.com/agntcy/dir/blob/main/client/sign.go#L23 that end up not being used.

I think we can change them to `optional` here and also add comments referring the default values. If these values are not provided, then the client will use the defaults (this logic is already there for the CLI at https://github.com/agntcy/dir/blob/main/cli/cmd/sign/options.go#L25 but not at the client level).

api/proto/sign/v1alpha2/sign_service.proto
```diff
@@ -0,0 +1,94 @@
...
+import "google/protobuf/empty.proto";
+
+service SignService {
+  // Sign agent records using one of the auth methods
+  rpc SignOIDC(SignOIDCRequest) returns (SignOIDCResponse);
+  rpc SignWithKey(SignWithKeyRequest) returns (SignWithKeyResponse);
+
+  // Verify signed agent records using on of the auth methods
+  rpc VerifyOIDC(VerifyOIDCRequest) returns (google.protobuf.Empty);
+  rpc VerifyWithKey(VerifyWithKeyRequest) returns (google.protobuf.Empty);
```
paralta: It would be nice to add a short description of the two available method: OIDC-based identity signing vs traditional key-based signing.

paralta (commented):
LGTM overall, just two comments 👍

api/proto/sign/v1alpha2/sign_service.proto
arpad-csepi: Thank you for the comment. Changed the fields to optional but does not required to modify the client as the cli is ensures the value is always there (default or not)

paralta: I would still modify the client since some users will not be using the directory with cli. A good example is how the hub is using the directory

arpad-csepi: Fixed

api/proto/sign/v1alpha1/sign_service.proto
```diff
@@ -0,0 +1,97 @@
...
+    optional string oidc_provider_url = 4;
+
+    // Client identity for signing with OIDC (default value: sigstore)
+    optional string oidc_client_id = 5;
+
+    // Client token for signing with OIDC
+    string oidc_token = 6;
+
+    // Private key for signing with OIDC
+    string key = 7;
```
paralta: Can you please confirm if these 3 fields are required in the client? I think they are being passed but they are not used there.

api/proto/sign/v1alpha1/sign_service.proto
```diff
@@ -0,0 +1,88 @@
...
+  // Sign agents using keyless OIDC based provider
+  rpc SignOIDC(SignOIDCRequest) returns (SignOIDCResponse);
+  // Sign agents using PEM-encoded private key encrypted with an optional passphrase
+  rpc SignWithKey(SignWithKeyRequest) returns (SignWithKeyResponse);
+
+  // Verify signed agents using keyless OIDC based provider
+  rpc VerifyOIDC(VerifyOIDCRequest) returns (google.protobuf.Empty);
+
+  // Verify signed agents using PEM-encoded formatted PEM public key encrypted
+  rpc VerifyWithKey(VerifyWithKeyRequest) returns (google.protobuf.Empty);
```
ramizpolic: both commands can be done as part of two interfaces `Sign` and `Verify`. the messages for each can leverage proto `oneOf` to indicate what kind of signing and verification process is requested -- either OIDC or Key. IMO, this does not have to flow via separate methods, but rather as part of the requests

arpad-csepi: I did the requested change. Not sure if it is beneficial to us to combine the services to a common entry due to the actual data is wrapped in multiple layers and creates a bit more code because of that.

api/proto/sign/v1alpha1/sign_service.proto
```diff
@@ -0,0 +1,104 @@
...
+    SignWithKeyRequest key = 2;
+  }
+}
+
+message SignResponse {
+  oneof response {
+    SignOIDCResponse oidc = 1;
+    SignWithKeyResponse key = 2;
+  }
+}
```
ramizpolic:
```suggestion
message SignRequest {
  // Agent to be signed
  core.v1alpha1.Agent agent = 1;
  
  // Signing provider to use
  SignRequestProvider provider = 2;
}

message SignRequestProvider {
  oneof provider {
    SignWithOIDC oidc = 1;
    SignWithKey key = 2;
  }
}

message SignResponse {
  // Signed agent
  core.v1alpha1.Agent agent = 1;
}
```

api/proto/sign/v1alpha1/sign_service.proto
```diff
@@ -0,0 +1,104 @@
...
+
+import "core/v1alpha1/agent.proto";
+import "google/protobuf/empty.proto";
+
+service SignService {
+  // Sign agents using keyless OIDC based provider or using PEM-encoded private key with an optional passphrase
+  rpc Sign(SignRequest) returns (SignResponse);
+
+  // Verify signed agents using keyless OIDC based provider or using PEM-encoded formatted PEM public key encrypted
+  rpc Verify(VerifyRequest) returns (google.protobuf.Empty);
```
ramizpolic:
```suggestion
  rpc Verify(VerifyRequest) returns (VerifyResponse);
```

lets avoid using empty response messages and rather define our own

```go
message VerifyResponse {
	bool success = 1;
}
```

api/proto/sign/v1alpha1/sign_service.proto
ramizpolic: adding a separate message for oneof helps with future extendability

api/proto/sign/v1alpha1/sign_service.proto
```diff
@@ -0,0 +1,104 @@
...
+
+    // Timestamp authority access URL (default value: https://timestamp.sigstage.dev/api/v1/timestamp)
+    optional string timestamp_url = 3;
+
+    // OIDC provider access URL (default value: https://oauth2.sigstage.dev/auth)
+    optional string oidc_provider_url = 4;
+  }
+
+  // Agent data for signing
+  core.v1alpha1.Agent agent = 1;
```
ramizpolic: not needed here, we always pass the agent for every provider. use the updated messages captured above

api/proto/sign/v1alpha1/sign_service.proto
```diff
@@ -0,0 +1,104 @@
...
+  // Token for OIDC provider
+  string id_token = 2;
+
+  // Signing options for OIDC
+  SignOpts options = 3;
+}
+
+message SignWithKeyRequest {
+  // Agent data for signing
+  core.v1alpha1.Agent agent = 1;
```
ramizpolic: same comment as for OIDC

api/proto/sign/v1alpha1/sign_service.proto
```diff
@@ -0,0 +1,104 @@
...
+  string expected_signer = 3;
+}
+
+message VerifyWithKeyRequest {
+  // Signed agent data
+  core.v1alpha1.Agent agent = 1;
+
+  // Public key to validate the signed agent
+  bytes public_key = 2;
+}
```
ramizpolic:
```suggestion
message VerifyRequest {
  // Signed agent to be verified
  core.v1alpha1.Agent agent = 1;
  
  // Verification provider to use
  VerifyRequestProvider provider = 2;
}

message VerifyRequestProvider {
  oneof provider {
    VerifyWithOIDC oidc = 1;
    VerifyWithKey key = 2;
  }
} 
```

arpad-csepi: Done

ramizpolic (approved):
lgtm

PR #202, opened 2025-07-03
feat: new hub https://hub.agntcy.org/directory

jubarbot-cisco:
The new official location of the directory is
https://hub.agntcy.org/directory.

Old dirctl client will still work for some time because, at the moment, there is a 302 temporary HTTP redirection from https://hub.agntcy.org to https://hub.agntcy.org/directory.


paralta (approved):
LGTM 👍

PR #203, opened 2025-07-04
feat: add --insecure to dirctl hub commands

jubarbot-cisco:
Here is a proposition to add a new `--insecure` flag to disable certificate validity checks. It can be handy when developing locally.

ex: `./bin/dirctl hub orgs --server-address https://localhost --insecure`

If `--insecure`, skips tls certificate check when:
- getting config.json from the hub,
- doing gRPC connections to the hub.

idp and okta client connections are still done securely.

## Discussion
If that flag makes sense and is accepted by the maintainers, then we can even propose to move this flag "upper", it order to also have it for other non-hub subcommands

## Remark
The other (non-hub) sub commands are apparently using a insecure gRPC connection to the dir: 
https://github.com/agntcy/dir/blob/main/client/client.go#L61.

Is it the expected behavior ?

### DIFF
--- a/hub/auth/internal/webserver/server.go
+++ b/hub/auth/internal/webserver/server.go
@@ -54,8 +54,8 @@ func StartLocalServer(ctx context.Context, h *Handler, port int, errCh chan erro
 
 		var resp *http.Response
 
-		resp, err = httpUtils.CreateSecureHTTPClient().Do(req)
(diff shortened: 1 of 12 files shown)


muscariello:
certificate validation is mandatory. more so for the hub.

PR #209, opened 2025-07-08
feat: data synchronisation on-demand

paralta:
* Refactor `search` in server into `database` for broader usage
* Add sync database schema to track sync operations and status
* Add sync controller, client, and CLI commands for sync management
* Implement sync scheduler to periodically check for pending syncs and trigger worker operations
* Implement worker logic to pull records from remote directories and push them to local storage. Please note that this is a base implementation processing each record sequentially in a single pass, thus reliability improvements will be required


muscariello (changes requested):
The synch architecture is wrong AFAICT.
Let's clarify this in details.

PR #215, opened 2025-07-13
feat: remove python code generation

adamtagscherer:
This PR removes the python language stub generation from proto files from the project, because it happens in BSR already.

next PR -> remove Go code generation from the project


muscariello (approved):
LGTM

PR #216, opened 2025-07-14
feat(sdk): add sdk release process

arpad-csepi:


sdk/python/v2/client.py
```diff
@@ -0,0 +1,192 @@
...
+
+import grpc
+import core.v1alpha2 as core
+import routing.v1alpha2.routing_service_pb2_grpc as routing_services
+import routing.v1alpha2.routing_service_pb2 as routing_types
+import search.v1alpha2.search_service_pb2_grpc as search_services
+import search.v1alpha2.search_service_pb2 as search_types
+import sign.v1alpha2.sign_service_pb2_grpc as sign_services
+import sign.v1alpha2.sign_service_pb2 as sign_types
+import store.v1alpha1.store_service_pb2_grpc as store_services
```
adamtagscherer: typo

.github/workflows/reusable-release-sdk.yaml
```diff
@@ -0,0 +1,82 @@
...
+        description: "Make a python SDK release."
+        default: false
+
+jobs:
+  sdk:
+    name: SDKs
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
```
adamtagscherer: do we need to write the packages?

.github/workflows/reusable-release-sdk.yaml
```diff
@@ -0,0 +1,82 @@
...
+        shell: bash
+        run : |
+          task deps
+
+      - name: Setup Node env
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22.x'
+          registry-url: https://registry.npmjs.org/
+          scope: "@agntcy"
```
adamtagscherer, line 61: do we need this scope?

sdk/javascript/main_test.js
```diff
@@ -0,0 +1,6 @@
+import { assertEquals } from "@std/assert";
+import { add } from "./main.js";
+
+Deno.test(function addTest() {
```
adamtagscherer: Why are we using Deno here? Don't we use Node?

sdk/javascript/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/agntcy/dir.git"
+  },
+  "license": "Apache-2.0",
+  "author": "",
+  "type": "commonjs",
+  "main": "main.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
adamtagscherer: Why don't we call the test in `main_test.js` here if we have that?

adamtagscherer (changes requested):

.github/workflows/reusable-release-sdk.yaml
arpad-csepi: We don't need it anymore, deleted

.github/workflows/reusable-release-sdk.yaml
arpad-csepi: Not required, but useful. This scope based solution allowes us to use multiple registry in the future. Can be removed if desired

sdk/javascript/package.json
arpad-csepi: The `npm init` command does not populate the test script with anything. I integrated Jest testing framework to here.

sdk/javascript/main_test.js
arpad-csepi: Remained from testing Deno runtime, rewritten to be Node native and Jest compatible.

adamtagscherer (approved):

PR #218, opened 2025-07-15
Update digest and size in an agent after json marshalled/modified.

sfph:
The object received is marshalled into JSON to drop any unwanted key/value pairs.  This will change the size and digest of the object so this needs to be adjusted as well.


muscariello:
@sfph we have refactory in progress here #198 as the current implementation is a temporary fix which was done to be able to launch agent hub in time. 
I prefer you keep going using the python tooling while we complete the refactory which covers:

- storage layer 100% OCI image spec standard compatible 
- tagging 100% compliant to OCI image spec: search attributes
- synch 100% compliant to the OCI distribution spec.

This will make downstream application integrations much easier.

PR #225, opened 2025-07-19
feat(api): version upgrade of API and store

tkircsi:
This PR is a cleaned up, rebased version of PR #198  and includes some improvements to the store service layer. (annotations, tags) 
The API upgrade will continue on this PR and branch.


api/core/v1/canonical.go
```diff
@@ -0,0 +1,56 @@
...
+import (
+	"encoding/json"
+
+	"google.golang.org/protobuf/encoding/protojson"
+)
+
+// CanonicalMarshal marshals the record using canonical JSON serialization.
+// This ensures deterministic, cross-language compatible byte representation.
+// The output is used for both CID calculation and storage to maintain consistency.
+func (r *Record) CanonicalMarshal() ([]byte, error) {
```
ramizpolic: Great job here @tkircsi! I would only advise to create a matching unit test for these two functions so we can catch potential issues.

ramizpolic: nit: rename `CanonicalMarshal` and `CanonicalUnmarshal` to `MarshalCanonical` and `UnmarshalCanonical` as it indicates the operation and the way the operation is performed rather than vice-versa. Its a minor usability improvement.

api/core/v1/record.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return ""
+	}
+
+	return cidVal.String()
+}
+
+// MustGetCid is a convenience method that panics if CID calculation fails.
+// Use this only when you're certain the record is valid.
+func (r *Record) MustGetCid() string {
```
ramizpolic, line 44: i think returning an error here is safer than panicking. on some places (e.g. storage layer), it may be more useful to catch the error than to panic.

api/objects/v1/agent.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return data, nil
+}
+
+func (x *Agent) LoadFromFile(path string) ([]byte, error) {
+	reader, err := os.Open(path)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open file: %w", err)
+	}
```
ramizpolic, line 96: memory leak: the file reader is never closed after reading

cli/cmd/list/peer.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"strings"
 
-	routetypes "github.com/agntcy/dir/api/routing/v1alpha1"
+	routetypes "github.com/agntcy/dir/api/routing/v1alpha2"
```
ramizpolic, line 10: nit: there are many inconsistencies with the way we import services and types. for example, on some places we are importing as `Xv1` and on other places `Xtypes`. i would suggest we ensure that all imports are version specific, ie. `Xv1` or `Xv1alpha1`

cli/cmd/push/push.go
```diff
@@ -5,17 +5,14 @@
...
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
-	"os"
 
-	coretypes "github.com/agntcy/dir/api/core/v1alpha1"
 	"github.com/agntcy/dir/cli/presenter"
+	"github.com/agntcy/dir/cli/types"
+	agentUtils "github.com/agntcy/dir/cli/util/agent"
```
ramizpolic, line 14: i think we should fully remove the notion of agent across the solution. instead, we can have `typeutils` or `objectutil` package rather than `agentutils`

cli/cmd/sign/sign.go
```diff
@@ -71,7 +70,7 @@ func runCommand(cmd *cobra.Command, source io.ReadCloser) error {
 		return errors.New("failed to get client from context")
 	}
 
-	agent := &coretypes.Agent{}
+	agent := &objectsv1.Agent{}
```
ramizpolic, line 73: why are we still using Agent here? can we rename to `Record`?

cli/cmd/verify/verify.go
```diff
@@ -62,11 +64,23 @@ func runCommand(cmd *cobra.Command, source io.ReadCloser) error {
 		return errors.New("failed to get client from context")
 	}
 
-	agent := &coretypes.Agent{}
+	// Load OASF data (supports v1, v2, v3) into a Record
+	record, err := types.LoadOASFFromReader(source)
```
ramizpolic: i would move this to a dedicated util package so we can use it across all imports. maybe `typeutils` or `objectutils` package like the one I mentioned above.

cli/types/types.go
```diff
@@ -4,36 +4,75 @@
 package types
```
ramizpolic: I would suggest a different, a bit more user-friendly way to define OASF object handler. we can implement it via struct, and expose it as default object. for example:

```go
// file: types/oasf/handler.go

// Handler simplifies the loading of OASF objects with different versions
// into a unified Record object.
var Handler OASFHandler = &handler{}

type OASFHandler interface {
	GetRecordVersion(reader io.Reader) (string, error)
	GetRecord(reader io.Reader) (*corev1.Record, error)
}

type handler struct {}

func GetRecordVersion(reader io.Reader) (string, error) {

}

func (h *handler) GetRecord(reader io.Reader) (*corev1.Record, error) {

}
```

I would advise for this approach. Feel free to adjust the methods. The point is of this approach is that allows easy imports, extensions, and method namings like:

```go
func main() {
	version, _ := oasf.Handler.GetRecordVersion(rd)
	record, _ := oasf.Handler.GetRecord(rd)
}

```

ramizpolic: if not using this approach, then we can always move this to a fully dedicated package and the usage remains the almost the same, like

```go
import (
	oasfHandler "<src>/utils/oasf/handler"
)

func main() {
	version, _ := oasfHandler.GetRecordVersion(rd)
	record, _ := oasfHandler.GetRecord(rd)
}
```

wdyt @tkircsi

client/store.go
```diff
@@ -4,96 +4,203 @@
...
+	maxRecordSize = 4 * 1024 * 1024 // 4MB - max record size as per v1alpha2 spec
+)
+
+// Push sends a complete record to the store and returns a record reference.
+// The record must be ≤4MB as per the v1alpha2 store service specification.
+func (c *Client) Push(ctx context.Context, record *corev1.Record) (*corev1.RecordRef, error) {
+	// Validate record size (optional but recommended)
+	if size := proto.Size(record); size > maxRecordSize {
+		return nil, fmt.Errorf("record too large: %d bytes (max 4MB)", size)
+	}
```
ramizpolic: note: we can always impose this limit natively as part of the gRPC service when creating the contollers, so it can be a configurable limit on the server side that can be set via config

https://medium.com/@mofcloud/grpc-configure-data-transmit-size-f25991e75670

client/store.go
```diff
@@ -4,96 +4,203 @@
...
-// TODO: when pushing extension data, squash items and push individually before pushing agent.
-// TODO: this allows larger agent file while keeping data in extension compact.
-func (c *Client) Push(ctx context.Context, ref *coretypes.ObjectRef, reader io.Reader) (*coretypes.ObjectRef, error) {
+const (
+	maxRecordSize = 4 * 1024 * 1024 // 4MB - max record size as per v1alpha2 spec
+)
+
+// Push sends a complete record to the store and returns a record reference.
+// The record must be ≤4MB as per the v1alpha2 store service specification.
+func (c *Client) Push(ctx context.Context, record *corev1.Record) (*corev1.RecordRef, error) {
```
ramizpolic, line 15: this will create a socket connection for each client push call rather than streaming multiple items in a row. id would suggest we create a new method in another PR to add support for streamable push, like

```suggestion
func (c *Client) Push(ctx context.Context, recordCh <-chan *corev1.Record, chan<- corev1.RecordRef) error {
```

in other words, the expectation is:
- push multiple records serially in a row, one by one
- return serially in provided chan each reference
- if any of the pushes fail, fail fully the operation
- once failed, we know the last successful record that was pushed

important: @tkircsi we need to figure out a way to do this easily in Golang. lets pin this once we have capacity to work on this. similar to all other streamable storage APIs

server/store/localfs/localfs.go
ramizpolic, line 1: we can fully remove localfs

ramizpolic: we can leave it for the next PR once we start cleaning things up

server/store/oci/constants.go
```diff
@@ -0,0 +1,94 @@
...
+	// Count metadata (simple keys).
+	MetadataKeyAuthorsCount      = "authors-count"
+	MetadataKeySkillsCount       = "skills-count"
+	MetadataKeyLocatorTypesCount = "locator-types-count"
+	MetadataKeyExtensionCount    = "extension-names-count"
+
+	// Derived from MetadataKey constants to ensure consistency.
+
+	// Core Identity (derived from MetadataKey constants).
+	ManifestKeyName        = manifestDirObjectKeyPrefix + "/" + MetadataKeyName
```
ramizpolic, line 55: path.Join() does `/` joining by always using the same `/` character

ramizpolic: unlike `filepath.Join` which uses a OS-specific path character

cli/cmd/list/digest.go
ramizpolic, line 1: are we still relying on the digest here? this may be outdated

api/core/v1/record.go
tkircsi: This method is only used in `record_test.go`, maybe we can remove it. Otherwise, `GetCid` returns only string for simpler usability and empty string if CID calculation fails. There are at least two possibilities. One is that the caller checks for an empty string, and two is that functions expecting a CID must handle this.

tkircsi: Because this method is used only in tests now, let's discuss it later.

api/core/v1/canonical.go
tkircsi: Fixed

api/objects/v1/agent.go
tkircsi: Fixed

server/store/localfs/localfs.go
tkircsi: Ok

server/store/oci/constants.go
tkircsi: can't use function calls in const declarations

api/objects/v2/agent_record.go
ramizpolic, line 1: is there any reason why we renamed OASF to Objects? IMO this impacts the understanding on where these objects are coming from.

server/store/oci/internal.go
```diff
@@ -0,0 +1,213 @@
...
+// pushData pushes record data to OCI and returns the blob descriptor.
+func (s *store) pushData(ctx context.Context, record *corev1.Record) (ocispec.Descriptor, error) {
+	recordCID := record.GetCid()
+	if recordCID == "" {
+		return ocispec.Descriptor{}, status.Error(codes.InvalidArgument, "record CID is required") //nolint:wrapcheck // Mock should return exact error without wrapping
+	}
+
+	// Marshal the record using canonical JSON marshaling
+	// This ensures the stored content matches the CID calculation
+	recordBytes, err := record.MarshalCanonical()
```
ramizpolic: i understand that we can easily parse raw bytes into concrete `core.Record` but since we will always have `schema_version` property across all OASF objects, it is relatively easy to construct `core.Record` when we have the data. This is already the case with `LoadOASFObject`.

this is problematic because for example, `calcCID(core.Record) != calcCID(oasf.v1)`. the goal is that our CID is the actual CID of the OASF data, and not of DIR data.

server/store/oci/tags.go
ramizpolic, line 1: could you please include a short README file to the `server/store/oci` that includes the process of creating tags, and maybe an example of what tags are created for some example record. from usage and documentation perspective, it would be super useful because then we can understand fully what happens on the OCI level without digging through the code

cli/cmd/list/peer.go
ramizpolic: issue created for this to track separately

cli/cmd/push/push.go
ramizpolic: created an issue for this

cli/cmd/sign/sign.go
ramizpolic: created an issue for this

cli/cmd/verify/verify.go
ramizpolic: created an issue for this

cli/types/types.go
ramizpolic: created an issue for this

client/store.go
ramizpolic: created an issue for this

client/store.go
ramizpolic: i still think that we should avoid doing this as part of the controller but pass this responsibility to the gRPC service handler instead

server/store/oci/internal.go
tkircsi: Why is it necessary for others to be able to calculate the CID? Even if they know the byte array, how would they know which hash, CID version, and codec to use?

api/objects/v2/agent_record.go
tkircsi: They are stored in the buf registry in this way(https://buf.build/agntcy/oasf) and I couldn't figure out how to generate them under a different package name in buf.gen.yaml.

client/store.go
tkircsi: Fixed

server/store/oci/internal.go
muscariello: I think we agreed to move to OCI standard compliance where the cid is taken from the oci digest. This makes the cid a non user calculated quantity. Like @tkircsi  mentioned that's the point about using the standard.

server/store/oci/tags.go
tkircsi: fixed

server/store/oci/internal.go
tkircsi: 1. AFAIK OCI registries do not calculate OCI digests, they only check them. Digest calculation is the responsibility of the client.
2. It is probably possible to store only the OASF object in the OCI registry without the wrapper Record, but this slightly increases the complexity of the `MarshalCanonical` and `UnmarshalCanonical` methods, and it is necessary to examine what modifications this entails. The question is what advantages this has.

ramizpolic: Whether it is a CID or digest is irrelevant. What is important is that the hash of the data we return can be validated against the data that is pushed/pulled. This is not the case when we have some OASF object that we wrap around DIR Record because those two objects are different in terms of structure. 

We do not need to calculate the CID here, but rather digest only, and then simply convert the SHA256 digest to CID. In the operations here, this is done in the opposite way (we marshal the data, calculate CID, convert the CID to digest, and then push it that way). Rather, we should stick to the following flow:

- Extract the actual OASF object from DIR Record (get the actual v1, v2, or v3 object) - can do this easily from proto
- Canonical marshal extracted OASF data into bytes - already have the method
- Generate OCI digest for that data - there is OCI helper method that does this for us
- Push the data with valid digest
- Convert digest to CID

api/objects/v2/agent_record.go
ramizpolic: lets update this separately then, its not a issue for this PR

server/store/oci/internal.go
ramizpolic: > 1. AFAIK OCI registries do not calculate OCI digests, they only check them. Digest calculation is the responsibility of the client.
> 2. It is probably possible to store only the OASF object in the OCI registry without the wrapper Record, but this slightly increases the complexity of the `MarshalCanonical` and `UnmarshalCanonical` methods, and it is necessary to examine what modifications this entails. The question is what advantages this has.

1. True
2. User can validate the data against the hash (CID/digest/whatever), without having to understand the internals of DIR

tkircsi: > Whether it is a CID or digest is irrelevant. What is important is that the hash of the data we return can be validated against the data that is pushed/pulled. This is not the case when we have some OASF object that we wrap around DIR Record because those two objects are different in terms of structure.
> 
> We do not need to calculate the CID here, but rather digest only, and then simply convert the SHA256 digest to CID. In the operations here, this is done in the opposite way (we marshal the data, calculate CID, convert the CID to digest, and then push it that way). Rather, we should stick to the following flow:
> 
> * Extract the actual OASF object from DIR Record (get the actual v1, v2, or v3 object) - can do this easily from proto
> * Canonical marshal extracted OASF data into bytes - already have the method
> * Generate OCI digest for that data - there is OCI helper method that does this for us
> * Push the data with valid digest
> * Convert digest to CID

Okay. I can see the benefits of this. Even if Record changes in the future (v2), the OASF types in the registry can still be retrieved and used with the canonical marshal/unmarshal functions. I will check what this affects in the code.

tkircsi: fixed. please review it.

api/core/v1/canonical.go
```diff
@@ -0,0 +1,34 @@
...
+	default:
+		return nil, errors.New("unsupported record type")
+	}
+}
+
+// UnmarshalCanonical unmarshals canonical OASF object JSON bytes to a Record.
+// This function detects the OASF version from the data and constructs the appropriate Record wrapper.
+func UnmarshalCanonical(data []byte) (*Record, error) {
+	return LoadOASFFromBytes(data)
+}
```
ramizpolic: nit: i feel like both of these methods can be renamed to `MarshalOASF` and `UnmarshalOASF` to indicate that all we want to do is handle OASF native objects along with Record ops. We could also move these two methods to OASF. 

Other functions such as `LoadOASFFromReader` and others are simply internal functions for these two methods since they return a Record anyway.

server/store/oci/internal.go
```diff
@@ -0,0 +1,213 @@
...
+
+var internalLogger = logging.Logger("store/oci/internal")
+
+// pushData pushes record data to OCI and returns the blob descriptor.
+func (s *store) pushData(ctx context.Context, record *corev1.Record) (ocispec.Descriptor, error) {
+	recordCID := record.GetCid()
+	if recordCID == "" {
+		return ocispec.Descriptor{}, status.Error(codes.InvalidArgument, "record CID is required") //nolint:wrapcheck // Mock should return exact error without wrapping
+	}
+
```
ramizpolic, line 29: not needed

server/store/oci/internal.go
```diff
@@ -0,0 +1,213 @@
...
+	if err != nil {
+		return ocispec.Descriptor{}, status.Errorf(codes.Internal, "failed to marshal record: %v", err)
+	}
+
+	// Calculate digest and size for the descriptor
+	// OCI descriptors require these fields to be valid
+	ociDigest, err := getDigestFromCID(recordCID)
+	if err != nil {
+		return ocispec.Descriptor{}, status.Errorf(codes.InvalidArgument, "failed to get digest from CID: %v", err)
+	}
```
ramizpolic, line 42: not needed

server/store/oci/internal.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Create complete blob descriptor
+	blobDesc := ocispec.Descriptor{
+		MediaType:   "application/json",
+		Digest:      ociDigest,
```
ramizpolic, line 47:
```suggestion
		Digest:      ocidigest.NewDigestFromBytes(recordBytes),
```

server/store/oci/utils.go
```diff
@@ -0,0 +1,37 @@
...
+	if err != nil {
+		return "", status.Errorf(codes.Internal, "failed to decode multihash: %v", err)
+	}
+
+	if decoded.Code != mh.SHA2_256 {
+		return "", status.Errorf(codes.InvalidArgument, "unsupported hash type: 0x%x", decoded.Code)
+	}
+
+	return ocidigest.NewDigestFromBytes(ocidigest.SHA256, decoded.Digest), nil
+}
```
ramizpolic, line 33:
```suggestion
func getCIDfromDigest(digest ocidigest.Digest) (string, error) {
	// Get hex digest from as bytes
	hexDigest, err := hex.DecodeString(digest.Hex())
	if err != nil {
		return cid.Undef, fmt.Errorf("failed to decode hex: %w", err)
	}

	// Create a multihash from the SHA256 digest
	mh, err := multihash.Encode(hexDigest, multihash.SHA2_256)
	if err != nil {
		return cid.Undef, fmt.Errorf("failed to encode multihash: %w", err)
	}

	// Create a CID v1 with the multihash, raw codec, change as needed
	c := cid.NewCidV1(cid.Raw, mh)

	return c, nil
}
```

server/store/oci/internal.go
ramizpolic: i have left updated flow example below in the review comments

api/core/v1/canonical.go
tkircsi: fixed

server/store/oci/internal.go
tkircsi: this cause invalid digest. let's check it in a different pr

server/store/oci/internal.go
tkircsi: this cause invalid digest. let's check it in a different pr

server/store/oci/utils.go
tkircsi: same as above

ramizpolic: this was untested, so it would need to be checked.

PR #227, opened 2025-07-22
feat: data sync

paralta:
* Refactor search in server into database for broader usage
* Add sync database schema to track sync operations and status
* Add sync controller, client, and CLI commands for sync management
* Implement sync scheduler to periodically check for pending syncs and trigger worker operations
* Implement worker logic to leverage `zot sync extension` which copies and caches a full copy of every record from the upstream to the downstream registry, updating the cache whenever polling discovers a change in the upstream registry.


USAGE.md
```diff
@@ -241,6 +241,28 @@ dirctl list info
 dirctl list info --network
 ```
 
+### Sync
+
+The sync feature enables one-way synchronization of records and other objects between remote Directory instances and your local node. This feature supports distributed AI agent ecosystems by allowing you to replicate content from multiple remote directories, creating local mirrors for offline access, backup, and cross-network collaboration.
+
+**How Sync Works with Zot**: Directory leverages [Zot](https://zotregistry.dev/), a cloud-native OCI registry, as the underlying synchronization engine. When you create a sync operation, the system dynamically configures Zot's sync extension to pull content from remote registries. Objects are stored as OCI artifacts (manifests, blobs, and tags), enabling container-native synchronization with automatic polling, retry mechanisms, and secure credential exchange between Directory nodes.
```
ramizpolic, line 248: question: if a sync is created on DST node, it will continuously sync the data between nodes from SRC to DST? In other words, is it true that once the sync is created, it will automatically keep updating our local storage in the future?

api/core/v1alpha1/oasf-validator/skills.go
```diff
@@ -20,7 +20,7 @@ import (
 	corev1alpha1 "github.com/agntcy/dir/api/core/v1alpha1"
 )
 
-const OASFEndpoint = "https://schema.oasf.agntcy.org/0.3.1/api/skills"
+const OASFEndpoint = "https://schema.oasf.outshift.com/0.3.1/api/skills"
```
ramizpolic: TODO as a separate task: this needs to be configurable

client/sync.go
```diff
@@ -0,0 +1,70 @@
...
+		if errors.Is(err, io.EOF) {
+			break
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("failed to list syncs: %w", err)
+		}
+
+		items = append(items, item)
+	}
```
ramizpolic: we do not support pagination at the moment? if so, it would be better to do this in the background and return a channel instead of collecting all the results in a slice to allow streaming/pagination

server/controller/v1alpha2/search.go
```diff
@@ -13,29 +13,29 @@ import (
 	"github.com/agntcy/dir/utils/logging"
 )
 
-var logger = logging.Logger("controller/search")
+var searchLogger = logging.Logger("controller/search")
```
ramizpolic, line 16: nit: any reason why this was renamed? i think it can remain `logger`

ramizpolic: i see now, the problem here is that the search controller is at the wrong place. it should be:

`controller/search/v1alpha2/search.go`

and in the same way, the sync controller under

`controller/sync/v1alpha1/sync.go`

paralta: I thought the search and sync controllers could even be moved to `controller/search.go` and `controller/sync.go` after the big refactor

api/core/v1alpha1/oasf-validator/skills.go
paralta: this small fix was required to make the e2e tests pass

USAGE.md
paralta: Yes, https://zotregistry.dev/v2.0.4/articles/mirroring/#mirroring-modes the periodic sync mode is required to have the full mirror of the DST node. You can stop this by deleting the sync on SRC though

api/core/v1alpha1/oasf-validator/skills.go
ramizpolic: of course, im just outlining it here so we are aware of this

USAGE.md
ramizpolic: i would suggest we include the summary of these comments as its super important to understand how the sync behaves after creation

paralta: would this explanation fit in `USAGE.md`? I feel like the "How Sync Works with Zot" is already heavy for this file

client/sync.go
paralta: fixed! thanks for catching this 👍

USAGE.md
ramizpolic: i think we can include one more sentence here because it does related to the sync feature usage

server/controller/v1alpha2/sync.go
```diff
@@ -0,0 +1,185 @@
...
+
+	return &storetypes.CreateSyncResponse{
+		SyncId: id,
+	}, nil
+}
+
+func (c *syncCtlr) ListSyncs(req *storetypes.ListSyncsRequest, srv storetypes.SyncService_ListSyncsServer) error {
+	syncLogger.Debug("Called sync controller's ListSyncs method", "req", req)
+
+	syncs, err := c.db.GetSyncs()
```
ramizpolic: we are not making use of pagination here although this seems to be supported

server/controller/v1alpha2/sync.go
```diff
@@ -0,0 +1,185 @@
...
+			SyncId:             sync.GetID(),
+			RemoteDirectoryUrl: sync.GetRemoteDirectoryURL(),
+			Status:             sync.GetStatus(),
+		}
+
+		syncLogger.Debug("Sending sync object", "sync_id", sync.GetID(), "status", sync.GetStatus())
+
+		if err := srv.Send(resp); err != nil {
+			return fmt.Errorf("failed to send sync object: %w", err)
+		}
```
ramizpolic: nit

```suggestion
		syncLogger.Debug("Sending sync object", "sync_id", sync.GetID(), "status", sync.GetStatus())

		if err := srv.Send(&storetypes.ListSyncsItem{
			SyncId:             sync.GetID(),
			RemoteDirectoryUrl: sync.GetRemoteDirectoryURL(),
			Status:             sync.GetStatus(),
		}); err != nil {
			return fmt.Errorf("failed to send sync object: %w", err)
		}
```

USAGE.md
paralta:
```bash
# Create a sync operation to start periodic poll from remote
dirctl sync create https://remote-directory.example.com:8888

# List all sync operations
dirctl sync list

# Check the status of a specific sync operation
dirctl sync status <sync id>

# Delete a sync operation to stop periodic poll from remote
dirctl sync delete <sync id>
```

yesterday, I edited the command comments below to mention this "periodic poll"

server/sync/scheduler.go
```diff
@@ -0,0 +1,146 @@
...
+
+// Scheduler monitors the database for pending sync operations.
+type Scheduler struct {
+	db        types.DatabaseAPI
+	workQueue chan<- synctypes.WorkItem
+	interval  time.Duration
+}
+
+// NewScheduler creates a new scheduler instance.
+func NewScheduler(db types.DatabaseAPI, workQueue chan<- synctypes.WorkItem, interval time.Duration) *Scheduler {
```
ramizpolic: can we pass only the sync database interface rather than full db here? would help reduce the db surface access

server/sync/worker.go
```diff
@@ -0,0 +1,386 @@
...
+	case synctypes.WorkItemTypeSyncCreate:
+		// TODO Check if store is oci and zot. If not, fail
+		err := w.addSync(workCtx, item)
+		if err != nil {
+			logger.Error("Sync failed", "worker_id", w.id, "sync_id", item.SyncID, "error", err)
+
+			finalStatus = storetypes.SyncStatus_SYNC_STATUS_FAILED
+		} else {
+			finalStatus = storetypes.SyncStatus_SYNC_STATUS_IN_PROGRESS
+		}
```
ramizpolic: nit: default final status on success rather than else

```suggestion
		finalStatus = storetypes.SyncStatus_SYNC_STATUS_IN_PROGRESS
		err := w.addSync(workCtx, item)
		if err != nil {
			logger.Error("Sync failed", "worker_id", w.id, "sync_id", item.SyncID, "error", err)

			finalStatus = storetypes.SyncStatus_SYNC_STATUS_FAILED
		}
```

server/sync/worker.go
```diff
@@ -0,0 +1,386 @@
...
+	if _, err := url.Parse(rawURL); err != nil {
+		return "", fmt.Errorf("invalid URL format: %w", err)
+	}
+
+	return rawURL, nil
+}
+
+func Ptr[T any](v T) *T {
+	return &v
+}
```
ramizpolic: nit: can be a private `toPtr` function

server/controller/v1alpha2/sync.go
paralta: fixed 👍

ramizpolic (approved):
LGTM

PR #240, opened 2025-07-28
feat: streaming store client

tkircsi:
Implement channel-based streaming architecture following established Go concurrency patterns. This introduces self-contained streaming functions that manage their own goroutine lifecycle while maintaining full backward compatibility.

Technical Approach
- Self-Contained Functions: Created streaming package with generator pattern functions (PushStream, PullStream, LookupStream, DeleteStream)
- Context-Aware: Proper cancellation handling with select statements
- Single Stream Reuse: One gRPC stream handles multiple records per operation
- Progressive API: Simple → Batch → Streaming methods for each operation

### DIFF
--- a/client/store.go
+++ b/client/store.go
@@ -122,6 +122,7 @@ func (c *Client) PushBatch(ctx context.Context, records []*corev1.Record) ([]*co
 	recordChan := make(chan *corev1.Record, len(records))
 	go func() {
 		defer close(recordChan)
+
 		for _, record := range records {
 			recordChan <- record
 		}
@@ -131,12 +132,14 @@ func (c *Client) PushBatch(ctx context.Context, records []*corev1.Record) ([]*co
 	results := c.PushStream(ctx, recordChan)
 
 	var refs []*corev1.RecordRef
+
 	var firstError error
 
 	for result := range results {
 		if result.Error != nil && firstError == nil {
 			firstError = result.Error
 		}
+
 		if result.RecordRef != nil {
 			refs = append(refs, result.RecordRef)
 		}
@@ -155,6 +158,7 @@ func (c *Client) PullBatch(ctx context.Context, recordRefs []*corev1.RecordRef) 
 	refChan := make(chan *corev1.RecordRef, len(recordRefs))
 	go func() {
 		defer close(refChan)
+
 		for _, ref := range recordRefs {
 			refChan <- ref
 		}
@@ -164,12 +168,14 @@ func (c *Client) PullBatch(ctx context.Context, recordRefs []*corev1.RecordRef) 
 	results := c.PullStream(ctx, refChan)
 
 	var records []*corev1.Record
+
 	var firstError error
 
 	for result := range results {
 		if result.Error != nil && firstError == nil {
 			firstError = result.Error
 		}
+
 		if result.Record != nil {
 			records = append(records, result.Record)
 		}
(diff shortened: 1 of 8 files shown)


client/store.go
```diff
@@ -5,195 +5,246 @@ package client
...
+type PullResult = streaming.PullResult
+
+// LookupResult represents the result of a lookup operation.
+// This is an alias to the streaming package's LookupResult for clean API exposure.
+type LookupResult = streaming.LookupResult
+
+// DeleteResult represents the result of a delete operation.
+// This is an alias to the streaming package's DeleteResult for clean API exposure.
+type DeleteResult = streaming.DeleteResult
+
```
ramizpolic, line 33: nit: not sure if we need these because the current client methods are too coupled in one place.

perhaps extending the client across packages would be better for decoupling, something like creating a dedicated package for service-based methods, e.g. `client/store`, `client/search`, etc.

the main client can then simply be exposed as a collection of sub-client methods, `type Client interface { StoreClient, SearchClient,...}`

client/store.go
```diff
@@ -5,195 +5,246 @@ package client
...
-	var refs []*corev1.RecordRef //nolint:prealloc // We don't know the number of records in advance
+	var firstError error
 
-	for i := range records {
-		recordRef, err := stream.Recv()
-		if err != nil {
-			return nil, fmt.Errorf("failed to receive record ref %d: %w", i, err)
+	for result := range results {
+		if result.Error != nil && firstError == nil {
+			firstError = result.Error
```
ramizpolic, line 182: the Store API docs indicate that if an error occurs, no more records will be returned, hence we can return here

client/streaming/README.md
ramizpolic, line 1: i think the implementation may be a bit too complex since all methods follow the same pattern. IMO this can be solved using one interface that works the same way for streaming store APIs, ie.

```go
type OpResult[OutT any] struct {
	Result OutT
	Err error
}

type StreamingOp [ClientT, InT, OutT any] interface {
	Client() ClientT
	InCh() <-chan InT
	OutCh() <-chan OpResult[OutT]
}
```

let me spend some more time reviewing the streaming API implementation and we can merge.
although, i have a hunch that a bit of abstraction would help a bit here for reusability

tkircsi: Sometimes explicit repetition is more valuable than clever abstraction.

client/store.go
tkircsi: I don't think we need that. In that case the `results` channel will be closed and exits from `for`. I double check the docs and code.

tkircsi: It works. We would like to improve StoreAPI so that it always processes requests arriving in the stream and does not exit at the first error.

client/streaming/pull.go
```diff
@@ -0,0 +1,125 @@
...
+
+	corev1 "github.com/agntcy/dir/api/core/v1"
+	storetypes "github.com/agntcy/dir/api/store/v1alpha2"
+)
+
+// PullResult represents the result of a pull operation.
+type PullResult struct {
+	Record *corev1.Record
+	Error  error
+	Index  int // For correlating with input order if needed
```
ramizpolic, line 21: Instead of order/index, it would be better if this was reference/input param instead

```suggestion
	RecordRef  *corev1.RecordRef // For correlating with input order if needed
```

tkircsi: I agree

PR #248, opened 2025-07-29
feat(sdk/javascript): add grpc client code and example

arpad-csepi:


sdk/javascript/v2/client.js
```diff
@@ -41,40 +41,161 @@ class Client {
...
+        });
     }
 
-    verify(req) {
-        throw new Error('Not implemented');
+    delete(ref) {
+        return new Promise((resolve, reject) => {
+            const call = this.storeClient.delete((error, _) => {
+                if (error.code !== 12) {
+                    console.error('Error form server:', error)
```
adamtagscherer: form -> from

PR #250, opened 2025-07-29
feat: migrate record signature to OCI native signature

paralta:
* Refactor `Sign` and `Verify` to use new `Record`
* Add new `PushWithOptions` and `PullWithOptions` RPCs to store service API to support OCI native signature
* Support new RPCs in server by pushing/pulling the new Signature artifacts to/from the OCI registry
* Add e2e test for signature flow

Please note that for now we are implementing the verification flow within the Directory but the long term plan is to leverage `Zot`'s signature verification https://zotregistry.dev/v2.0.4/articles/verifying-signatures/#what-is-needed-for-verifying-signatures


server/controller/store.go
```diff
@@ -217,3 +185,235 @@ func (s storeCtrl) Delete(stream storev1.StoreService_DeleteServer) error {
...
+			return err
+		}
+
+		pushedRef, err := s.pushRecordToStore(stream.Context(), request.GetRecord(), recordCID)
+		if err != nil {
+			return err
+		}
+
+		// Handle signature artifact if provided
+		if err := s.pushSignatureToStore(stream.Context(), request.GetOptions().GetSignature(), recordCID); err != nil {
```
tkircsi: I think, there is a possible nil pointer dereference of `request.GetOptions().GetSignature()`. We should verify GetOptions() return value.

server/controller/store.go
```diff
@@ -217,3 +185,235 @@ func (s storeCtrl) Delete(stream storev1.StoreService_DeleteServer) error {
...
+		}
+
+		// Pull the record
+		record, err := s.pullRecordFromStore(stream.Context(), recordRef)
+		if err != nil {
+			return err
+		}
+
+		// Try to get signature artifact if requested and store supports it
+		signature, err := s.pullSignatureFromStore(stream.Context(), recordRef, options.GetIncludeSignature())
```
tkircsi: The same possible nil pointer deref here `options.GetIncludeSignature()`

server/controller/store.go
```diff
@@ -217,3 +185,235 @@ func (s storeCtrl) Delete(stream storev1.StoreService_DeleteServer) error {
...
+		return "", status.Error(codes.Internal, "failed to calculate record CID")
+	}
+
+	storeLogger.Debug("CID calculated successfully", "cid", recordCID)
+
+	return recordCID, nil
+}
+
+// pushRecordToStore pushes a record to the store and adds it to the search index.
+func (s storeCtrl) pushRecordToStore(ctx context.Context, record *corev1.Record, recordCID string) (*corev1.RecordRef, error) {
```
tkircsi: Do we need the `recordCID` param only for logging reasons?

server/store/oci/oci.go
```diff
@@ -26,6 +31,18 @@ import (
...
+const (
+	// SignatureArtifactMediaType is the media type for signature artifacts.
+	SignatureArtifactMediaType = "application/vnd.dev.cosign.artifact.sig.v1+json"
+
+	// SignatureManifestMediaType is the media type for signature manifests.
+	SignatureManifestMediaType = "application/vnd.oci.image.manifest.v1+json"
+
+	// Signature annotations.
+	SignatureTypeAnnotation = "org.opencontainers.artifact.type"
+)
```
tkircsi, line 42: Can we put this into `constants.go`?

server/store/oci/oci.go
```diff
@@ -319,3 +336,256 @@ func (s *store) Delete(ctx context.Context, ref *corev1.RecordRef) error {
...
+
+	blobDesc := ocispec.Descriptor{
+		MediaType: mediaType,
+		Digest:    digest.FromBytes(signatureJSON),
+		Size:      int64(len(signatureJSON)),
+	}
+
+	// Push the signature blob
+	err = s.repo.Push(ctx, blobDesc, strings.NewReader(string(signatureJSON)))
+	if err != nil && !strings.Contains(err.Error(), "already exists") {
```
tkircsi:
```suggestion
		// Use oras.PushBytes for consistency with Push method
	blobDesc, err := oras.PushBytes(ctx, s.repo, mediaType, signatureJSON)
	if err != nil {
```

server/controller/store.go
paralta: good point, thanks for catching this

server/store/oci/oci.go
paralta: as their usage is restricted to this file, I would like to keep them here

server/store/oci/oci.go
paralta: neat 👍

server/store/oci/oci.go
tkircsi: Ok. We can move it later if needed.

server/controller/store.go
```diff
@@ -217,3 +185,235 @@ func (s storeCtrl) Delete(stream storev1.StoreService_DeleteServer) error {
...
+
+// pullRecordFromStore pulls a record from the store with validation.
+func (s storeCtrl) pullRecordFromStore(ctx context.Context, recordRef *corev1.RecordRef) (*corev1.Record, error) {
+	// Check if record exists
+	_, err := s.store.Lookup(ctx, recordRef)
+	if err != nil {
+		st := status.Convert(err)
+
+		return nil, status.Errorf(st.Code(), "failed to lookup record: %s", st.Message())
+	}
```
tkircsi: I'm wondering if we really need Lookup here. Both oci.Pull and oci.Lookup call the `fetchAndParseManifest` method, where we already do a Resolve.
server/store/oci/internal.go
```go
// Resolve manifest from remote tag (this also checks existence and validates CID format)
    
if err != nil {
        internalLogger.Debug("Failed to resolve manifest", "cid", cid, "error", err)
        return nil, nil, status.Errorf(codes.NotFound, "record not found: %s", cid)
    }
```
What do you think?

tkircsi (approved):
Just one more comment. Otherwise LGTM.

server/controller/store.go
paralta: you are absolutely right 👍 the `Pull` method in oci.go already looks up, thanks for catching this

PR #285, opened 2025-08-11
feat: Add distributed label-based announce/discovery via DHT

tkircsi:
### New Distributed Routing Features

• **Custom DHT Validators**: Implemented `SkillValidator`, `DomainValidator`, and `FeatureValidator` with `record.NamespacedValidator` to enable custom key-value storage in libp2p DHT for `/skills/`, `/domains/`, and `/features/` namespaces

• **Label-Based DHT Announcements**: Added `announceLabelToDHT()` functionality that announces skill/domain/feature mappings (e.g., `/skills/golang/CID123`) to the DHT network using `DHT().PutValue()` for distributed discovery

• **Dual Announcement Handling**: Created `handleLabelAnnouncement()` and `handleCIDProviderAnnouncement()` to distinguish between label mappings and traditional CID provider announcements, with proper routing based on key prefixes

• **DHT Record TTL Management**: Introduced 48-hour DHT record TTL (`DHTRecordTTL`) with coordinated 36-hour republishing interval (`LabelRepublishInterval`) to prevent label mappings from expiring in the distributed network

• **Label Republishing System**: Implemented `startLabelRepublishTask()` background process that periodically re-announces local label mappings to maintain their presence in the DHT network beyond the default 24-36 hour expiration

• **Network-Aware Dual Storage**: Developed intelligent storage strategy where labels are stored both locally (for fast queries) and announced to DHT network (for distributed discovery), with automatic fallback to local-only when no peers available

• **Enhanced Handler Notifications**: Extended `handlerSync` struct with `AnnouncementType` and `LabelKey` fields to support both traditional CID announcements and new label-based announcements with proper event routing

• **Validator Key Format Enforcement**: Added `MinLabelKeyParts` validation ensuring label keys follow `/namespace/path/CID` format with proper CID validation for both key components and values


server/routing/handler.go
```diff
@@ -27,8 +28,10 @@ type handler struct {
 }
 
 type handlerSync struct {
-	Ref  *corev1.RecordRef
-	Peer peer.AddrInfo
+	Ref              *corev1.RecordRef
+	Peer             peer.AddrInfo
+	AnnouncementType string // "CID" or "LABEL"
+	LabelKey         string // For label announcements like "/skills/golang/CID1", "/domains/web/CID2"
 }
```
paralta:
```suggestion
	AnnouncementType AnnouncementType
	LabelKey         string // For label announcements like "/skills/golang/CID1", "/domains/web/CID2"
}

type AnnouncementType string

const (
    AnnouncementTypeCID   AnnouncementType = "CID"
    AnnouncementTypeLabel AnnouncementType = "LABEL"
)

```
maybe a new type here?

server/routing/routing.go
```diff
@@ -69,7 +82,8 @@ func (r *route) List(ctx context.Context, req *routingv1.ListRequest) (<-chan *r
 	// Use remote routing when:
 	// 1. Looking for a specific record (cid) - to find providers across the network
 	// 2. MaxHops is set - indicates network traversal
-	if req.GetLegacyListRequest().GetRef() != nil || req.GetLegacyListRequest().GetMaxHops() > 0 {
+	// But only if we have peers available for network operations
+	if (req.GetLegacyListRequest().GetRef() != nil || req.GetLegacyListRequest().GetMaxHops() > 0) && r.hasPeersInRoutingTable() {
```
paralta, line 98: could this check be just `if r.hasPeersInRoutingTable()`?

server/routing/routing_remote.go
```diff
@@ -290,57 +289,177 @@ func (r *routeRemote) handleNotify(ctx context.Context) {
...
+	// TODO: we should subscribe to some records so we can create a local copy
+	// of the record and its skills.
+	// for now, we are only testing if we can reach out and fetch it from the
+	// broadcasting node
+
+	// FRAUD DETECTION: Validate that the announcing peer actually has the content
+	// Step 1: Try to lookup metadata from the announcing peer
+	_, err = r.service.Lookup(ctx, notif.Peer.ID, notif.Ref)
+	if err != nil {
+		remoteLogger.Error("FRAUD DETECTED: Peer announced CID but failed metadata lookup",
```
paralta: `FRAUD DETECTED:` this string is inconsistent with the other log messages in the codebase. I think we could remove and report the clean and simple error

paralta (approved):
LGTM 👍 just a couple of suggestions

server/routing/constants.go
```diff
@@ -0,0 +1,60 @@
...
+	ProviderRecordTTL = 48 * time.Hour
+
+	// RefreshInterval defines how often DHT routing tables are refreshed.
+	// This is a shorter interval for maintaining network connectivity.
+	RefreshInterval = 30 * time.Second
+
+	// TestRefreshInterval defines a faster refresh interval for testing.
+	// This makes tests run faster by reducing wait times for DHT operations.
+	TestRefreshInterval = 1 * time.Second
+)
```
ramizpolic: we could reduce consts here to handle all the DHT-mgmt options centrally (e.g. by using DHTRecordTTL for both labels and providers, same with RepublishInterval)

I suggest removing the TestRefreshInterval and setting/overriding the default value via config during tests. I dont think we should pull the testing data directly into the code.

```suggestion
const (
	// RecordTTL defines how long DHT records persist before expiring.
	// This is configured via dht.MaxRecordAge() and affects all PutValue operations.
	// Default DHT TTL is 36h, but we use 48h for better network resilience.
	RecordTTL = 48 * time.Hour

	// RepublishInterval defines how often we republish DHT records to prevent expiration.
	// This should be significantly less than DHTRecordTTL to ensure records don't expire.
	// We use 36h (75% of DHTRecordTTL) to provide a safe margin for network delays.
	RepublishInterval = 36 * time.Hour

	// CleanupInterval defines how often we clean up stale announcements.
	// This should match DHTRecordTTL to stay consistent with DHT behavior and prevent
	// our local cache from having stale entries that no longer exist in the DHT.
	CleanupInterval = 48 * time.Hour

	// RefreshInterval defines how often DHT routing tables are refreshed.
	// This is a shorter interval for maintaining network connectivity.
	RefreshInterval = 30 * time.Second
)
```

server/routing/handler.go
ramizpolic: in addition, we can also extend with additional type for label-type handling

```go

// AnnouncementType defines types of announcements
type AnnouncementType string

const (
    AnnouncementTypeCID   AnnouncementType = "CID"
    AnnouncementTypeLabel AnnouncementType = "LABEL"
)

// LabelType defines types of AnnouncementTypeLabel
type LabelAnnouncementType string

const {
	LabelAnnouncementTypeDomain LabelAnnouncementType = "LABEL"
	// etc
}

```

server/routing/handler.go
```diff
@@ -56,16 +59,57 @@ func (h *handler) GetProviders(ctx context.Context, key []byte) ([]peer.AddrInfo
...
 
 		return nil
 	}
 
+	// Route to appropriate handler based on key type
+	if strings.HasPrefix(keyStr, "/skills/") ||
+		strings.HasPrefix(keyStr, "/domains/") ||
+		strings.HasPrefix(keyStr, "/features/") {
+		return h.handleLabelAnnouncement(ctx, keyStr, prov)
+	}
```
ramizpolic: with label-type announcement type, we can compare directly against supported types like

```go

func (lt LabelAnnouncementType) IsValid(typ string) bool {

}
```

server/routing/routing.go
```diff
@@ -47,19 +57,22 @@ func New(ctx context.Context, store types.StoreAPI, opts types.APIOptions) (type
...
+	// Only publish to network if peers are available
+	if r.hasPeersInRoutingTable() {
+		err = r.remote.Publish(ctx, ref, record)
+		if err != nil {
+			st := status.Convert(err)
 
-		return status.Errorf(st.Code(), "failed to publish to the network: %s", st.Message())
+			return status.Errorf(st.Code(), "failed to publish to the network: %s", st.Message())
+		}
 	}
```
ramizpolic, line 88: we dont necessarily have to have any peers to enable publication. once peers connect, they can negotiate if they want to sync their routing tables to check availability

server/routing/routing_remote.go
```diff
@@ -73,10 +77,20 @@ func newRemote(ctx context.Context,
 					return nil, fmt.Errorf("failed to create provider manager: %w", err)
 				}
 
+				// create custom validators for label namespaces
+				labelValidators := CreateLabelValidators()
+				validator := record.NamespacedValidator{
+					"skills":   labelValidators["skills"],
+					"domains":  labelValidators["domains"],
+					"features": labelValidators["features"],
+				}
```
ramizpolic: consider creating a new package for validators, like `server/routing/validators` with an explicit methods/functions for each validator type

server/routing/routing.go
```diff
@@ -69,7 +82,8 @@ func (r *route) List(ctx context.Context, req *routingv1.ListRequest) (<-chan *r
 	// Use remote routing when:
```
ramizpolic, line 94: List method is not implemented properly at the moment. We should do this separately since the List needs to:
a) Query each requested label/query individually and wait via channel for responses
b) Construct a response object with `peerLabels = map[string][]string` that indicates `peer -> available labels`
c) Filter the response object where each item `len(peerLabels[peer]) > minLabelMatch`
d) Return the data via channel

server/routing/constants.go
```diff
@@ -0,0 +1,60 @@
...
+	ProtocolPrefix = "dir"
+
+	// ProtocolRendezvous is the rendezvous string used for peer discovery.
+	ProtocolRendezvous = "dir/connect"
+)
+
+// Validation rules and limits.
+const (
+	// MaxHops defines the maximum number of hops allowed in distributed queries.
+	MaxHops = 20
```
ramizpolic, line 40: we wont be needing any hops in the network search since the DHT will perform this out-of-box

server/routing/routing_local.go
```diff
@@ -79,12 +79,14 @@ func (r *routeLocal) Publish(ctx context.Context, ref *corev1.RecordRef, record
...
+	// Update metrics for all record labels and store them locally for queries
+	// Note: This handles ALL local storage for both local-only and network scenarios
+	// Network announcements are handled separately by routing_remote when peers are available
 	labels := getLabels(record)
 	for _, label := range labels {
-		// Add key with cid
-		labelKey := fmt.Sprintf("%s/%s", label, ref.GetCid())
-		if err := batch.Put(ctx, datastore.NewKey(labelKey), nil); err != nil {
+		// Store label key for local List queries
+		labelKey := datastore.NewKey(fmt.Sprintf("%s/%s", label, ref.GetCid()))
```
ramizpolic, line 105: consider extending the current datastore interface with custom methods for label reads/writes so that this type of logic is handled in one place and we can simply reuse it in the routing

server/routing/constants.go
tkircsi: It will be fixed with the List refactor when LegacyListRequest will be removed from ListRequest.

server/routing/routing.go
tkircsi: AFAIK  `DHT().PutValue()` fails if the DHT routing table is empty.

server/routing/routing.go
tkircsi: Yes, when the List refactor is ready and LegacyListRequest is removed from ListRequest.

server/routing/routing_local.go
tkircsi: Could you explain that?

server/routing/routing.go
tkircsi: You're right. List will be refactored in separate PR.

server/routing/routing_local.go
ramizpolic: example of extending https://github.com/agntcy/dir/blob/72c985f65ec3fbb2ab6d38cf56715a6d592d8d5d/server/types/datastore.go

```go
type Datastore interface {
	datastore.Batching

	WriteLabel(labelType, value)
	ReadLabel(labelType, value)
}
```

tkircsi: I think we need to reconsider this. Write/Read wouldn't be enough, because we're using the key for `Has, Put, Get and Delete`. The other part is sometimes we're using it with `batch` and sometimes with the `datastore` itself.

ramizpolic: my point was: it would be good to have reusable methods that read/write/perform ops for labels using the new explicit label-type rather than strings wherever possible. for example

```go
Write("/labels/skill/skill-ID") vs Write(labelSkill, "skill-id")
```

PR #286, opened 2025-08-11
feat: implement server-side verification with zot

paralta:
* Migrated verification from client-side to server-side
* Added `PushSignature` and `UploadPublicKey` RPCs to `SignService`, updated `StoreService` for record reference handling
* Integrated cosign and Zot for OCI-compliant signatures
* Updated verify command to use server-side verification only, enhanced push/sign commands with key-based and OIDC signing options
* Refactored from file-based to record reference operations
* Implemented public key upload to Zot registry for signature verification infrastructure
* Updated store operations to handle signature artifacts and improved integration with Zot registry
* Updated testing and infrastructure - modified E2E tests for server-side flow, added Zot configuration, enhanced Docker builds with verification tools


Taskfile.yml
```diff
@@ -1011,6 +1026,20 @@ tasks:
...
+    desc: Install cosign
+    internal: true
+    vars:
+      ARCH_TYPE: '{{ if eq ARCH "amd64" }}amd64{{ else }}{{ARCH}}{{ end }}'
+      BINARY_NAME: 'cosign-{{OS}}-{{.ARCH_TYPE}}'
+    cmds:
+      - cmd: echo "Downloading Cosign v{{.COSIGN_VERSION}}..."
+      - cmd: curl -O -L "https://github.com/sigstore/cosign/releases/download/v{{.COSIGN_VERSION}}/{{.BINARY_NAME}}"
+      - cmd: sudo mv {{.BINARY_NAME}} /usr/local/bin/cosign
+      - cmd: sudo chmod +x /usr/local/bin/cosign
```
ramizpolic: nits: cleanup and removal of sudo
TODO: please try to find a way to make this work without adding this to user installation

```suggestion
      - echo "Downloading Cosign v{{.COSIGN_VERSION}}..."
      - curl -O -L "https://github.com/sigstore/cosign/releases/download/v{{.COSIGN_VERSION}}/{{.BINARY_NAME}}"
      - mv {{.BINARY_NAME}} /usr/local/bin/cosign
      - chmod +x /usr/local/bin/cosign
```

USAGE.md
```diff
@@ -85,48 +85,29 @@ The verification process validates the agent signature against the identity prov
...
-   --oidc-issuer "(.*)github.com(.*)"
-
-## Replace the base agent model with a signed one
-rm -rf model.json
-mv signed.model.json model.json
+## Push record with signature
+dirctl push record.json --sign
+
+## Verify record
+dirctl verify <record-cid>
```
ramizpolic: are we not capable of verifying anymore who the user is or what the signing method was? IMO this prevents us from answering the questions like: "is this record coming from this user with this key/identity" or "was this record SAFELY signed (like when using keyless OIDC signing)"

ramizpolic: both of these questions are important to maintain trust schema when dealing/consuming specific records

api/proto/sign/v1/sign_service.proto
```diff
@@ -70,37 +73,8 @@ message SignResponse {
...
 message VerifyRequest {
-  // Record to be verified
-  core.v1.Record record = 1;
-
-  // Signature to verify against the record
-  Signature signature = 2;
-
-  // Verification provider to use
-  VerifyRequestProvider provider = 3;
-}
```
ramizpolic, line 81: same concerns as above

Taskfile.yml
paralta: an alternative would be to add cosign as a requirement for running cosign locally. this is not a requirement when running other installation types like kubernetes.

api/proto/sign/v1/sign_service.proto
```diff
@@ -11,13 +11,16 @@ service SignService {
   // Sign record using keyless OIDC based provider or using PEM-encoded private key with an optional passphrase
   rpc Sign(SignRequest) returns (SignResponse);
 
+  // Push a pre-generated signature to storage for a record
+  rpc PushSignature(PushSignatureRequest) returns (PushSignatureResponse);
```
ramizpolic: IMO this may not be the best place to handle storage ops. Maybe this should be moved to the store API

client/sign.go
```diff
@@ -40,196 +22,167 @@ type SignOpts struct {
 	OIDCClientID    string
 	OIDCToken       string
 	Key             string
+	RegistryAddress string
+	RepositoryName  string
```
ramizpolic: why do we need to expose this from the client side? can this not be filled automatically using cosign on the backend via package? im not sure we want to expose this on the northbound API

USAGE.md
paralta: The verify method is similar to the previous one, we are just moving it to the server. Zot is still leveraging the `cosign` features to verify the signatures

client/sign.go
paralta: we need to sign the record on the client-side (with the private key and with the interactive OIDC) and for that we need the `payload` which contains a `docker-reference` where we need these fields for Zot verification process

api/proto/sign/v1/sign_service.proto
paralta: I added this here because it's a signature-specific operation that handles not only the OCI signature artifact but also zot verification setup (public key upload), while StoreService is focused on record storage operations

api/proto/store/v1/store_service.proto
```diff
@@ -37,53 +37,45 @@ service StoreService {
...
-message PullWithOptionsRequest {
-  // Record reference to be fetched
-  core.v1.RecordRef record_ref = 1;
-
-  // Pull options containing optional flags
-  PullOptions options = 2;
+  // Pull referrer options
+  oneof options {
+    // Signature to be fetched as a referrer for the record
+    sign.v1.Signature signature = 2;
```
ramizpolic: bool pull_signature = 2;

api/proto/sign/v1/sign_service.proto
```diff
@@ -70,37 +70,8 @@ message SignResponse {
...
-
-  // Expected signer in the signed record
-  string expected_signer = 2;
-}
-
-message VerifyWithKey {
-  // Public key to validate the signed record
-  bytes public_key = 1;
+  // Record reference to be verified
+  core.v1.RecordRef record_ref = 1;
```
ramizpolic, line 74: is it feasible if we keep the on-demand verification of the identity rather than trustiness-only of the signature, e.g.

```suggestion
  // Record reference to be verified
  core.v1.RecordRef record_ref = 1;
  
  // Optional verification provider to use
  optional VerifyRequestProvider provider = 2; 
```

this way, we can still perform verification against the identity on-demand, rather than simply verifying the validity of the signature

api/proto/store/v1/store_service.proto
```diff
@@ -37,54 +37,44 @@ service StoreService {
...
+  }
 }
 
-// PushWithOptionsResponse is returned after successfully storing a record with options.
-message PushWithOptionsResponse {
-  // Reference to the stored record
-  core.v1.RecordRef record_ref = 1;
+// PushReferrerResponse
+message PushReferrerResponse {
```
ramizpolic, line 60: should we perhaps return the success and error messages here?

cli/cmd/pull/pull.go
ramizpolic, line 1: when we pull the record, will the signature field in the record.signature be present?
also, what happens on the push flow?

i am not sure i understand for now what happens to the record.signature property at all

cli/cmd/sign/options.go
```diff
@@ -5,33 +5,29 @@ package sign
...
+	flags.StringVar(&opts.RekorURL, "rekor-url", cosign.DefaultRekorURL,
 		"Sigstore Rekor URL")
-	flags.StringVar(&opts.TimestampURL, "timestamp-url", client.DefaultTimestampURL,
+	flags.StringVar(&opts.TimestampURL, "timestamp-url", cosign.DefaultTimestampURL,
 		"Sigstore Timestamp URL")
-	flags.StringVar(&opts.OIDCProviderURL, "oidc-provider-url", client.DefaultOIDCProviderURL,
+	flags.StringVar(&opts.OIDCProviderURL, "oidc-provider-url", cosign.DefaultOIDCProviderURL,
 		"OIDC Provider URL")
-	flags.StringVar(&opts.OIDCClientID, "oidc-client-id", client.DefaultOIDCClientID,
+	flags.StringVar(&opts.OIDCClientID, "oidc-client-id", cosign.DefaultOIDCClientID,
```
ramizpolic, line 34: consider moving this to a public function `RegisterFlags(flags)` so that we can register all sign flags when we pass our flagset. This can then be used from push directly from the init function rather than duplicating that code in push as well

cli/cmd/push/push.go
```diff
@@ -76,23 +76,56 @@ func runCommand(cmd *cobra.Command, source io.ReadCloser) error {
...
-	if opts.SignaturePath != "" {
-		signatureSource, err := os.ReadFile(opts.SignaturePath)
-		if err != nil {
-			return fmt.Errorf("failed to read signature file: %w", err)
-		}
-
-		signature := &signv1.Signature{}
-		if err := json.Unmarshal(signatureSource, signature); err != nil {
-			return fmt.Errorf("failed to unmarshal signature: %w", err)
+	if opts.Sign {
```
ramizpolic: can this logic be moved to a public function under sign command? i dont think we should be doing anything here beyond push operation. signing should be defined under signing logic (ie the command) because then both sign and push commands can reuse that logic

cli/cmd/verify/verify.go
```diff
@@ -30,125 +25,43 @@ Usage examples:
...
-	// Load OASF data (supports v1, v2, v3) into a Record
-	record, err := corev1.LoadOASFFromReader(recordSource)
-	if err != nil {
-		return fmt.Errorf("failed to load OASF: %w", err)
-	}
-
-	// Load signature from file
-	signatureData, err := io.ReadAll(signatureSource)
+	// Verify with Zot
+	response, err := c.VerifyWithZot(cmd.Context(), &signv1.VerifyRequest{
```
ramizpolic: there is no need to expose Zot in this context at all, this should be simply the `c.Verify`

client/config.go
```diff
@@ -14,7 +14,9 @@ import (
 const (
 	DefaultEnvPrefix = "DIRECTORY_CLIENT"
 
-	DefaultServerAddress = "0.0.0.0:8888"
+	DefaultServerAddress   = "0.0.0.0:8888"
+	DefaultRegistryAddress = "127.0.0.1:5000"
+	DefaultRepositoryName  = "dir"
```
ramizpolic: not needed

client/sign.go
```diff
@@ -42,194 +24,168 @@ type SignOpts struct {
...
 	}
 
-	// Sign the record JSON data.
-	sigBundle, err := sign.Bundle(&sign.PlainData{Data: recordJSON}, signKeypair, signOpts)
+	// Push signature to store
+	err = c.pushSignatureToStore(ctx, req.GetRecordRef().GetCid(), signatureObj)
 	if err != nil {
-		return nil, fmt.Errorf("failed to sign record: %w", err)
+		return nil, fmt.Errorf("failed to store signature: %w", err)
 	}
```
ramizpolic: i do not think that we should be pushing the signature during signing. the sign should simply generate the signature. we can use pushSignature method if we want to do this

client/sign.go
```diff
@@ -42,194 +24,168 @@ type SignOpts struct {
...
-	sigData := sigBundle.GetMessageSignature()
+	return &signv1.SignResponse{
+		Signature: signatureObj,
+	}, nil
+}
 
-	// Extract data from the signature bundle.
-	sigBundleJSON, err := protojson.Marshal(sigBundle)
+// pushSignatureToStore stores a signature using the PushReferrer RPC.
+func (c *Client) pushSignatureToStore(ctx context.Context, recordCID string, signature *signv1.Signature) error {
```
ramizpolic:
```suggestion
func (c *Client) PushSignatureReferrer(ctx context.Context, recordCID string, signature *signv1.Signature) error {
```

client/store.go
```diff
@@ -200,78 +199,24 @@ func (c *Client) PullBatch(ctx context.Context, recordRefs []*corev1.RecordRef)
...
-
-	return response, nil
-}
-
-// PullWithOptions retrieves a record along with its associated OCI artifacts.
-func (c *Client) PullWithOptions(ctx context.Context, recordRef *corev1.RecordRef, includeSignature bool) (*storev1.PullWithOptionsResponse, error) {
-	// Create streaming client
-	stream, err := c.StoreServiceClient.PullWithOptions(ctx)
+// PushWithSigning pushes a record and optionally signs it using the separate signing flow.
+func (c *Client) PushWithSigning(ctx context.Context, record *corev1.Record, sign bool, signatureProvider *signv1.SignRequestProvider) (*corev1.RecordRef, error) {
```
ramizpolic: do we really need this? seems redundant

ramizpolic: we already have all the primitives we need, i dont think we need to simply further.

instead, we would need the `PushRefferer` and `PullReferrer` functions

client/verify.go
```diff
@@ -4,170 +4,18 @@
...
-	}
-
-	if req.GetSignature() == nil {
-		return nil, errors.New("signature must be set")
-	}
-
-	// Extract signature data.
-	sigBundleRawJSON, err := base64.StdEncoding.DecodeString(req.GetSignature().GetContentBundle())
+// VerifyWithZot verifies the signature of the record using zot's verification API via the server.
+func (c *Client) VerifyWithZot(ctx context.Context, req *signv1.VerifyRequest) (*signv1.VerifyResponse, error) {
```
ramizpolic: i would consider either renaming this to only `Verify` or if we want to have local and remote verification, then `VerifyRemote`
```suggestion
// VerifyRemote verifies the signature of the record on the the server.
func (c *Client) VerifyRemote(ctx context.Context, req *signv1.VerifyRequest) (*signv1.VerifyResponse, error) {
```

server/controller/sign.go
```diff
@@ -0,0 +1,83 @@
...
+			errMsg = "Signature verification failed"
+		}
+
+		return &signv1.VerifyResponse{
+			Success:      verified,
+			ErrorMessage: &errMsg,
+		}, nil
+	}
+
+	return nil, status.Error(codes.Unimplemented, "zot verification not available in this store configuration") //nolint:wrapcheck
```
ramizpolic: simplification, usually we want to preemptively exit if we can rather than wrap around valid use-cases

```suggestion
	// Check if we support zot verification on the store API
	if zotStore, ok := s.store.(interface {
		VerifyWithZot(ctx context.Context, recordCID string) (bool, error)
	}); !ok {
		return nil, status.Error(codes.Unimplemented, "zot verification not available in this store configuration")
	}

	signLogger.Debug("Attempting zot verification", "recordCID", recordCID)

	verified, err := zotStore.VerifyWithZot(ctx, recordCID)
	if err != nil {
		return nil, status.Errorf(codes.Internal, "zot verification failed: %v", err)
	}

	signLogger.Debug("Zot verification completed", "recordCID", recordCID, "verified", verified)

	var errMsg string
	if !verified {
		errMsg = "Signature verification failed"
	}

	return &signv1.VerifyResponse{
		Success:      verified,
		ErrorMessage: &errMsg,
	}, nil
```

server/controller/store.go
```diff
@@ -194,101 +193,77 @@ func (s storeCtrl) Delete(stream storev1.StoreService_DeleteServer) error {
...
-		}
+			storeLogger.Error("Failed to store signature", "error", err, "cid", request.GetRecordRef().GetCid())
 
-		recordRef := request.GetRecordRef()
-
-		// Validate record reference
-		if err := s.validateRecordRef(recordRef); err != nil {
-			return err
+			return nil, status.Errorf(codes.Internal, "failed to store signature: %v", err)
 		}
```
ramizpolic: same logic here regarding preemptive exit

ramizpolic: i have noticed that we have extended store interafaces, perhaps we could create common types for these like

```
type SignatureStore interface {
	PushSignature()
	PullSignature()
	VerifySignature()
}
```

then you could check if the store is an instance of this interface, which would make the logic more reusable

server/store/oci/oci.go
```diff
@@ -333,249 +325,111 @@ func (s *store) Delete(ctx context.Context, ref *corev1.RecordRef) error {
 	}
 }
 
-// PushSignature stores OCI signature artifacts for a record.
+// PushSignature stores OCI signature artifacts for a record using cosign attach signature and uploads public key to zot for verification.
 func (s *store) PushSignature(ctx context.Context, recordCID string, signature *signv1.Signature) error {
 	logger.Debug("Pushing signature artifact to OCI store", "recordCID", recordCID)
```
ramizpolic, line 330: consider moving the logic for signature under `server/store/oci/sign_store.go` file to separate from the common storage interface for OCI

server/store/oci/oci.go
```diff
@@ -333,249 +325,111 @@ func (s *store) Delete(ctx context.Context, ref *corev1.RecordRef) error {
...
-
-	// Check if the repository supports the referrers API
-	referrerLister, ok := s.repo.(ReferrersLister)
-	if !ok {
-		logger.Debug("Repository does not support Referrers API, falling back to tag schema")
-
-		return nil, errors.New("repository does not support Referrers API")
+// VerifyWithZot queries zot's verification API to check if a signature is valid.
+func (s *store) VerifyWithZot(ctx context.Context, recordCID string) (bool, error) {
+	verifyOpts := &zot.VerificationOptions{
```
ramizpolic, line 411: should not be zot-specific from the naming perspective

utils/zot/zot.go
ramizpolic, line 1: zot seems to have all the neccessary objects defined under the SDK, refs:
- https://pkg.go.dev/zotregistry.dev/zot@v1.4.4-0.20250726071026-966d4584ba72/pkg/extensions/config/sync#pkg-index
- https://pkg.go.dev/zotregistry.dev/zot@v1.4.4-0.20250726071026-966d4584ba72/pkg/extensions/search
- https://pkg.go.dev/zotregistry.dev/zot@v1.4.4-0.20250726071026-966d4584ba72/pkg/extensions/config

i see more SDK packages for Zot ops, i would suggest to take a look if we can reuse things from there

cli/cmd/pull/pull.go
paralta: The signature will no longer be embedded in the record object

client/sign.go
paralta: adding an extra step for the user to push the signature? I think that would worsen usability

api/proto/sign/v1/sign_service.proto
paralta: Let's limit would verification to zot for now, in the future we can leverage `cosign verify-blob`

server/store/oci/oci.go
paralta: we need it here because in the controller we want to be able to check if this store type supports zot verification

cli/cmd/sign/sign.go
```diff
@@ -36,50 +35,48 @@ with the default OIDC provider.
...
-			return fmt.Errorf("failed to get reader: %w", err)
+			recordCID = args[0]
+		} else {
+			return errors.New("record CID is required")
 		}
 
-		return runCommand(cmd, source)
+		return runCommand(cmd, recordCID)
 	},
 }
```
ramizpolic, line 58: just as a note, cobra has native validators for this: https://github.com/spf13/cobra/blob/v1.9.1/args.go#L94

cli/cmd/verify/verify.go
```diff
@@ -30,125 +25,42 @@ Usage examples:
...
 1. Verify a record from file:
 
-	dirctl verify record.json signature.json
-
-2. Verify a record from standard input:
-
-	dirctl pull <digest> | dirctl verify --stdin 
-	# TODO Update to pull record and signature from store
-
+	dirctl verify <record-cid>
```
ramizpolic, line 28: TODO: add task to have client-side verification as well (similar to how it was done before) that leverages verify-blob

client/store.go
```diff
@@ -200,78 +199,16 @@ func (c *Client) PullBatch(ctx context.Context, recordRefs []*corev1.RecordRef)
...
-	options := &storev1.PullOptions{
-		IncludeSignature: includeSignature,
-	}
-
-	// Create request
-	request := &storev1.PullWithOptionsRequest{
+// PushReferrer pushes a record and optionally signs it using the separate signing flow.
+func (c *Client) PushSignatureReferrer(ctx context.Context, recordRef *corev1.RecordRef, signatureProvider *signv1.SignRequestProvider) error {
+	// Sign the record if requested
+	_, err := c.Sign(ctx, &signv1.SignRequest{
```
ramizpolic: this command is doing multiple things, i dont think it should be done this way, same for signing. This should be the expected flow:

- Sign SDK only creates signatures, it does not upload those
- Verify SDK can verify signatures locally or remotely (currently, we have remote only)
- The users should use the signing/verify and push in the following way
- signCMD
	- Pass digest to sign service
	- Sign SDK responds with signature
	- PushSignatureRefferer
- pushCMD with sign
	- push the artifact
	- pass digest to sign service
	- sign SDK responds with signature
	- PushSignatureReffer

Any other way of doing this is performing multiple things at the same time, which makes it non-atomic and non-reusable

ramizpolic: we can decide if we want to have this operation performed on the service side or only via the SDK. IMO it is useful if we have it as a service as well for other clients. The only issue is the 'key'-based signing because the dir service wont have access to the file, so it would need to be transported or somehow imported to the dir-server.

PR #290, opened 2025-08-12
fix: suppress command usage display on runtime errors

paralta:
When CLI commands encountered runtime errors (like "record not found"), they would display both the error message and full usage information, creating verbose and confusing output. 

This change adds `SilenceUsage: true` to all Cobra command definitions to prevent automatic usage display on runtime errors while preserving help functionality via the `--help` flag.

**Old logs:**
Screenshot 2025-08-12 at 17 26 25


cli/cmd/info/info.go
```diff
@@ -24,6 +24,7 @@ Usage example:
 	dirctl info <digest>
 
 `,
+	SilenceUsage: true,
```
tkircsi: Wouldn't it be better to set this in RootCmd and enable it where necessary?

tkircsi (commented):
Just a small comment.

cli/cmd/info/info.go
paralta: great point 👍 let me update! thanks for reviewing

ramizpolic (approved):

PR #291, opened 2025-08-12
Backport/v0.2.x - Patch fix for `agents` to `records` migration

nishantapatil3:
The SaaS hub service proto files have been updated and corresponding RPC calls for `dirctl hub` have to be updated to be able to perform hub related actions. Below two rpc calls have been changed to records.

PushAgents -> PushRecords
PullAgents -> PullRecords


ramizpolic:
Hi @nishantapatil3, can you please fix the DCO issue so we can review once the pipeline is done

ramizpolic:
another note, please make sure that the pipeline is green so we can review it. there are linter and api checks which need to be up-to-date. To check for lint issues, run `task lint:go`, and to fix: `FIX=true task lint:go`

.gitignore
```diff
@@ -30,3 +30,4 @@ agent_model.json
 .licensei.cache
 
 tmp/
+tools/python/client/agntcy_dir_client_sdk.egg-info
```
nishantapatil3: This directory was generated during `task test:unit`, IMO this python-egg should be ignored in git

ramizpolic: we could also ignore all eggs, ie `*.egg-info`

hub/api/v1alpha1/record.proto
```diff
@@ -0,0 +1,46 @@
...
+
+  optional string description = 17 [(buf.validate.field).string.max_len = 350];
+  repeated string authors = 18;
+  repeated string extensions = 19;
+
+  // Mcp related field
+  optional Mcp mcp = 20;
+  
+  // A2A related field
+  optional A2A a2a = 21;
```
ramizpolic: please consider removing this and using native mcp and a2a OASF extensions directly from the record. i am not sure why this would be needed here as it is already part of the record extension data.

refs:
- https://schema.oasf.outshift.com/0.6.0/objects/record
- https://schema.oasf.outshift.com/0.6.0/features/mcp
- https://schema.oasf.outshift.com/0.6.0/features/a2a

ramizpolic (approved):
LGTM

hub/api/v1alpha1/record.proto
nishantapatil3: Got it! After checking the proto files, these proto are not required for `dirctl hub` subcommand

.gitignore
nishantapatil3: fixed!

PR #295, opened 2025-08-13
feat: add spire federation support

ramizpolic:
## Summary
This PR simplifies the SPIFFE/SPIRE authentication configuration by removing the explicit trust domain configuration requirement and letting SPIRE handle trust domain validation automatically. It also enables federation via deployment configuration.

## Key Changes

- Removed trust domain control flows from codebase and passed the responsibility to SPIRE Server
- Use `tlsconfig.AuthorizeAny()` instead of explicit trust domain authorization. In the follow-up PR, authorization policies will be added for access control. In addition, this change also ensures that
	- SPIRE handles trust domain validation automatically through bundle verification
- Added `task test:spire` to validate SPIRE deployment between two trust domains running in separate Kind clusters
- Updated helm chart deployments for `dir` and `dirctl` to leverage SPIRE with or without SPIRE/SPIRE federation

## Follow-ups

- #304
- #307


lgecse (approved):
LGTM!

PR #310, opened 2025-08-20
feat: enable search on synced records

paralta:
This PR introduces post-sync monitoring to the Directory server, enabling automatic detection and handling of registry changes during sync operations. The implementation adds a monitoring service that tracks registry snapshots and triggers indexing when new records are detected.

* Implemented a new monitoring service with configurable monitoring intervals, registry snapshot comparison, and indexing of synced records. This service starts when sync operations are active and stops when no syncs are running
* Added detailed search examples to `USAGE.md` covering name, version, skill, locator, and extension-based queries with pagination support

### DIFF
--- a/server/sync/monitor/monitor.go
+++ b/server/sync/monitor/monitor.go
@@ -284,7 +284,6 @@ func (s *MonitorService) createRegistrySnapshot(ctx context.Context) (*RegistryS
 	return &RegistrySnapshot{
 		Timestamp:    time.Now(),
 		Tags:         tags,
-		TagCount:     len(tags),
 		ContentHash:  contentHash,
 		LastModified: time.Now(),
 	}, nil
--- a/server/sync/monitor/types.go
+++ b/server/sync/monitor/types.go
@@ -11,7 +11,6 @@ import (
 type RegistrySnapshot struct {
 	Timestamp    time.Time `json:"timestamp"`
 	Tags         []string  `json:"tags"`
-	TagCount     int       `json:"tag_count"`
 	ContentHash  string    `json:"content_hash"` // Hash of all tags for quick comparison
 	LastModified time.Time `json:"last_modified"`
 }
@@ -20,7 +19,6 @@ type RegistrySnapshot struct {
 var EmptySnapshot = &RegistrySnapshot{
 	Timestamp:    time.Now(),
 	Tags:         []string{},
-	TagCount:     0,
 	ContentHash:  "",
 	LastModified: time.Now(),
 }


server/sync/monitor/monitor.go
```diff
@@ -0,0 +1,348 @@
...
+	// Clear active syncs
+	s.activeSyncs = make(map[string]struct{})
+
+	logger.Info("Monitor service stopped")
+
+	return nil
+}
+
+// StartSyncMonitoring begins monitoring when a sync operation starts.
+func (s *MonitorService) StartSyncMonitoring(syncID string) error {
```
ramizpolic, line 100: can we pass the context here in any way, otherwise this may hang forever unless on force-exit. we can pass the actual server context to the MonitorService and reuse it here as well

server/sync/monitor/types.go
```diff
@@ -0,0 +1,33 @@
...
+package monitor
+
+import (
+	"time"
+)
+
+// RegistrySnapshot represents the state of the local registry at a point in time.
+type RegistrySnapshot struct {
+	Timestamp    time.Time `json:"timestamp"`
+	Tags         []string  `json:"tags"`
```
ramizpolic, line 13: can you elaborate which tags we keep as a ref? my concern here is that the number of tags we have in the OCI, even as names, should not be collected in memory as their size can be very large, potentially >millions

server/sync/monitor/types.go
```diff
@@ -0,0 +1,33 @@
...
+	Timestamp:    time.Now(),
+	Tags:         []string{},
+	TagCount:     0,
+	ContentHash:  "",
+	LastModified: time.Now(),
+}
+
+// RegistryChanges represents detected changes between registry snapshots.
+type RegistryChanges struct {
+	NewTags    []string  `json:"new_tags"`
```
ramizpolic, line 28: same concern regarding the number of tags

server/sync/worker.go
```diff
@@ -24,21 +25,23 @@ import (
...
+	db             types.DatabaseAPI
+	store          types.StoreAPI
+	workQueue      <-chan synctypes.WorkItem
+	timeout        time.Duration
+	monitorService *monitor.MonitorService
 }
 
 // NewWorker creates a new worker instance.
-func NewWorker(id int, db types.SyncDatabaseAPI, store types.StoreAPI, workQueue <-chan synctypes.WorkItem, timeout time.Duration) *Worker {
+func NewWorker(id int, db types.DatabaseAPI, store types.StoreAPI, workQueue <-chan synctypes.WorkItem, timeout time.Duration, monitorService *monitor.MonitorService) *Worker {
```
ramizpolic, line 37: there is a hard guarantee assumption here that, if there are no syncs running, we did the job of indexing them. i would really make the monitoring service run even if there are no syncs running (decouple it from sync). consider the situation where:
- there is a running sync
- we indexed 50% of the sync data
- sync is completed (ie. the data was sync)
- the indexing/monitoring exists because there is no sync anymore

NOTE: This assumes that we have a way to know the state of a sync (running/completed/etc), but even without this scenario, the same can happen by, instead of having a state for sync completion, I decide to delete the sync.

server/sync/monitor/types.go
paralta: we currently do not filter tags but we could restrict tags to valid `CID`s. however, your concern remains if the number of local records is too high

server/sync/worker.go
paralta: good point. we can either we make sure we run a full post-sync operation after the sync is deleted or we need to decouple

server/sync/monitor/monitor.go
paralta: it will not hang because we are using `cancelMonitor` https://github.com/agntcy/dir/pull/310/files#diff-d731ae2bb1f437af8c6eecf385875ce262ac9c4bc6fbe75a36ce2b0bf30623daR42` in `Stop()`. If I pass the context here, the monitoring will stop immediately after starting

server/sync/worker.go
paralta: https://github.com/agntcy/dir/pull/310/commits/bbc5c01413f9a7a8ed72b380f83a2d6c141df31d please look into this graceful shutdown implementation. 

decoupling is also an option but it would require us to run this monitoring even when sync is never enabled

api/core/v1/cid.go
```diff
@@ -83,3 +83,10 @@ func CalculateDigest(data []byte) (ocidigest.Digest, error) {
 	// Create OCI digest
 	return ocidigest.NewDigestFromBytes(ocidigest.SHA256, hash[:]), nil
 }
+
+// IsValidCID validates a CID string.
+func IsValidCID(cidString string) bool {
+	_, err := cid.Decode(cidString)
+
+	return err == nil
+}
```
ramizpolic, line 92: This is not needed for now

server/sync/monitor/monitor.go
```diff
@@ -0,0 +1,406 @@
...
+	s.lastSnapshot = snapshot
+}
+
+// createRegistrySnapshot creates a snapshot of the current registry state.
+func (s *MonitorService) createRegistrySnapshot(ctx context.Context) (*RegistrySnapshot, error) {
+	// List all tags in the repository
+	// No sorting needed, OCI spec requires tags in lexical order
+	var tags []string
+
+	err := s.repo.Tags(ctx, "", func(tagDescriptors []string) error {
```
ramizpolic, line 257: Is there a method that gives us both the tag and the digest together? this way, we can properly validate with `corev1.ConvertDigestToCID(digest) == tag`

server/sync/worker.go
ramizpolic: that is true, but the logic would be executed only once the changes are detected. im not sure that i follow the implementation there. are we waiting 120 seconds once there is no more syncs to finish up if there are any leftover tasks?

server/sync/monitor/types.go
ramizpolic: this may become an issue for large-serving dirs. an alternative would be to check against sqlite/db if the CID is already indexed, which reduces memory but increases latency. rough estimates show that this service could handle at best ~30m records per 1GB RAM, without extra for the tag diffs, snapshots, and data duplications.

if okay with this, lets start collecting machine requirements for running dir in production depending on the size of data plane.

server/sync/monitor/monitor.go
```diff
@@ -0,0 +1,406 @@
...
+	}, nil
+}
+
+// createContentHash creates a hash of the tags for quick comparison.
+func (s *MonitorService) createContentHash(tags []string) string {
+	// Create a deterministic hash of all tags
+	content := strings.Join(tags, "|")
+	hash := sha256.Sum256([]byte(content))
+
+	return hex.EncodeToString(hash[:])
```
ramizpolic: this is inefficient. it is better to loop over and sum individually rather than creating a string from all the tags. consider cases when there are a lot of tags.

server/sync/worker.go
paralta: it does not need to be 120s, it just needs to be >60s to account with the sync deletion. The same way we need to wait [here](https://github.com/agntcy/dir/blob/feat/post-sync-search/e2e/dirctl_network_sync_test.go#L155) for the test to succeed. We could possibly simplify and just run `performMonitoringCheck` after this period

server/sync/monitor/monitor.go
paralta: I could not find a method that does this, we would have to call `Resolve()` here

server/sync/monitor/monitor.go
paralta: indeed! updated at https://github.com/agntcy/dir/pull/310/commits/43feddbef9cc37027f665ec49a14e0d8d7b33175

server/sync/monitor/monitor.go
ramizpolic: lets not requery OCI for every tag

ramizpolic: later we could leverage ZOT Search API because it can give the `(TAG, DIGEST)` in one go

server/sync/monitor/monitor.go
ramizpolic, line 1: IMO we could remove the activeSync map and simply use a counter insted. When `StartSyncMonitoring` is called, bump the counter, similar with `StopSyncMonitoring`. In the same way, we can also remove `TagsCount` since we have `len(Tags)`.

I feel that the code here is a bit more complex than it needs to be. I would reconsider simplifying a bit and expose better primitives. For example, the overall starting/stopping logic should not be triggered via `Start/Stop SyncMonitoring`, but instead it only changes some state while the control loop that runs somewhere (e.g. in `New`) should launch the processes. The private methods further increase the complexity with this logic flow. IMO this is too difficult to analyze at the moment.

PR #322, opened 2025-08-25
Api key based auth

thelasttoto:
Support for api keys
Support for new auth flow (remove tenant logic)
support for orgs


thelasttoto:
bad branch

PR #333, opened 2025-08-27
feat: policy-based publish

paralta:
This PR implements the Policy-based Publish feature as outlined in issue #293.

* Removed obsolete `--network` option from publish and unpublish commands
* Updated API `PublishRequest`/`UnpublishRequest` to support record `refs`, `queries` or simply publish/unpublish all stored records. Please note that the CLI publish/unpublish behaviour is not affected or modified by these changes
* Added new `PublicationService`
* Skip SDK tests due to https://github.com/agntcy/dir/issues/334
* Implement publication service to manage publication requests and announce records to the DHT

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -468,7 +468,7 @@ tasks:
           {{ .HELM_CHART_PATH }} \
           -f {{ .HELM_VALUES_PATH }} \
           --set apiserver.image.tag="{{ .IMAGE_TAG }}" \
-          {{ if .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.publication.scheduler_interval="{{ .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
+          {{ if .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.config.publication.scheduler_interval="{{ .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
           --namespace {{ .HELM_NAMESPACE }} \
           --create-namespace \
           --install \
@@ -499,7 +499,7 @@ tasks:
           {{ .HELM_CHART_PATH }} \
           -f {{ .HELM_VALUES_PATH }} \
           --set apiserver.image.tag="{{ .IMAGE_TAG }}" \
-          {{ if .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.publication.scheduler_interval="{{ .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
+          {{ if .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.config.publication.scheduler_interval="{{ .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
           {{ .HELM_EXTRA_ARGS }} \
           --namespace {{ .HELM_NAMESPACE }} \
           --create-namespace \
@@ -628,7 +628,7 @@ tasks:
           {{ .HELM_CHART_PATH }} \
           -f {{ .HELM_VALUES_PATH }} \
           --set apiserver.image.tag="{{ .IMAGE_TAG }}" \
-          {{ if .PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.publication.scheduler_interval="{{ .PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
+            {{ if .PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.config.publication.scheduler_interval="{{ .PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
           --set apiserver.privKey="$(cat /tmp/node.privkey)" \
           --set apiserver.config.routing.key_path="/etc/agntcy/dir/node.privkey" \
           --set apiserver.config.routing.listen_address="/ip4/0.0.0.0/tcp/8999" \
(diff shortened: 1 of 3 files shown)


api/proto/routing/v1/routing_service.proto
```diff
@@ -46,18 +46,42 @@ service RoutingService {
...
+    bool all_records = 3;
+  }
+}
+
+message RecordRefs {
+  repeated core.v1.RecordRef refs = 1;
+}
+
+message RecordQueries {
+  repeated core.v1.RecordQuery queries = 1;
```
tkircsi: Why is it necessary to define `RecordRef` and `RecordQueries` locally?

api/proto/search/v1/search_service.proto
tkircsi, line 1: What is the difference between this `SearchService.Search` and the `RoutingService.List`?

api/proto/routing/v1/routing_service.proto
```diff
@@ -46,18 +46,42 @@ service RoutingService {
...
+  oneof request {
+    // References to the records to be published.
+    RecordRefs record_refs = 1;
+
+    // Queries to match against the records to be published.
+    RecordQueries queries = 2;
+
+    // Publish all stored records.
+    bool all_records = 3;
+  }
```
tkircsi: I'm not sure it's a good idea to mix the Publish function with the Search function. These are two separate things. I think Search is the client's job. The original 
`message PublishRequest {
  // Reference to the record to be published.
  string record_cid = 1;
}`
I think it would be appropriate if the long-stalled feature: streaming store client #240 https://github.com/agntcy/dir/pull/240 PR had been implemented. I think this would be the appropriate process
1. The client starts a Search and receives the CIDs
2. The client publishes the CIDs through the streaming client.

server/controller/routing.go
```diff
@@ -34,22 +34,16 @@ func NewRoutingController(routing types.RoutingAPI, store types.StoreAPI) routin
...
+	switch req.GetRequest().(type) {
+	case *routingv1.PublishRequest_RecordRefs:
+		if err := c.processRecordRefs(ctx, req.GetRecordRefs().GetRefs(), c.routing.Publish); err != nil {
+			return nil, err
+		}
+
+	case *routingv1.PublishRequest_Queries:
+	case *routingv1.PublishRequest_AllRecords:
+	default:
+		return nil, status.Errorf(codes.InvalidArgument, "invalid publish request")
```
tkircsi: According to my previous comment, this unnecessarily complicates the Publish service, and I don't think Publish is entirely "secure" in this form.

tkircsi (commented):
I left a few comments, I'd be happy to sync about this.

ramizpolic:
I would try not to mix the RecordQuery for local search and routing. The search on these services is not the same. For example, local search may implement wildcard search while routing may not. My only concern is the double usage of a functionally different object for these services which may impact user experience. However, it may be possible for `RecordQuery` to accommodate  both service with extensive documentation, but I feel like it may be easier if they are simply decoupled.

api/proto/routing/v1/routing_service.proto
ramizpolic: No, that is not how it works @tkircsi. Publication creates a job that will keep updating the routing table from the results of local search using CIDs only. This is all performed on our own dir node.

tkircsi: @ramizpolic  OK. I understand. But we don't need this PublishRequest (refs, queries, all) whether we are on the server side or the client side. I think Publish has one task, which is to publish one or more records based on CID.

api/proto/routing/v1/routing_service.proto
paralta: initially I tried the following message
```
message UnpublishRequest {
  oneof request {
    // References to the records to be unpublished.
    repeated core.v1.RecordRef refs = 1;

    // Queries to match against the records to be unpublished.
    repeated core.v1.RecordQuery queries = 1;

    // Unpublish all stored records.
    bool all_records = 3;
  }
}
```
but when running `task gen` this fails with 
```
routing/v1/routing_service.proto:64:5:syntax error: unexpected "repeated"
routing/v1/routing_service.proto:67:5:syntax error: unexpected "repeated"
```
This is a constraint when using `oneof`

api/proto/search/v1/search_service.proto
paralta: There are a couple of open issues on this topic that need refinement (https://github.com/agntcy/dir/issues/224 https://github.com/agntcy/dir/issues/205). But currently search is local via DB and list is across the network via DHT

install/charts/dir/apiserver/templates/deployment.yaml
```diff
@@ -36,6 +36,12 @@ spec:
           env:
             - name: DIRECTORY_LOGGER_LOG_LEVEL
               value: {{ .Values.log_level }}
+            {{- if .Values.publication }}
+            {{- if .Values.publication.scheduler_interval }}
+            - name: DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL
+              value: {{ .Values.publication.scheduler_interval }}
+            {{- end }}
+            {{- end }}
```
ramizpolic: are we not loading this from config already? not sure if we need this here.

ramizpolic: please extend the values.yaml helm chart file to include the publication config details as well. after that, the taskfile scheduler interval can simply override these default helm values

server/controller/routing.go
```diff
@@ -81,50 +74,11 @@ func (c *routingCtlr) List(req *routingv1.ListRequest, srv routingv1.RoutingServ
...
-	}
-
-	err = c.routing.Unpublish(ctx, ref, record)
-	if err != nil {
-		st := status.Convert(err)
-
-		return nil, status.Errorf(st.Code(), "failed to unpublish: %s", st.Message())
-	}
+	// Unpublish is intentionally not implemented.
+	// Records will be deleted from the network once their retention period (TTL) expires.
```
ramizpolic, line 81: can only be removed from our local store

install/charts/dir/apiserver/templates/deployment.yaml
paralta: we need this override for the tests, otherwise they will fail with the default 1h publication interval

paralta: > please extend the values.yaml helm chart file to include the publication config details as well. after that, the taskfile scheduler interval can simply override these default helm values

added to `values.yaml`, but if I remove it from the deployment template the override is not applied and the e2e tests fail

Taskfile.yml
```diff
@@ -468,6 +468,7 @@ tasks:
           {{ .HELM_CHART_PATH }} \
           -f {{ .HELM_VALUES_PATH }} \
           --set apiserver.image.tag="{{ .IMAGE_TAG }}" \
+          {{ if .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.publication.scheduler_interval="{{ .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
```
ramizpolic:
```suggestion
          {{ if .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}--set apiserver.config.publication.scheduler_interval="{{ .DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL }}"{{ end }} \
```

install/charts/dir/values.yaml
```diff
@@ -55,6 +55,17 @@ apiserver:
...
+  # Publication configuration
+  publication:
+    # How frequently the scheduler checks for pending publications
+    scheduler_interval: "1h"
+    
+    # Maximum number of publication workers running concurrently
+    worker_count: 1
+    
+    # Timeout for individual publication operations
+    worker_timeout: "30m"
```
ramizpolic: this needs to be under config afaik

install/charts/dir/apiserver/templates/deployment.yaml
ramizpolic: yes, because the publication in the helm values is not under config, so its not applying the overrides. ive added comments below on how to fix this

install/charts/dir/values.yaml
ramizpolic: instead of `apiserver.publication` it needs to be under `apiserver.config.publication`

PR #337, opened 2025-08-28
feat(sdk): sign and verification

### LINKED ISSUE #302: [Feature]: Implement sign and verification in the SDKs
### Problem Statement

The sign and verificaton are client sided, so hard to implement in the same way across multiple languages in the SDKs. However there is an inprogress refactor around these services which can help the client sided implementation.

### Proposed Solution

- Create an easy solution for getting the dependencies for signing and verification
- Implement the functions with the refactored services
- Test the SDK code

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

arpad-csepi:
Resolves #302

### DIFF
--- a/sdk/javascript/__tests__/v1/client_test.js
+++ b/sdk/javascript/__tests__/v1/client_test.js
@@ -90,10 +90,6 @@ describe('Client', () => {
 
     const client = new Client(Config.loadFromEnv());
 
-    beforeEach(async () => {
-        await setTimeout(3000); // Wait for api server
-    });
-
     afterAll(async () => {
         client.storeClient.close();
         client.routingClient.close();
--- a/sdk/python/v1/test_client.py
+++ b/sdk/python/v1/test_client.py
@@ -1,6 +1,5 @@
 import os
 import subprocess
-import time
 import unittest
 
 import core.v1.record_pb2 as core_record_pb2
@@ -65,8 +64,6 @@ def init_records(count, test_function_name, push=True, publish=False):
             for record_ref, record in example_records.values():
                 req = routingv1.PublishRequest(record_cid=record_ref.cid)
                 client.publish(req=req)
-
-    time.sleep(3)
 
     return example_records
 


arpad-csepi:
> there should be no cosign dependency in the SDK logic except maybe in tests. im not sure why we need this. key generation should happen outside of the SDK if needed

No need for cosign in the SDK, that was a mistake. Moved to the tests.

Taskfile.yml
```diff
@@ -207,6 +208,28 @@ tasks:
...
       - go -C . test ./... {{.EXTRA_ARGS}}
 
+  sdk:deps:common:
+    desc: Common dependencies for SDKs
+    cmds:
+      - task: deps:cosign
+      - task: cli:compile
+
+  sdk:deps:iodc-token-generation:
+    desc: Get Fulcio OIDC token for CI
```
ramizpolic: nit: more general deps for SDK CICD job

```suggestion
  sdk:deps:cicd:
    desc: Get SDK dependencies for CICD
```

.github/workflows/reusable-test.yaml
```diff
@@ -132,8 +136,17 @@ jobs:
...
           docker images
 
-      - name: Test SDKs
+      - name: Add dependencies for SDKs testing
         run: |
+          CLIENT_ID="https://github.com/${{ github.repository }}/.github/workflows/reusable-test.yaml@${{ github.ref }}"
+
+          task sdk:deps:common
           task sdk:deps:javascript
+          echo "$PWD/bin/" >> $GITHUB_PATH # Add deps bin to shell
```
ramizpolic: which deps do we require so that we need to add these to the path? if its just `dirctl`, this can be passed through the config

Taskfile.yml
```diff
@@ -42,6 +42,7 @@ vars:
   UV_VERSION: '0.6.1'
   UV_BIN: '{{ .BIN_DIR }}/uv-{{.UV_VERSION}}'
   UV_PUBLISH_TOKEN: '{{ .UV_PUBLISH_TOKEN | default "" }}'
+  COSIGN_VERSION: '2.5.3'
   HUB_API_VERSION: 'main'
```
ramizpolic:
```suggestion
  COSIGN_VERSION: '2.5.3'
  COSIGN_BIN: '{{ .BIN_DIR }}/cosign-{{.COSIGN_VERSION}}'
  HUB_API_VERSION: 'main'
```

Taskfile.yml
```diff
@@ -1320,6 +1348,16 @@ tasks:
...
 
+  deps:cosign:
+    desc: Install sigstore cosign
+    dir: '{{ .BIN_DIR }}'
+    internal: true
+    cmds:
+      - curl -sfL https://github.com/sigstore/cosign/releases/download/v{{.COSIGN_VERSION}}/cosign-{{OS}}-{{ARCH}} -o cosign
+      - chmod +x cosign
+    status:
+      - test -x cosign
```
ramizpolic:
```suggestion
      - curl -sfL https://github.com/sigstore/cosign/releases/download/v{{.COSIGN_VERSION}}/cosign-{{OS}}-{{ARCH}} -o cosign-{{.COSIGN_VERSION}}'
      - chmod +x cosign-{{.COSIGN_VERSION}}'
    status:
      - test -x cosign-{{.COSIGN_VERSION}}'
```

.github/workflows/reusable-test.yaml
ramizpolic: if cosign is needed, this can also be setup via `COSIGN_PATH` env-var

sdk/javascript/v1/client.js
```diff
@@ -264,6 +272,124 @@ class Client {
...
+
+        // INFO: Assume that the request is key based if the provider array
+        // first item is a buffer (key file)
+        try {
+            key_file_found = Buffer.isBuffer(provider[0]);
+        } catch (error) {
+            if (!error.message.includes("Cannot read properties of undefined")) {
+                throw err;
+            }
+        }
```
ramizpolic: is there no explicit field that is set for key-based and oidc-based signing? afaict `req` is the `SignRequest` type, so we can decide without assumptions. is this the case or not?

ramizpolic:
```js
let oidc_provider = req.u[1][0];
let key_provider = req.u[1][1];
```

can we use the check if key_provider/oidc_signing is not null checks

Taskfile.yml
arpad-csepi: Renamed to `sdk:deps:cicd:iodc-token-generation` because this task output is processed (piped) to get the values out of from the taskfile subshell into GH action shell. So it should not be added any more CICD dependency to it because of this.

sdk/javascript/v1/client.js
arpad-csepi: We can use it, I correct this part to your suggestion, thank you!

.github/workflows/reusable-test.yaml
arpad-csepi: Implemented

sdk/javascript/__tests__/v1/client_test.js
```diff
@@ -86,10 +88,10 @@ describe('Client', () => {
         return test_records;
     }
 
-    const client = new Client(new Config());
+    const client = new Client(Config.loadFromEnv());
 
     beforeEach(async () => {
-        // await setTimeout(3000); // Wait for api server
+        await setTimeout(3000); // Wait for api server
```
ramizpolic: note: this may become flaky. better check would be to curl the endpoint with a timeout until it becomes available. we dont need to do this now, just want to bring awareness if the issue arrises in the future

sdk/python/v1/client.py
```diff
@@ -339,3 +357,135 @@ def delete(
...
+                )
+            if oidc_signer.options.fulcio_url != "":
+                command = (
+                    *command,
+                    "--fulcio-url",
+                    f"{oidc_signer.options.fulcio_url}",
+                )
+            if oidc_signer.options.fulcio_url != "":
+                command = (*command, "--rekor-url", f"{oidc_signer.options.rekor_url}")
+            if oidc_signer.options.fulcio_url != "":
```
ramizpolic:
```suggestion
            if oidc_signer.options.timestamp_url != "":
```

sdk/python/v1/client.py
```diff
@@ -339,3 +357,135 @@ def delete(
...
+                    "--oidc-provider-url",
+                    f"{oidc_signer.options.oidc_provider_url}",
+                )
+            if oidc_signer.options.fulcio_url != "":
+                command = (
+                    *command,
+                    "--fulcio-url",
+                    f"{oidc_signer.options.fulcio_url}",
+                )
+            if oidc_signer.options.fulcio_url != "":
```
ramizpolic:
```suggestion
            if oidc_signer.options.rekor_url != "":
```

ramizpolic (approved):

Issue #326, opened 2025-08-26
[Feature]: Migrate List service from legacy labels to RecordQuery interface


tkircsi:
### Problem Statement

The current List service is constrained by the legacy label-based query system (LegacyListRequest) which prevents it from effectively querying locally stored records using the modern RecordQuery interface that supports semantic skills, domains, and features. The existing string-based label filtering approach lacks the type safety and semantic structure needed to properly match against the enhanced local datastore that now contains both local records and remotely announced label mappings from DHT discovery. Without RecordQuery support, the List service cannot provide a unified query experience with the Search service, forcing clients to use different query mechanisms for local versus distributed operations. This architectural inconsistency limits the ability to build cohesive discovery workflows where local queries can seamlessly complement network-wide searches using the same query vocabulary and scoring mechanisms.

### Proposed Solution

Replace the LegacyListRequest with the modern RecordQuery interface in the List service to enable semantic querying of locally stored records using skills, domains, and features. Implement query processing logic that translates RECORD_QUERY_TYPE_SKILL and RECORD_QUERY_TYPE_LOCATOR queries into efficient datastore operations against the local record storage, filtering only records that this peer actually hosts and provides. 

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

PR #342, opened 2025-09-01
refactor: Remove Legacy List API and Migrate to RecordQuery-Based System

tkircsi:
This pull request removes the legacy label-based List API and migrates to a modern RecordQuery-based system as outlined in GitHub Issue #326. The refactor establishes clear architectural boundaries between local record inventory (List) and network-wide discovery (Search), while enhancing API consistency across all services.

### Key Changes

• **Remove legacy structures** - Eliminated `LegacyListRequest` and `LegacyListResponse` from the routing service API
• **Implement RecordQuery system** - Migrated from string-based labels to typed queries with support for skills and locators  
• **Establish local-only List operation** - List now correctly returns only records that the local peer is providing, with no network interaction
• **Enhance API consistency** - Updated all routing messages to use `core.v1.RecordRef` instead of string CIDs, aligning with Store and Sign services
• **Improve response format** - Enhanced `ListResponse` with essential metadata (labels, annotations) for CLI tools while maintaining performance
• **Fix test infrastructure** - Resolved race conditions in tests by using isolated datastores and updated all client libraries and CLI commands
• **Implement self-descriptive keys** - Introduced enhanced key format `/<namespace>/<path>/<cid>/<peer_id>` for better observability and performance
• **Streamline List implementation** - Unified approach using /records/ index with efficient query filtering


api/proto/routing/v1/routing_service.proto
```diff
@@ -121,57 +121,13 @@ message ListRequest {
...
-    // Found record if any.
-    // If empty, then only the labels are important.
-    optional core.v1.RecordRef ref = 4;
-  }
-
-  // Returned items that match a given request
-  repeated Item items = 1;
+  // Labels associated with this record (skills, domains, features)
+  // Derived from the record content for CLI display purposes
+  repeated string labels = 2;
```
ramizpolic, line 132: nit for future: we can return explicit pairs `(type, value)` rather than labels. no change needed

cli/cmd/list/peer.go
```diff
@@ -13,42 +13,60 @@ import (
...
+				Value: skillName,
+			})
+		case strings.HasPrefix(label, "locator="):
+			locatorType := strings.TrimPrefix(label, "locator=")
+			queries = append(queries, &routingv1.RecordQuery{
+				Type:  routingv1.RecordQueryType_RECORD_QUERY_TYPE_LOCATOR,
+				Value: locatorType,
+			})
+		}
+	}
```
ramizpolic: can you verify if this is okay? seems to me its two different ways of handling labels for locators/skills

e2e/client_test.go
```diff
@@ -17,6 +18,40 @@ import (
...
+			featureName := strings.TrimPrefix(label, "/features/")
+			_ = featureName
+			// Note: features might need to be mapped to locator or handled differently
+			// For now, skip features as they're not in the current RecordQueryType
+		case strings.HasPrefix(label, "locator="):
+			locatorType := strings.TrimPrefix(label, "locator=")
+			queries = append(queries, &routingv1.RecordQuery{
+				Type:  routingv1.RecordQueryType_RECORD_QUERY_TYPE_LOCATOR,
+				Value: locatorType,
+			})
```
ramizpolic: same here regarding label handling

cli/cmd/list/peer.go
tkircsi: Thanks! Fixed.

e2e/client_test.go
tkircsi: Fixed.

PR #345, opened 2025-09-01
fix: on Hub, load client_id for apikey accessToken from Phoenix UI

thelasttoto:
Previously, the client_id to use when request access token for apikey hub access was hard coded.
Now, the value is read with the AuthConfig retrieves from the Phoenix UI


jubarbot-cisco (approved):
LGTM

ramizpolic (approved):

PR #355, opened 2025-09-03
feat: add wildcard search support with * pattern matching

paralta:
This PR implements wildcard search functionality across all record fields, enabling flexible pattern matching for directory searches. Users can now search for records using `*`, `?`, and `[]` wildcards in names, versions, skills, locators, and extensions.

* Wildcard pattern support - Added `*` (zero or more), `?` (exactly one), and `[]` (character ranges) wildcards across all searchable fields
* Database improvements - Migrated from LIKE to GLOB operator with case-insensitive matching and parameterized queries
* Enhanced documentation - Updated protobuf schema and CLI with comprehensive wildcard usage examples
* E2E testing - Added extensive e2e and unit tests covering all wildcard patterns, combinations, and edge cases
* Backward compatibility - All existing exact-match functionality preserved without breaking changes

### DIFF
--- a/cli/cmd/search/options.go
+++ b/cli/cmd/search/options.go
@@ -20,10 +20,3 @@ func init() {
 
 	flags.VarP(&opts.Query, "query", "q", "Search query terms")
 }
-
-// ResetGlobalState resets the global options state for testing.
-func ResetGlobalState() {
-	opts.Limit = 100
-	opts.Offset = 0
-	opts.Query = Query{} // Reset the query slice
-}
--- a/e2e/utils/common.go
+++ b/e2e/utils/common.go
@@ -395,6 +395,21 @@ func ResetCLIState() {
 }
 
 // ResetSearchCommandState resets the global state in search command.
+//
+//nolint:errcheck
 func ResetSearchCommandState() {
-	searchcmd.ResetGlobalState()
+	if cmd := searchcmd.Command; cmd != nil {
+		// Reset flags to default values
+		cmd.Flags().Set("limit", "100")
+		cmd.Flags().Set("offset", "0")
+
+		// For the query flag, reset it by accessing the underlying value
+		if queryFlag := cmd.Flags().Lookup("query"); queryFlag != nil {
+			queryFlag.Changed = false
+			// Cast to the Query type and reset it
+			if queryValue, ok := queryFlag.Value.(*searchcmd.Query); ok {
+				*queryValue = searchcmd.Query{}
+			}
+		}
+	}
 }
--- a/server/database/utils/wildcard.go
+++ b/server/database/utils/wildcard.go
@@ -14,20 +14,10 @@ func ContainsWildcards(pattern string) bool {
 
 // containsListWildcard checks if a pattern contains list wildcard characters [].
 func containsListWildcard(pattern string) bool {
-	// Look for properly formed list wildcards [...]
-	for i := range len(pattern) {
-		if pattern[i] == '[' {
-			// Find the closing bracket
-			for j := i + 1; j < len(pattern); j++ {
-				if pattern[j] == ']' {
-					// Found a complete list wildcard [...]
-					return true
-				}
-			}
-		}
-	}
+	openIdx := strings.Index(pattern, "[")
+	closeIdx := strings.Index(pattern, "]")
 
-	return false
+	return openIdx != -1 && closeIdx > openIdx
 }
 
 // BuildWildcardCondition builds a WHERE condition for wildcard or exact matching.
@@ -56,7 +46,7 @@ func BuildWildcardCondition(field string, patterns []string) (string, []interfac
 
 // BuildSingleWildcardCondition builds a WHERE condition for a single field with wildcard or exact matching.
 // Returns the condition string and argument for the WHERE clause.
-func BuildSingleWildcardCondition(field, pattern string) (string, interface{}) {
+func BuildSingleWildcardCondition(field, pattern string) (string, string) {
 	if ContainsWildcards(pattern) {
 		return "LOWER(" + field + ") GLOB ?", strings.ToLower(pattern)
 	}


server/database/utils/wildcard_test.go
```diff
@@ -0,0 +1,362 @@
...
+		},
+		{
+			name:     "escape both percent and underscore",
+			pattern:  "test%_*",
+			expected: "test\\%\\_%",
+		},
+		{
+			name:     "no wildcards with SQL chars",
+			pattern:  "test%_",
+			expected: "test%_",
```
tkircsi: shouldn't this be "test\\\\%\\\\_"?

server/database/utils/wildcard.go
```diff
@@ -0,0 +1,59 @@
...
+package utils
+
+import (
+	"strings"
+)
+
+// ContainsWildcards checks if a pattern contains wildcard characters (*).
+func ContainsWildcards(pattern string) bool {
+	return strings.Contains(pattern, "*")
+}
```
tkircsi:
```suggestion
// ContainsWildcards checks if a pattern contains unescaped wildcard characters (*).
// Escaped asterisks (\*) are not considered wildcards.
func ContainsWildcards(s string) bool {
	for i := 0; i < len(s); {
		ch := s[i]

		// Handle escapes from user input
		if ch == '\\' {
			if i+1 < len(s) {
				// Skip the escaped character (including \*)
				i += 2
				continue
			}
			// Trailing backslash
			i++
			continue
		}

		// Check for unescaped wildcard
		if ch == '*' {
			return true
		}

		i++
	}
	return false
}
```

server/database/utils/wildcard.go
```diff
@@ -0,0 +1,59 @@
...
+
+	// Escape existing SQL wildcards
+	result := strings.ReplaceAll(pattern, "%", `\%`)
+	result = strings.ReplaceAll(result, "_", `\_`)
+
+	// Convert wildcard patterns to SQL patterns
+	result = strings.ReplaceAll(result, "*", "%")
+
+	return result
+}
```
tkircsi:
```suggestion
// WildcardToSQL converts a user pattern to a SQL LIKE pattern.
//
// Rules:
// - Unescaped `*` => `%` (wildcard many)
// - Literal `%` or `_` are escaped as `\%` and `\_`
// - Literal `\` becomes `\\`
// - A backslash escapes the next special char: \*, \%, \_, \\
// - Everything else is copied as-is.
//
// Use with:  "... LIKE ? ESCAPE '\\'".
func WildcardToSQL(s string) string {
	var b strings.Builder
	b.Grow(len(s) * 2) // generous

	for i := 0; i < len(s); {
		ch := s[i]

		// Handle escapes from user input
		if ch == '\\' {
			if i+1 < len(s) {
				next := s[i+1]
				switch next {
				case '*':
					// literal asterisk
					b.WriteByte('*')
				case '%':
					// literal percent
					b.WriteString(`\%`)
				case '_':
					// literal underscore
					b.WriteString(`\_`)
				case '\\':
					// literal backslash -> must be escaped in SQL LIKE
					b.WriteString(`\\`)
				default:
					// unknown escape: keep backslash as literal then char
					b.WriteString(`\\`)
					b.WriteByte(next)
				}
				i += 2
				continue
			}
			// trailing backslash -> treat as literal backslash
			b.WriteString(`\\`)
			i++
			continue
		}

		// Normal chars
		switch ch {
		case '*':
			// unescaped * => wildcard
			b.WriteByte('%')
		case '%':
			// escape % so it's literal unless user intended wildcard with *
			b.WriteString(`\%`)
		case '_':
			// escape _ so it's literal
			b.WriteString(`\_`)
		case '\\':
			// double it for a literal backslash
			b.WriteString(`\\`)
		default:
			b.WriteByte(ch)
		}
		i++
	}
	return b.String()
}
```

server/database/utils/wildcard_test.go
```diff
@@ -0,0 +1,362 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := WildcardToSQL(tt.pattern)
+			if result != tt.expected {
+				t.Errorf("WildcardToSQL(%q) = %q, want %q", tt.pattern, result, tt.expected)
+			}
+		})
+	}
+}
+
```
tkircsi, line 269: Just a couple of backslah tests
```
func TestWildcardToSQLEscaping(t *testing.T) {
	tests := []struct {
		name     string
		input    string
		expected string
		meaning  string
	}{
		{
			name:     "simple value",
			input:    "simple value",
			expected: "simple value",
			meaning:  "literal match (no wildcards)",
		},
		{
			name:     "asterisk wildcard",
			input:    "SIM*ple value",
			expected: "SIM%ple value",
			meaning:  "* → %",
		},
		{
			name:     "SQL wildcards as literals",
			input:    "SIM%PLE _VALUE",
			expected: "SIM\\%PLE \\_VALUE",
			meaning:  "%/_ treated as literals",
		},
		{
			name:     "multiple percent signs",
			input:    "ABC%%%%%%%%CD%%%%",
			expected: "ABC\\%\\%\\%\\%\\%\\%\\%\\%CD\\%\\%\\%\\%",
			meaning:  "literal % (no wildcard effect)",
		},
		{
			name:     "escaped asterisk",
			input:    "file\\*name",
			expected: "file*name",
			meaning:  "\\* becomes literal *",
		},
		{
			name:     "literal backslash",
			input:    "path\\\\dir",
			expected: "path\\\\dir",
			meaning:  "literal backslash",
		},
		{
			name:     "complex escaping with wildcard",
			input:    "test\\*pattern*end",
			expected: "test*pattern%end",
			meaning:  "\\* becomes literal *, * becomes %",
		},
		{
			name:     "escaped asterisk only",
			input:    "no\\*wildcards\\*here",
			expected: "no*wildcards*here",
			meaning:  "all \\* become literal *",
		},
		{
			name:     "SQL chars with wildcard",
			input:    "test%_*data",
			expected: "test\\%\\_%data",
			meaning:  "escape SQL chars, convert * to %",
		},
	}

	for _, tt := range tests {
		t.Run(tt.name, func(t *testing.T) {
			result := WildcardToSQL(tt.input)
			if result != tt.expected {
				t.Errorf("WildcardToSQL(%q) = %q, want %q (meaning: %s)",
					tt.input, result, tt.expected, tt.meaning)
			}
		})
	}
}
```

server/database/utils/wildcard.go
ramizpolic: i would not go down this path. we are sanitizing the query, but this does not handle special regexp cases. can we explore native wildcard/globs supported by SQLite?

server/database/utils/wildcard.go
ramizpolic: same note as above, we cannot sanitize the string this way

ramizpolic:
I agree with @tkircsi, if we can utilize sqlite glob, it would be better than regexp search from security, simplicity, and performance standpoints

server/database/utils/wildcard_test.go
paralta: This pattern contains no wildcard `*`, so no escaping is needed because exact match `=` doesn't interpret `%`.

server/database/utils/wildcard.go
tkircsi: I think you should check this for SQLite GLOB as well.

cli/cmd/search/options.go
```diff
@@ -20,3 +20,10 @@ func init() {
 
 	flags.VarP(&opts.Query, "query", "q", "Search query terms")
 }
+
+// ResetGlobalState resets the global options state for testing.
+func ResetGlobalState() {
+	opts.Limit = 100
+	opts.Offset = 0
+	opts.Query = Query{} // Reset the query slice
+}
```
ramizpolic: lets avoid adding logic for tests under valid components. this can be overwritten using the CLI args

e2e/utils/common.go
```diff
@@ -388,4 +389,12 @@ func ResetCLIState() {
...
 	clicmd.RootCmd.SetErr(nil)
+
+	// Reset search command global state
+	ResetSearchCommandState()
+}
+
+// ResetSearchCommandState resets the global state in search command.
+func ResetSearchCommandState() {
+	searchcmd.ResetGlobalState()
 }
```
ramizpolic: this can be used to override params rather than invoking the method

server/database/utils/wildcard.go
```diff
@@ -0,0 +1,65 @@
...
+	if len(conditions) > 1 {
+		condition = "(" + condition + ")"
+	}
+
+	return condition, args
+}
+
+// BuildSingleWildcardCondition builds a WHERE condition for a single field with wildcard or exact matching.
+// Returns the condition string and argument for the WHERE clause.
+func BuildSingleWildcardCondition(field, pattern string) (string, interface{}) {
```
ramizpolic: why not return 
```suggestion
func BuildSingleWildcardCondition(field, pattern string) (string, string) {
```

ramizpolic: if a custom type is needed, i suggest creating
```
type Condition string
type Argument string
```

server/database/utils/wildcard.go
```diff
@@ -0,0 +1,65 @@
...
+			for j := i + 1; j < len(pattern); j++ {
+				if pattern[j] == ']' {
+					// Found a complete list wildcard [...]
+					return true
+				}
+			}
+		}
+	}
+
+	return false
```
ramizpolic: nit: for lazy-readers, more explicit

```suggestion
	openIdx := strings.Index(pattern, "[")
	closeIdx := strings.Index(pattern, "]")
	return openIdx != -1 && closeIdx > openIdx 
```

paralta: neat 👌

PR #362, opened 2025-09-04
feat: Implement Search API for network-wide record discovery using RecordQuery interface

tkircsi:
## 🎯 Core Routing Improvements

- **Implemented OR Logic with Minimum Threshold**: Enhanced search functionality to support OR logic with `minMatchScore` parameter, allowing records to match if they satisfy at least N queries instead of requiring all queries to match

- **Pull-Based Label Discovery Architecture**: Replaced direct DHT label announcements with a scalable pull-based system where peers announce CIDs via `DHT.Provide()`, and remote peers pull content on-demand to extract and cache labels locally

- **Added Domain and Feature Query Types**: Extended the routing API to support domain and feature queries in addition to existing skill and locator queries, with proper hierarchical matching logic

## 🚀 CLI Command Restructuring  

- **Reorganized Routing Commands**: Moved routing operations under a unified `dirctl routing` subcommand structure with `publish`, `unpublish`, `list`, `search`, and `info` operations

- **Enhanced Search Capabilities**: Added comprehensive search functionality with support for multiple query types, limit parameters, and minimum match scoring

## 🔧 Architecture & Performance

- **Background Cleanup Tasks**: Implemented automated cleanup for stale remote labels and CID provider republishing to maintain cache consistency

- **Query Deduplication**: Added server-side query deduplication to prevent client bugs from causing incorrect match scores

- **Improved Error Handling**: Enhanced error wrapping and validation throughout the routing system for better debugging and reliability

## 🧪 Testing & Quality

- **Comprehensive E2E Testing**: Added extensive end-to-end tests covering local and network deployment scenarios, multi-peer routing operations, and OR logic validation

- **Code Quality Improvements**: Refactored complex functions, removed unnecessary comments, and improved code organization for better maintainability


server/routing/routing_remote.go
```diff
@@ -154,6 +158,187 @@ func (r *routeRemote) Publish(ctx context.Context, ref *corev1.RecordRef, record
...
+					}
+
+					processedCIDs[keyCID] = true
+					processedCount++
+
+					if limitInt > 0 && processedCount >= limitInt {
+						break
+					}
+				}
+			}
```
paralta: I think that this current score calculation doesn't align with the proto definition https://github.com/agntcy/dir/blob/main/api/proto/routing/v1/routing_service.proto#L87. The proto states that `min_match_score` filters records that "match at least N queries," but the implementation calculates scores per individual label key rather than per record.

cli/README.md
```diff
@@ -0,0 +1,386 @@
+# Directory CLI (dirctl)
```
paralta, line 1: This detailed usage documentation would be better suited for the scenarios section at https://docs.agntcy.org/dir/scenarios/ rather than in this CLI README. Combining usage details in one location will also reduce documentation maintenance effort.

server/routing/routing_remote.go
tkircsi: Thanks! Fixed!

cli/README.md
tkircsi: I'll do it in that repo and I 'll remove this.

server/routing/cleanup_tasks.go
ramizpolic, line 1: do we need the cleanup task with the current datastore? if the local datastore has the TTL, we dont have to implement cleanup

server/routing/label_extraction.go
```diff
@@ -0,0 +1,81 @@
...
+// This is a pure function that can be used by both local and remote routing operations.
+//
+// The function extracts labels from:
+// - Skills: /skills/<skill_name>
+// - Domains: /domains/<domain_name> (from extensions with domain schema prefix)
+// - Features: /features/<feature_name> (from extensions with features schema prefix)
+// - Locators: /locators/<locator_type>
+//
+// Returns a slice of fully qualified label strings with namespace prefixes.
+func GetLabels(record *corev1.Record) []string {
```
ramizpolic:
```suggestion
type Label string

type LabelType uint

const (
	LabelTypeSkill LabelType = iota + 1
	LabelTypeDomain
	LabelTypeFeature
)

func (rl Label) Type() LabelType {
	if strings.StartsWith(rl, "/skill") {
		return LabelTypeSkill
	}
	
	// etc
}

func GetLabels(record *corev1.Record) []Label {
```

server/routing/cleanup_tasks.go
tkircsi: It is need now. Both local and remote labels are stored in the local datastore. Local datastore would clean up the local labels also which means we should "republish" them. Using local datastore TTL brings up design questions also, because we'd need to synchronize DHT and local datastore TTLs. The current clean-up logic checks more than TTL it removes orphaned local records also.

tkircsi: In order for local datastore TTL to be considered, the DHT and local label stores must first be separated. In the current situation, TTL would also affect DHT records.

server/routing/label_extraction.go
tkircsi: I like it, but it has some benefits and drawbacks also
Pros:
- Type Safety & API Clarity
- Extensible Design
- Some performance Benefits

Cons:
- Increased Complexity
- Validation Overhead
- Storage & Serialization Concerns

I'll take a closer look at this solution.

cli/cmd/root.go
```diff
@@ -66,14 +64,12 @@ func init() {
 		pull.Command,
 		push.Command,
 		delete.Command,
-		// routing commands
-		publish.Command,
-		list.Command,
-		unpublish.Command,
+		// routing commands (all under routing subcommand)
+		routing.Command, // Contains: publish, unpublish, list, search
 		network.Command,
```
ramizpolic, line 69: what is this network command? should we remove it?

server/routing/cleanup_tasks.go
ramizpolic: a simple way to fix this is to use two instances of the datastore, one for local routing, other one for DHT

tkircsi: That would only solve the DHT - local store connection. It must be solved separately so that live local records/labels are not deleted, only orphans and expired remotes. And we assume that we are using a KV datastore that supports TTL. Currently, there are two configurable options
1. BadgerDB - supports
2. Custom MapDatastore - does not support

cli/cmd/root.go
tkircsi: I don't know. It was added 5 months ago.
https://github.com/agntcy/dir/pull/85

server/routing/label_extraction.go
tkircsi: fixed

cli/cmd/root.go
ramizpolic: we need to check if its still needed, would be better to move this to routing

PR #363, opened 2025-09-05
feat: wire in validation

adamtagscherer:
This PR wires in the validation SDK when a record is pushed to the store service and after the sync event.


e2e/client_test.go
```diff
@@ -98,8 +98,8 @@ var _ = ginkgo.Describe("Running client end-to-end tests using a local single no
 			name:     "V3_Record_OASF_v0.5.0",
 			jsonData: expectedRecordV3JSON,
 			expectedSkillLabels: []string{
-				"/skills/Natural Language Processing/Text Completion",
-				"/skills/Natural Language Processing/Problem Solving",
+				"schema.oasf.agntcy.org/skills/contextual_comprehension",
+				"schema.oasf.agntcy.org/skills/entity_recognition",
```
ramizpolic, line 102: this is not valid. it needs to be

```suggestion
				"/skills/contextual_comprehension",
				"/skills/entity_recognition",
```

e2e/dirctl_network_deploy_test.go
```diff
@@ -100,8 +100,8 @@ var _ = ginkgo.Describe("Running dirctl end-to-end tests using a network multi p
 		// Should find the local record with expected labels
 		gomega.Expect(output1).To(gomega.ContainSubstring(cid))
 		gomega.Expect(output1).To(gomega.ContainSubstring("Local Record"))
-		gomega.Expect(output1).To(gomega.ContainSubstring("/skills/Natural Language Processing/Text Completion"))
-		gomega.Expect(output1).To(gomega.ContainSubstring("/skills/Natural Language Processing/Problem Solving"))
+		gomega.Expect(output1).To(gomega.ContainSubstring("/skills/schema.oasf.agntcy.org/skills/contextual_comprehension"))
+		gomega.Expect(output1).To(gomega.ContainSubstring("/skills/schema.oasf.agntcy.org/skills/entity_recognition"))
```
ramizpolic, line 104: same here

e2e/client_test.go
adamtagscherer: This is a v0.5.0 schema object where this is the proper skill naming, see: https://schema.oasf.outshift.com/0.5.0/skills/contextual_comprehension

I think you mistake it with the v0.6.0 schema version where is correct what you say.

ramizpolic: you are right

tkircsi: I think, this is a mismatch between the v0.5.0 and v0.6.0. Version v0.5.0 did not have this prefix before; it was added later. The extensions/domains and extensions/features tags previously had `schema.oasf.agntcy.org`, and we "cut" these out, for example in routing. https://github.com/agntcy/dir/blob/c9585955adacf3af5f6879e08038508218a6dcc5/server/routing/routing_local.go#L374
We also need to keep in mind, handling prefixes for records and annotations here https://github.com/agntcy/dir/blob/c9585955adacf3af5f6879e08038508218a6dcc5/server/store/oci/annotations.go#L17

I think we should discuss how to handle these in the future.

server/controller/store.go
```diff
@@ -29,16 +31,23 @@ var storeLogger = logging.Logger("controller/store")
...
 	storev1.UnimplementedStoreServiceServer
-	store types.StoreAPI
-	db    types.DatabaseAPI
+	store     types.StoreAPI
+	db        types.DatabaseAPI
+	validator *service.ValidationService
 }
 
-func NewStoreController(store types.StoreAPI, db types.DatabaseAPI) storev1.StoreServiceServer {
+func NewStoreController(store types.StoreAPI, db types.DatabaseAPI) (storev1.StoreServiceServer, error) {
```
tkircsi, line 39: Can we create a validation interface and add validation dependency as we do with Store and Database?

server/controller/store.go
```diff
@@ -57,6 +66,25 @@ func (s storeCtrl) Push(stream storev1.StoreService_PushServer) error {
...
+				return status.Errorf(codes.Internal, "failed to validate record: %v", err)
+			}
+
+			if !isValid {
+				storeLogger.Warn("Record validation failed", "errors", errors)
+
+				return status.Errorf(codes.InvalidArgument, "record validation failed: %v", errors)
+			}
+		}
+
```
tkircsi, line 87: Maybe we can wire this into `validateAndGetCID`. WDYT?

ramizpolic: im more in favour of keeping things like validation and CID calculation two separate methods

tkircsi: Agree. There is a `validateRecord` method also. Maybe we can use that.

PR #373, opened 2025-09-05
feat: fallback to client-side verification

paralta:
This pull request implements a fallback mechanism that enables client-side signature verification when server-side verification fails, improving system resilience and availability.

* Fallback Verification Logic: Enhanced the Verify method to automatically fallback to client-side cryptographic verification using Cosign when server verification is unsuccessful
* OCI Referrer Integration: Added methods to retrieve signature and public key data from OCI referrers, enabling independent verification without server dependency


lgecse (approved):
LGTM, thanks @paralta! :)

PR #377, opened 2025-09-08
feat: add API Key authentication with no-cache option

gicont:
Enables **authentication without session files** by providing client ID and secret via command flags or environment variables **for push and pull commands**. Added --no-cache flag to skip session file interactions entirely.

The **original plain-text secret** (not JSON-escaped) is **base64-encoded** immediately after API key creation.
It remains encoded **throughout the entire flow**, when **stored** in session files, passed via **CLI flags**, or **environment variables**

### DIFF
--- a/hub/utils/auth/auth.go
+++ b/hub/utils/auth/auth.go
@@ -4,7 +4,6 @@
 package auth
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


hub/utils/auth/auth.go
```diff
@@ -29,3 +31,72 @@ func CheckForCreds(cmd *cobra.Command, currentSession *sessionstore.HubSession,
...
+
+// resolveAPIKeyCredentials returns the effective clientID and secret, with JSON decoding applied if needed.
+func resolveAPIKeyCredentials(clientID, secret string) (string, string, string) {
+	effectiveClientID := clientID
+	effectiveSecret := secret
+
+	// Decode JSON-encoded strings from CLI flags if both are provided.
+	if effectiveClientID != "" && effectiveSecret != "" {
+		var decodedSecret string
+		if err := json.Unmarshal([]byte(`"`+effectiveSecret+`"`), &decodedSecret); err == nil {
```
mtrinell, line 43: I'm confused by this logic. Shouldn't be `err != nil` ?

jubarbot-cisco: I'm not comfortable with that unmarshal of the secret with double-quotes added around it.

Why do we even need to do that ?

Why is it even possible that the secret provided from the command line is a JSON (possibly unicode-escaped) string ?

In my opinion, the APIKey MUST be a string, not a JSON string, so, no need to unmarshal it. Or make it base64 encoded.

hub/utils/auth/auth.go
```diff
@@ -29,3 +31,72 @@ func CheckForCreds(cmd *cobra.Command, currentSession *sessionstore.HubSession,
...
+		return effectiveClientID, effectiveSecret, "command opts"
+	}
+
+	// Only use env vars if both are provided
+	envClientID := os.Getenv("DIRCTL_CLIENT_ID")
+	envSecret := os.Getenv("DIRCTL_CLIENT_SECRET")
+
+	if envClientID != "" && envSecret != "" {
+		var decodedSecret string
+		if err := json.Unmarshal([]byte(`"`+envSecret+`"`), &decodedSecret); err == nil {
```
jubarbot-cisco: Same remark as before.

jubarbot-cisco (changes requested):

hub/utils/auth/auth.go
gicont: using base64

hub/utils/auth/auth.go
gicont: using base64

jubarbot-cisco (approved):
LGTM

tkircsi (approved):
LGTM

PR #401, opened 2025-09-15
feat: add cid-based synchronisation support

paralta:
This PR introduces the ability to synchronise specific content by CID (Content Identifier) rather than syncing all objects from a remote Directory node.
* Add CID filtering to sync operations - New `--cids` flag in CLI and `cids` field in sync requests allow targeting specific content for synchronisation
* Update sync service API - Modified protobuf definitions and generated code to support CID-based sync requests
* Enhance sync worker logic - Updated sync worker to handle CID filtering during synchronisation operations
* Add e2e test cases - New test cases validate sync-by-tag functionality with test data for different sync scenarios


cli/cmd/sync/sync.go
```diff
@@ -181,3 +204,132 @@ func runDeleteSync(cmd *cobra.Command, syncID string) error {
...
+func parseSearchOutput(input io.Reader) ([]SearchResult, error) {
+	var results []SearchResult
+
+	scanner := bufio.NewScanner(input)
+
+	// Regular expressions to match the search output format
+	recordRegex := regexp.MustCompile(`^Record: (.+)$`)
+	apiAddressRegex := regexp.MustCompile(`^    api address (.+)$`)
+
+	var currentCID string
```
ramizpolic: This is unnecessary and will break as soon as we make changes in the search command. Instead, this would be more useful and easier:
- Add a new flag `--json` to Search, so that output is JSON searlized
- In Sync, use `--stdin` for piping JSON input
- Avoid regexp and only deal with JSON data

Piping example: `dirctl routing search --json | dirctl sync create --stdin`

paralta: @ramizpolic thanks for reviewing! fixed now

PR #406, opened 2025-09-16
fix: update apiserver.env to use new config values

usize:
Keys in the server config had been changed, which broke the docker-compose installer (the api was attempting to connect to zot on localhost instead of the internal docker url zot:5000).


tkircsi (approved):

ramizpolic (approved):
nice catch, thanks @usize! also raises an important point on testing the compose setup since all our tests are based around kind

PR #429, opened 2025-09-18
feat: port sign/verify to hub command

ramizpolic:
## Overview

This PR ports the signing and verification logic to hub command. It is intended as a temporary fix until the actual signing and verification from the main dirctl command works in the following way as well.

This logic also ensures to attach public key and signature to the record/agent signature so that later, these can be migrated natively to OCI for each record.

NOTE: We must sign the data in OCI cosign signature format, otherwise during migration, we will have to recompute the signatures. This way, the signing between versions is preserved.

**Testing**

```
cat e2e/shared/testdata/record_v031.json | bin/dirctl hub sign --stdin | bin/dirctl hub verify --stdin
```


tkircsi (approved):
LGTM

jubarbot-cisco (approved):
LGTM.
Tested.

Thanks !

PR #437, opened 2025-09-22
ci(homebrew): use local tap for testing homebrew formulae

arpad-csepi:
The homebrew binary beyond the 4.6.3 version break the local file installation which our CI relies on for testing the updated formulae. This fix is creating a new local tap for testing and copy the formula from the repository into it


lgecse (approved):

ramizpolic (approved):
LGTM

PR #456, opened 2025-09-26
feat: add validator

adamtagscherer:
This PR adds the validation to the store's push method, and after sync operation. It also updates the test JSON files to conform to the JSON schemas and updates the tests accordingly.

### DIFF
--- a/e2e/shared/testdata/record_v031.json
+++ b/e2e/shared/testdata/record_v031.json
@@ -3,7 +3,9 @@
   "version": "v1.0.0",
   "schema_version": "v0.3.1",
   "description": "Research agent for Cisco's marketing strategy.",
-  "authors": ["Cisco Systems"],
+  "authors": [
+    "Cisco Systems"
+  ],
   "created_at": "2025-03-19T17:06:37Z",
   "annotations": {
     "key": "value"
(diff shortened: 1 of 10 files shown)


e2e/shared/testdata/record_v031.json
```diff
@@ -54,5 +56,16 @@
...
+    "certificate": "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0t",
+    "content_bundle": "eyJ0ZXN0IjogInZhbHVlIn0=",
+    "content_type": "application/json",
+    "signature": "MEUCIQDTest123Signature456789",
+    "signed_at": "2025-09-11T10:00:00Z",
+    "annotations": {
+      "signer": "test-authority",
+      "purpose": "testing"
+    }
+  }
```
paralta, line 70: does the validation require signature? on the directory side the signature is a referrer external to the record

paralta (approved):
Small comment on the signature, otherwise looks good to me 👍

e2e/shared/testdata/record_v031.json
adamtagscherer: For schema version 0.3.1 the signature is a required field, see: https://schema.oasf.outshift.com/0.3.1/objects/agent

PR #457, opened 2025-09-26
feat: secure credential management and restrictive access control for zot sync operations

paralta:
This PR implements credential management for sync operations between Directory nodes, enabling secure and dynamic authentication with remote registries.

* **Implemented restrictive zot access control:** Added htpasswd authentication with role-based permissions where admin users (apiserver) have full access (read, create, update, delete) while sync users provided to remote nodes (or any other authenticated users) have read-only access
* **Implemented credential negotiation:** Updated sync worker to properly negotiate and pass credentials to zot configuration, replacing hardcoded authentication with dynamic credential management
* **Refactored zot configuration management:** Moved zot sync configuration logic from server/sync/worker.go to dedicated utility package utils/zot/ with functions for adding/removing registries and managing credentials


install/charts/dir/apiserver/values.yaml
```diff
@@ -64,6 +64,26 @@ config:
...
+    worker_timeout: "10m"
+    
+    # Registry monitor configuration
+    registry_monitor:
+      check_interval: "30s"
+    
+    # Authentication configuration for sync operations
+    auth_config:
+      username: ""
+      password: ""
```
ramizpolic: why do we need to configure sync auth config? is not done automatically?

utils/zot/credentials.go
```diff
@@ -0,0 +1,59 @@
...
+	zotsyncconfig "zotregistry.dev/zot/pkg/extensions/config/sync"
+)
+
+const (
+	// DefaultCredentialsPath is the default path to the zot credentials file.
+	DefaultCredentialsPath = "/etc/zot/credentials.json" //nolint:gosec
+)
+
+// updateCredentialsFile updates a credentials file for zot sync.
+func updateCredentialsFile(filePath string, remoteRegistryURL string, credentials zotsyncconfig.Credentials) error {
```
ramizpolic, line 22: is this file being updated via API calls? if so, this creates a race-condition

utils/zot/config.go
```diff
@@ -0,0 +1,248 @@
...
+	if err := os.WriteFile(filePath, updatedConfig, 0o644); err != nil { //nolint:gosec,mnd
+		return fmt.Errorf("failed to write updated zot config: %w", err)
+	}
+
+	return nil
+}
+
+// addRegistryToSyncConfig adds a registry to the zot sync configuration.
+func AddRegistryToSyncConfig(filePath string, remoteRegistryURL string, remoteRepositoryName string, credentials zotsyncconfig.Credentials, cids []string) error {
+	logger.Debug("Adding registry to zot sync", "remote_url", remoteRegistryURL)
```
ramizpolic, line 52: this can cause race-condition as well if its called from APIs

ramizpolic: i am not sure i understand what the flow is when creating the sync

install/charts/dir/apiserver/values.yaml
paralta: @ramizpolic thanks for reviewing! indeed, let me clean this up

utils/zot/credentials.go
paralta: we use a scheduler and only one worker architecture to process sync operations which prevents race conditions, even when multiple API requests are made concurrently.

utils/zot/config.go
paralta: the sync creation flow did not change in this PR, we just added the credential file. the flow (client creates sync request, scheduler picks up sync request, worker updates zot sync config file with new registry to sync) remains intact

utils/zot/credentials.go
paralta: this is an issue if the number of workers is increased though. I think this is outside the scope of the current changes

ramizpolic: i would still suggest to address this behaviour. adding one global lock mutex to managing zot config is enough. not having any docs on this behaviour is problematic

server/config/config.go
```diff
@@ -176,6 +176,9 @@ func LoadConfig() (*Config, error) {
 	_ = v.BindEnv("sync.registry_monitor.check_interval")
 	v.SetDefault("sync.registry_monitor.check_interval", syncmonitor.DefaultCheckInterval)
 
+	_ = v.BindEnv("sync.auth_config.username")
+	_ = v.BindEnv("sync.auth_config.password")
+
```
ramizpolic, line 181: there are still leftovers from sync authconfig. is this meant to be here?

utils/zot/credentials.go
paralta: I will report a bug in this 👍

paralta: > not having any docs on this behaviour is problematic

can you expand on what details need to be highlighted and where?

server/config/config.go
paralta: these are required, please check the Taskfile where `DIRECTORY_SERVER_SYNC_AUTH_CONFIG_USERNAME` and `DIRECTORY_SERVER_SYNC_AUTH_CONFIG_PASSWORD` environment variables are provided.

paralta: the sync credentials are created on deploy like the zot admin credentials

ramizpolic (approved):
LGTM

server/config/config.go
ramizpolic: @paralta in that case, the sync config in the helm values file are required. 
we also need those inside the docker compose config `install/docker`

Issue #459, opened 2025-09-29
[Bug]: CID-based Registry Sync Missing Cosign Signature Referrers


paralta:
### Bug Description

When registry synchronization is configured to sync specific CIDs/tags, the sync process successfully copies the main record artifacts and public keys from the remote registry, but fails to copy the associated cosign signature referrers. This results in records that cannot be verified as trusted, even though they have valid signatures in the source registry.

### Expected Behavior

CID-based sync should copy both records and their associated signature referrers to enable trusted verification.

### Affected Version

v1.0.0

### Steps to Reproduce

**Scenario 1: Full Repository Sync (Working)**
1. Configure registry sync for entire repository (no specific CIDs)
2. Trigger sync operation
3. Observe that all records, public keys, and signature referrers are copied
4. Verify records using VerifyWithZot - should return `IsTrusted: true`

**Scenario 2: CID-based Sync (Broken)**
1. Configure registry sync to sync specific CIDs
2. Trigger sync operation
3. Observe in logs that only the CID tag and public key artifacts are copied
4. Attempt to verify records using VerifyWithZot - should return `IsTrusted: false`
5. Check local registry - signature referrers are missing

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

PR #464, opened 2025-09-29
fix(utils/cosign): migrate cosign signature attachment to OCI 1.1 referrers spec

paralta:
This PR updates the signature attachment method to leverage the OCI 1.1 specification's referrers API to fix https://github.com/agntcy/dir/issues/459.

* **Signature attachment method:** Switched from `ociremote.WriteSignatures()` to [`ociremote.WriteSignaturesExperimentalOCI()`](https://github.com/sigstore/cosign/blob/main/pkg/oci/remote/write.go#L140) in cosign utils to use OCI 1.1 behavior for signature referrers
* **Signature discovery:** Updated signature pulling logic to use OCI referrers API instead of cosign tag-based discovery (`sha256-abc123.sig` format)
* **Test coverage:** Extended E2E tests to include signature verification after sync operations for both full sync and sync by CID

### DIFF
--- a/client/verify.go
+++ b/client/verify.go
@@ -100,7 +100,7 @@ func (c *Client) pullSignatureReferrer(ctx context.Context, recordCID string) (*
 	}
 
 	signature := response.GetSignature()
-	if signature == nil {
+	if signature == nil || signature.GetSignature() == "" {
 		return nil, errors.New("no signature found in referrer response")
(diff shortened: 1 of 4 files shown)


paralta:
@tkircsi thanks for reviewing and for the suggested improvements 🚀

PR #465, opened 2025-09-30
feat: add mcp server

adamtagscherer:
This PR creates an MCP server for the directory.


mcp/schemas/oasf-0.7.0.json
ramizpolic, line 1: i would suggest not to implement this part here. the `oasf-sdk` can be extended with a method to fetch a specific schema, like:

```go
schemaBytes, err := validator.GetSchema("0.7.0")
```

mcp/README.md
```diff
@@ -0,0 +1,51 @@
...
+[Model Context Protocol (MCP)](https://modelcontextprotocol.io/) server for creating and validating OASF agent records.
+
+## Tools
+
+### `agntcy_oasf_create_record`
+
+Creates an OASF agent record from JSON and returns its CID.
+
+**Input:** `record_json` (string)  
+**Output:** `success` (bool), `cid` (string), `schema_version` (string), `error_message` (string)
```
ramizpolic: The `create_record` tool is supposed to generate a valid `record_json` from any of these inputs:
- Codebase (via `input_dir` arg)
- Text (via `input_text` arg)

The `create_record` tool is not supposed to generate CID or anything like that. It always needs to return a valid OASF Record JSON object.

adamtagscherer: Okay so after some thinking I ended up removing the `agntcy_oasf_create_record` tool.

Why?

So the only thing it did was to calculate the CID for the record that it got in a parameter.
I found out that we don't need any tool to create a record, because the resources - JSON Schema 07 for 0.3.1 and 0.7.0 - for it are exposed and the Agent that uses the MCP can create the records from it.

So right now if I tell the Agent in Cursor or VSCode or whatever, that `create and agent for me from the repository on this URL: ...` or `create me an OASF Record that can do work in the biology field` then the Agent can understand the repository's composition and transform it to a JSON record or it can do the free text to JSON translation and start validate it against the Schemas.

akijakya: I tried it out locally with Cursor CLI and it had a hard time creating a valid record with only a validation MCP tool. I think the `GetSchemaContent` and `GetAvailableSchemaVersions` functions in oasf-sdk introduced in [this PR](https://github.com/agntcy/oasf-sdk/pull/74/files) should also be exposed as separate tools in the MCP server so the agent can pull what schemas are available, and the schema the record should be valid against.

Edit: I realize these schemas are added to the MCP as resources but I couldn't get my agent to read them before creating a record, so maybe turning them into tools would still be a better approach?

adamtagscherer: Resources are created for this reason to provide files and other things as context to models: https://modelcontextprotocol.io/specification/2025-06-18/server/resources

Quoting from the website:

`Resources allow servers to share data that provides context to language models, such as files, database schemas, or application-specific information. Each resource is uniquely identified by a [URI](https://datatracker.ietf.org/doc/html/rfc3986).`

`Tools enable models to interact with external systems, such as querying databases, calling APIs, or performing computations.`

I think a JSON Schema 07 is exactly a file that should be provided for the agent as a context.
I also experience that it's hard for the agent to discover that this resource is available for it, but it can be solved by explicitly saying to it that it's provided. I think converting this to a tool won't make it more discoverable for the agent and also tools are not for this purpose as of my understanding.

Let's talk about it today and figure out how to progress with this.

ramizpolic: I agree with @akijakya, we should still have a tool `create_record` to be able to create records. The tool can use schema in the context + prompt on how to perform record creation from input.

mcp/mcp
ramizpolic, line 1: lets remove the binary from package

mcp/main.go
```diff
@@ -0,0 +1,69 @@
...
+	"log"
+	"strings"
+
+	"github.com/agntcy/dir/mcp/tools"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+func main() {
+	// Create MCP server for Directory operations
+	server := mcp.NewServer(&mcp.Implementation{
```
ramizpolic, line 17: we are still missing `create_record` tool, this is a requirement for our MCP server

ramizpolic (dismissed):

mcp/main.go
adamtagscherer: Not sure if I fully understand why we need this tool.
With `agntcy_oasf_get_schema`, an AI assistant has complete access to all required fields, domains, taxonomies, etc.

Record creation is just JSON construction:
- Get the schema → `agntcy_oasf_get_schema`
- Build a JSON object following the schema structure
- Validate it → `agntcy_oasf_validate_record`

LLMs excel at this type of structured data generation.

If we don't implement `agntcy_oasf_create_record` we get the following benefits:
- we don't have to maintain another tool
- no need to update it every time when OASF schema version evolves
- schema changes automatically propagate to AI generating tools

The tool doesn't add meaningful functionality beyond what an AI can do with the schema.

Can you elaborate why would you like to have this tool implemented? what would it do? how do you imagine this tool being implemented?

akijakya (approved):
LGTM 🚀

mcp/main.go
adamtagscherer: Here is a picture running the prompt in the README.md in random repository `I want to create a new OASF agent record from the repository in this directory. Can you help me?`.
You can see that it gets the available OASF versions, gets the actual schema for 0.7.0, creates the record JSON and validates it.

Screenshot 2025-10-19 at 11 31 34

ramizpolic (approved):

Issue #353, opened 2025-09-03
[Feature]: Expose Event Listener API via Routing


ramizpolic:
### Problem Statement

Currently, all the announcement events across the networks are handled internally in the Routing handler for the sake of constructing the routing tables. Some applications however may need to listen for these events in order to perform additional operations. Consider the flow where an app is listening for the newest agent with a given skill to perform some app-specific operation.

### Proposed Solution

On the Routing service, expose another RPC for event listening, such as an example below

```grpc
service RoutingService {
    rpc Listen(ListenRequest) (stream ListenResponse);
}

message ListenRequest {
    repeated SearchQuery queries = 1;
}

message ListenResponse {
    string record_cid = 1;
    string peer_addr = 2;
    repeated string labels = 3;
}
```

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

PR #472, opened 2025-10-01
feat(dir): add GossipSub for efficient label announcements

tkircsi:
## Problem

The previous **pull-based label discovery** architecture had significant limitations:

- **High bandwidth usage**: Remote peers had to pull entire records (KB-MB) just to extract labels
- **Limited propagation**: Only k-closest DHT peers (~20) received provider notifications
- **Lazy discovery**: Labels were only cached when DHT provider notifications were received
- **Network inefficiency**: Each new remote record required a full network pull operation

## Solution

Implemented a **hybrid dual-protocol architecture** that combines GossipSub with the existing DHT infrastructure:

- **GossipSub** (primary): Direct label announcements propagated to ALL subscribed peers
- **DHT.Provide()** (existing): Content discovery and provider tracking (kept as-is)
- **DHT+Pull** (fallback): Automatic fallback when GossipSub unavailable

### Core Implementation

**New Package** (`server/routing/pubsub/`):
- `constants.go` - Protocol constants (topic, message size limits)
- `messages.go` - LabelAnnouncement wire format with validation
- `manager.go` - GossipSub manager with lifecycle management

**Modified Files**:
- `routing_remote.go` - Integrated GossipSub, extracted shared publishing logic
- `cleanup_tasks.go` - Updated republishing to include GossipSub labels
- `routing/config/config.go` - Added GossipSubConfig (enable/disable only)
- `server/config/config.go` - Added viper configuration bindings

### Architecture Improvements

**DRY Refactoring**: Extracted `publishToNetwork()` method used by both initial publishing and periodic republishing tasks

**Callback Injection**: CleanupManager uses function injection to reuse publishing logic without circular dependencies

**Protocol Constants**: Topic name and message size are hardcoded (not configurable) to prevent network fragmentation

### Configuration

Only enable/disable is configurable (protocol parameters are constants):

```yaml
routing:
  gossipsub:
    enabled: true  # Default: true
```

Environment variable: `DIRECTORY_SERVER_ROUTING_GOSSIPSUB_ENABLED=true`

### DIFF
--- a/server/types/label_test.go
+++ b/server/types/label_test.go
@@ -74,7 +74,7 @@ func TestGetLabelsFromRecord(t *testing.T) {
 		recordJSON := `{
 			"name": "test-agent-v2",
 			"version": "2.0.0",
-			"schema_version": "v0.7.0",
+			"schema_version": "0.7.0",
 			"authors": ["test"],
 			"created_at": "2023-01-01T00:00:00Z",
 			"skills": [


server/routing/pubsub/manager.go
```diff
@@ -0,0 +1,282 @@
...
+	ctx         context.Context //nolint:containedctx // Needed for long-running message handler goroutine
+	host        host.Host
+	pubsub      *pubsub.PubSub
+	topic       *pubsub.Topic
+	sub         *pubsub.Subscription
+	localPeerID string
+	topicName   string // Topic name (protocol constant)
+
+	// Callback invoked when label announcement is received
+	onLabelAnnouncement func(context.Context, *LabelAnnouncement)
```
ramizpolic: I feel like we are doing this the other way. Instead of doing `label` announcement, its actually `record` announcement with extra data, one of which is labels themselves.

IMO this would be clearer if we had `PublishEvent` as a general publication event, or even more specifically, `RecordPublishEvent`

 
```suggestion
	onRecordAnnouncement func(context.Context, *RecordPublishEvent)
```

server/routing/pubsub/manager.go
```diff
@@ -0,0 +1,282 @@
...
+// Parameters:
+//   - ctx: Context for operation timeout/cancellation
+//   - cid: Content ID of the record
+//   - labelList: List of labels extracted from the record
+//
+// Returns:
+//   - error: If validation or publishing fails
+//
+// Note: This is non-blocking. GossipSub handles propagation asynchronously.
+func (m *Manager) PublishLabels(ctx context.Context, cid string, labelList []labels.Label) error {
```
ramizpolic: perhaps a better way to do it is that we pass recordRef or a full record from which we construct the full publication event

```suggestion
func (m *Manager) PublishRecord(ctx context.Context, ref RecordRef) error {
```

server/routing/pubsub/manager.go
```diff
@@ -0,0 +1,282 @@
...
+// Example:
+//
+//	manager.SetOnLabelAnnouncement(func(ctx context.Context, ann *LabelAnnouncement) {
+//	    for _, labelStr := range ann.Labels {
+//	        label := labels.Label(labelStr)
+//	        key := BuildEnhancedLabelKey(label, ann.CID, ann.PeerID)
+//	        // ... store in datastore ...
+//	    }
+//	})
+func (m *Manager) SetOnLabelAnnouncement(fn func(context.Context, *LabelAnnouncement)) {
```
ramizpolic: likewise, in this method, we can register event listeners for record publications


```suggestion
func (m *Manager) SetRecordAnnouncementCallback(fn func(context.Context, *RecordPublishEvent)) {
```

server/routing/pubsub/manager.go
```diff
@@ -0,0 +1,282 @@
...
+func (m *Manager) GetTopicPeers() []string {
+	peers := m.topic.ListPeers()
+	peerIDs := make([]string, len(peers))
+
+	for i, p := range peers {
+		peerIDs[i] = p.String()
+	}
+
+	return peerIDs
+}
```
ramizpolic, line 282: i do not know if it is efficient to have only one topic across the network. i think for the time being, its okay, but later on, it would be useful to subscribe to a set of topics regarding the publication data (e.g. specific labels -- skills, domains, etc) which can be policy-driven via config in the server.
note that in this case, this method will return all the peers for the topic across the whole network

server/routing/pubsub/messages.go
```diff
@@ -0,0 +1,110 @@
...
+//
+// Example:
+//
+//	{
+//	  "cid": "bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi",
+//	  "peer_id": "12D3KooWD3bfmNbuuuT5Zch8fj9Cg9dQR2FpGm7JzCfCzPWZnxLn",
+//	  "labels": ["/skills/AI/ML", "/domains/research", "/modules/tensorflow"],
+//	  "timestamp": "2025-10-01T10:00:00Z"
+//	}
+type LabelAnnouncement struct {
```
ramizpolic: i would suggest using explicit event for record publications, eg


```suggestion
type RecordAnnouncement struct {
```

or 


```suggestion
type RecordPublishEvent struct {
```

server/routing/pubsub/messages.go
ramizpolic, line 1: can we move these types to `server/routing/pubsub/types/events.go` or `server/routing/types/events.go`? IMO it makes it cleaner to separate the business logic from message types used across the interfaces

server/routing/cleanup_tasks.go
```diff
@@ -46,19 +45,34 @@ func (f *remoteLabelFilter) Filter(e query.Entry) bool {
...
+// Parameters:
+//   - dstore: Datastore for label storage
+//   - storeAPI: Store API for record operations
+//   - server: P2P server for DHT operations
+//   - publishFunc: Callback for publishing (from routeRemote.publishToNetwork)
+func NewCleanupManager(
+	dstore types.Datastore,
+	storeAPI types.StoreAPI,
+	server *p2p.Server,
+	publishFunc func(context.Context, string, *corev1.Record) error,
```
ramizpolic: this function type can be defined in `server/routing/types` such as

```go
type PublishEventHandler func(...) (...)
```

server/routing/routing_remote.go
```diff
@@ -168,7 +171,30 @@ func newRemote(ctx context.Context,
 
 	routeAPI.service = rpcService
 
-	routeAPI.cleanupManager = NewCleanupManager(dstore, storeAPI, server)
+	// Initialize GossipSub manager if enabled
+	// Protocol parameters (topic, message size) are defined in pubsub.constants
+	// and are NOT configurable to ensure network-wide compatibility
+	if opts.Config().Routing.GossipSub.Enabled {
+		pubsubManager, err := pubsub.New(ctx, server.Host())
```
ramizpolic: the pubsubManager is never closed on success (somewhere in the routing remote) which can result in potential memory leak and network-hogging

server/routing/routing_remote.go
```diff
@@ -179,26 +205,73 @@ func newRemote(ctx context.Context,
...
+	return nil
+}
+
+func (r *routeRemote) Publish(ctx context.Context, ref *corev1.RecordRef, record *corev1.Record) error {
+	remoteLogger.Debug("Publishing record to network", "ref", ref)
+
+	// Use shared publishing logic
+	if err := r.publishToNetwork(ctx, ref.GetCid(), record); err != nil {
+		return status.Errorf(codes.Internal, "%v", err)
+	}
```
ramizpolic: there is no need to have 2 identical functions here, lets keep one

server/routing/pubsub/manager.go
tkircsi: That's a good idea. Let's think about the scope and create an issue for it.

server/routing/pubsub/messages.go
tkircsi: I'd move `messages.go` and the `PublishEventHandler` type declaration into a file `events.go` instead of creating sub-packages. This is a tiny package and no need to split yet.

server/routing/cleanup_tasks.go
tkircsi: see above

server/routing/routing_remote.go
tkircsi: Nice catch! I need to implement a full Stop service for the routing now.

server/routing/pubsub/manager.go
tkircsi: fixed

server/routing/pubsub/manager.go
tkircsi: fixed

server/routing/pubsub/manager.go
tkircsi: fixed

server/routing/pubsub/messages.go
tkircsi: fixed

server/routing/pubsub/messages.go
tkircsi: fixed

server/routing/routing_remote.go
tkircsi: fixed

server/routing/routing_remote.go
tkircsi: fixed

server/routing/pubsub/manager.go
```diff
@@ -0,0 +1,302 @@
...
+//	manager.SetOnLabelAnnouncement(func(ctx context.Context, ann *LabelAnnouncement) {
+//	    for _, labelStr := range ann.Labels {
+//	        label := labels.Label(labelStr)
+//	        key := BuildEnhancedLabelKey(label, ann.CID, ann.PeerID)
+//	        // ... store in datastore ...
+//	    }
+//	})
+func (m *Manager) SetOnRecordPublishEvent(fn func(context.Context, *RecordPublishEvent)) {
+	m.onRecordPublishEvent = fn
+}
```
ramizpolic, line 204: i like that we are capable of registering handlers for publication events, but this approach will always override if we want to register multiple handlers. 

wdyt if we support multiple handlers instead, so other places can also register for event listening
```
func (m *Manager) RegisterPublishEventHandler(handlerName string, handlerFn PublishEventHandler) error { 
	m.mu.Lock()
	defer m.mu.Unlock()

	if _, has := s.handlers[handlerName]; has {
		return fmt.Errorf("handler %s already registered", handlerName)
	}
	s.handlers[handlerName] = handleFn
}
```

install/charts/dir/apiserver/values.yaml
```diff
@@ -64,6 +64,13 @@ config:
     # Datastore to use for routing persistance.
     # datastore_dir: ""
 
+    # GossipSub configuration for efficient label announcements
+    # When enabled, labels are propagated via GossipSub mesh to ALL subscribed peers
```
paralta, line 68: When enabled we still use `DHT+Pull mechanism`, right? I think this could be clearer in this comment here

paralta (approved):
LGTM 👍 small comment

server/routing/pubsub/manager.go
tkircsi: Great idea! I planned to do something similar in this epic. https://github.com/agntcy/dir/issues/353 This can be an event source of the system.

install/charts/dir/apiserver/values.yaml
tkircsi: If GossipSub is enabled, we only use the DHT+Pull mechanism for label announcements as a fallback or in case of a race condition, i.e., if DHT.Provide arrives first.

`handleCIDProviderNotification`:
	// FALLBACK: Labels not cached yet, need to pull record
	// This happens when:
	// - GossipSub message hasn't arrived yet (race condition)
	// - GossipSub is disabled
	// - GossipSub message was lost
	// - Peer doesn't support GossipSub
	
`handleRecordPublishEvent`:
// handleRecordPublishEvent processes incoming record publication events from GossipSub.
// This is the primary label discovery mechanism when GossipSub is enabled.

`handleCIDProviderNotification` checks whether the given labels are already cached, and if not, it simply calls Pull.

Issue #454, opened 2025-09-25
[Feature]: Use encoded signature and key referrer objects in Store API


ramizpolic:
### Problem Statement

Instead of using current type-based service imports for Signature and public key objects in the Storage API Pull/Push Referrer RPCs, we should switch to encoded object handling. Managing service object imports in the Store API creates the same problem of cross-service coupling that we had with support for multiple versions of Record objects.

### Proposed Solution

- Use native referrer objects in the Store API without the type-bindings from other services:

```proto
message PullReferrerRequest {
  // Record reference
  core.v1.RecordRef record_ref = 1;

  oneof options {
    // Record referrer type to be pulled
    string referrer_type = 2;
  }
}

// PushReferrerRequest represents a record with optional OCI artifacts for push operations.
message PushReferrerRequest {
  // Record reference
  core.v1.RecordRef record_ref = 1;

  // RecordReferrer object to be stored for the record
  core.v1.RecordReferrer referrer = 2;
}
```

- Encode `Signature` proto object into `RecordReferrer` with `RecordReferrer{type=agntcy.dir.sign.v1.Signature, data=structpb.Marshal(signature)}`.
- Encode `PublicKey` data into `RecordReferrer` with `RecordReferrer{type=agntcy.dir.sign.v1.PublicKey, data=structpb.Marshal(key)}`.e 

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

PR #480, opened 2025-10-01
feat(dir/server/store): handle generic referrers

paralta:
This PR refactors the OCI store to implement generic referrer methods and uses these new methods to handle public keys, while maintaining existing signature workflows.

* **Fix misleading log messages -** Updated publish command output to clarify that records are not immediately discoverable and require publication service processing
* **Update method signatures -** Changed PullSignature and PullPublicKey methods to return slices for handling multiple artifacts
* **Implement generic referrer push and pull methods -** Added reusable referrer handling infrastructure in the OCI store with configurable media types
* **Refactor public key handling -** Migrated public key operations to use the new generic referrer system instead of custom implementation

The signature handling was intentionally not refactored since signature artifact creation and push to the registry is managed by cosign, maintaining the existing proven workflow for signature operations.

### DIFF
--- a/client/sign.go
+++ b/client/sign.go
@@ -153,6 +153,18 @@ func (c *Client) SignWithKey(ctx context.Context, req *signv1.SignRequest) (*sig
 }
 
 func (c *Client) pushReferrersToStore(ctx context.Context, recordCID string, signature *signv1.Signature, publicKey string) error {
+	if recordCID == "" {
+		return errors.New("record CID is required")
+	}
+
+	if publicKey == "" {
+		return errors.New("public key is required")
+	}
+
+	if signature == nil || signature.GetSignature() == "" {
+		return errors.New("signature is required and must not be empty")
+	}
+
 	// convert public key to struct
 	publicKeyStruct, err := structpb.NewStruct(map[string]any{
 		"publicKey": publicKey,
--- a/server/controller/store.go
+++ b/server/controller/store.go
@@ -274,9 +274,18 @@ func (s storeCtrl) pushSignatureReferrer(ctx context.Context, request *storev1.P
 func (s storeCtrl) pushReferrer(ctx context.Context, request *storev1.PushReferrerRequest) *storev1.PushReferrerResponse {
 	storeLogger.Debug("Pushing referrer", "cid", request.GetRecordRef().GetCid(), "type", request.GetReferrer().GetType())
 
+	if request.GetReferrer() == nil {
+		errMsg := "referrer is required"
+
+		return &storev1.PushReferrerResponse{
+			Success:      false,
+			ErrorMessage: &errMsg,
+		}
+	}
+
 	if request.GetReferrer().GetType() == ociutils.PublicKeyArtifactMediaType {
-		publicKeyValue, ok := request.GetReferrer().GetData().AsMap()["publicKey"]
-		if !ok {
+		publicKey, err := request.GetReferrer().GetPublicKey()
+		if err != nil {
 			errMsg := "publicKey field not found in referrer data"
 
 			return &storev1.PushReferrerResponse{
@@ -285,17 +294,7 @@ func (s storeCtrl) pushReferrer(ctx context.Context, request *storev1.PushReferr
 			}
 		}
 
-		publicKey, ok := publicKeyValue.(string)
-		if !ok {
-			errMsg := "publicKey field is not a string"
-
-			return &storev1.PushReferrerResponse{
-				Success:      false,
-				ErrorMessage: &errMsg,
-			}
-		}
-
-		err := s.uploadPublicKey(ctx, request.GetRecordRef().GetCid(), publicKey)
+		err = s.uploadPublicKey(ctx, request.GetRecordRef().GetCid(), publicKey)
 		if err != nil {
 			errMsg := fmt.Sprintf("failed to upload public key: %v", err)
 
@@ -338,6 +337,10 @@ func (s storeCtrl) pushReferrer(ctx context.Context, request *storev1.PushReferr
 
 func (s storeCtrl) uploadPublicKey(ctx context.Context, recordCID string, publicKey string) error {
 	storeLogger.Debug("Uploading public key", "cid", recordCID)
+
+	if publicKey == "" {
+		return errors.New("public key is required")
+	}
 
 	ociStore, ok := s.store.(interface {
 		UploadPublicKey(context.Context, string) error


server/types/store.go
```diff
@@ -34,12 +34,21 @@ type SignatureStoreAPI interface {
...
+	// Push referrer to content store
+	PushReferrer(context.Context, string, *corev1.RecordReferrer) error
+
+	// Pull referrers from content store by type
+	PullReferrersByType(context.Context, string, string) ([]*corev1.RecordReferrer, error)
 
-	// Pull public key from content store
-	PullPublicKey(context.Context, string) (string, error)
+	// Pull all referrers from content store
+	PullAllReferrers(context.Context, string) ([]*corev1.RecordReferrer, error)
```
ramizpolic: this should return a channel or handle the referrers individually, we shouldnt return all the referrers in one go, it is an expensive operation.

consider that OCI gives you a way to _walk_ the referrers individually for a given object, which means that a more suitable method here could be

```suggestion
	WalkReferrers(ctx context.Context, cid core.RecordRef, refType string, walkFn func(referrerData)) error
```

this gives you a way to implement pull refererrs nicely by specifying the walkFn to return in the controller, something like

```go
// store_controller.go

func PullReferrer(inReq) {
	walkFn := func(referrerData) {
		// return fetched data via controller
		controller <- referrerData
	}

	err := store.WalkReferrer(ctx, inReq.Ref, inReq.type, walkFn)
	if err != nil {
		// did not process all referrers
	}

	// indicate that the controller finished
	controller <- done
}

```

cli/cmd/pull/pull.go
```diff
@@ -77,15 +86,25 @@ func runCommand(cmd *cobra.Command, cid string) error {
...
+			Options: &storev1.PullReferrerRequest_PullReferrerType{
+				PullReferrerType: ociutils.PublicKeyArtifactMediaType,
 			},
 		})
 		if err != nil {
 			return fmt.Errorf("failed to pull public key: %w", err)
 		}
 
-		presenter.Print(cmd, "Public key: "+response.GetPublicKey())
+		publicKeyValue, ok := response.GetReferrer().GetData().AsMap()["publicKey"]
```
ramizpolic: avoid doing this, lets have types defined for all the referrer objects internally, including public key which is just the payload string

i would expect logic defined for referrer object handling so that converting between `pbstruct` and appropriate object type is possible. something like:

```go
// note: it would be useful to define this as a method
// on RecordReferrer, but causes a cyclic import

// api/types/referrer.go

type ReferrerObjectType uint

const (
	SignatureReferrerObjectType = iota + 1
	PubKeyReferrerObjectType
)

type ReferrerObject interface {
	GetType() ReferrerObjectType
	GetSignature() (signature, error)
	GetPublicKey() (pubKey, error)
}

func New(r RecordReferrer) ReferrerObject {

} 

func (r ReferrerObject) Marshal() ([]byte, error) {

}

func UnmarshalReferrer([]byte) (ReferrerObject, error) {

}
```

cli/cmd/pull/pull.go
```diff
@@ -77,15 +86,25 @@ func runCommand(cmd *cobra.Command, cid string) error {
...
+		if !ok {
+			return errors.New("publicKey field not found in referrer data")
+		}
+
+		publicKey, ok := publicKeyValue.(string)
+		if !ok {
+			return errors.New("publicKey field is not a string")
+		}
+
+		presenter.Println(cmd, "Public key: "+publicKey)
```
ramizpolic: the goal of output is always to allow it to be piped or stored somewhere. with the `Public key:` prefix, it is difficult to get it as an artifact. IMO we should do one of the two:
- always output JSON data in these cases so that i can store it with `pull <cid> --signature` > sig.json
- use `--json` flag to explicitly define that data output will be printed in json which allows storage in the given format, so that we can `pull <cid> --signature --json > sig.json`

paralta: > the goal of output is always to allow it to be piped or stored somewhere.

at some point we need to review all cmd outputs because this is not the case for most of dir cmds (including `search`, `list`, etc), maybe json output should be our default output?

paralta: let me address this in a different pull request, since this is not in the scope of this one

client/sign.go
```diff
@@ -151,13 +153,24 @@ func (c *Client) SignWithKey(ctx context.Context, req *signv1.SignRequest) (*sig
 }
 
 func (c *Client) pushReferrersToStore(ctx context.Context, recordCID string, signature *signv1.Signature, publicKey string) error {
+	// convert public key to struct
+	publicKeyStruct, err := structpb.NewStruct(map[string]any{
+		"publicKey": publicKey,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to convert public key to struct: %w", err)
+	}
```
ramizpolic, line 174: can we make this an object defined in the gRPC itself, so that we can reuse it for type-safety and as an explicit definition

```
// sign/v1/pubkey.proto

message PublicKey {
	string key = 1;
}
```

ramizpolic: alternativelly, i would not touch this part at all and leave it for #454

cli/cmd/pull/pull.go
paralta: https://github.com/agntcy/dir/issues/485

server/controller/store.go
```diff
@@ -274,39 +271,89 @@ func (s storeCtrl) pushSignatureReferrer(ctx context.Context, request *storev1.P
...
-func (s storeCtrl) pushPublicKeyReferrer(ctx context.Context, request *storev1.PushReferrerRequest) *storev1.PushReferrerResponse {
-	storeLogger.Debug("Pushing public key referrer", "cid", request.GetRecordRef().GetCid())
+func (s storeCtrl) pushReferrer(ctx context.Context, request *storev1.PushReferrerRequest) *storev1.PushReferrerResponse {
+	storeLogger.Debug("Pushing referrer", "cid", request.GetRecordRef().GetCid(), "type", request.GetReferrer().GetType())
 
-	// Try to use signature storage if the store supports it
-	sigStore, ok := s.store.(interface {
-		PushPublicKey(context.Context, string, string) error
+	if request.GetReferrer().GetType() == ociutils.PublicKeyArtifactMediaType {
+		publicKeyValue, ok := request.GetReferrer().GetData().AsMap()["publicKey"]
```
tkircsi: Maybe we should add nil check for `GetReferrer` and after `GetData` also.

server/controller/store.go
```diff
@@ -274,39 +271,89 @@ func (s storeCtrl) pushSignatureReferrer(ctx context.Context, request *storev1.P
...
+	storeLogger.Debug("Referrer stored successfully", "cid", request.GetRecordRef().GetCid(), "type", request.GetReferrer().GetType())
 
 	return &storev1.PushReferrerResponse{
 		Success: true,
 	}
 }
 
+func (s storeCtrl) uploadPublicKey(ctx context.Context, recordCID string, publicKey string) error {
+	storeLogger.Debug("Uploading public key", "cid", recordCID)
+
```
tkircsi, line 340: We should add an empty string check for `publicKey`?

client/sign.go
```diff
@@ -151,13 +153,24 @@ func (c *Client) SignWithKey(ctx context.Context, req *signv1.SignRequest) (*sig
 }
 
 func (c *Client) pushReferrersToStore(ctx context.Context, recordCID string, signature *signv1.Signature, publicKey string) error {
+	// convert public key to struct
```
tkircsi, line 168: We should add some check here
```go
	if recordCID == "" {
		return fmt.Errorf("recordCID cannot be empty")
	}
	if signature == nil {
		return fmt.Errorf("signature cannot be nil")
	}
	if publicKey == "" {
		return fmt.Errorf("publicKey cannot be empty")
	}
```

server/controller/store.go
paralta: @tkircsi thanks for reviewing! just updated this 👍

tkircsi: I think, the `if request.GetReferrer() != nil` check is still missing.

PR #491, opened 2025-10-03
feat(dir): encode referrers

paralta:
This PR refactors the Store API to use encoded referrer objects instead of type-based service imports,.

* **Store API Protocol Update:** Modified `PushReferrer` and `PullReferrer` RPCs to handle native `RecordReferrer` objects with `referrer_type` string instead of strongly-typed service objects, removing Sign service dependencies from Store API
* **Referrer Encoding Utilities:** Added utils/referrer package with encode/decode functions for `Signature` and `PublicKey` objects into `RecordReferrer` format


cli/cmd/pull/pull.go
```diff
@@ -82,48 +82,50 @@ func runCommand(cmd *cobra.Command, cid string) error {
...
+		publicKeyType := referrerutils.PublicKeyArtifactMediaType
+
 		resultCh, err := c.PullReferrer(cmd.Context(), &storev1.PullReferrerRequest{
 			RecordRef: &corev1.RecordRef{
 				Cid: cid,
 			},
-			Options: &storev1.PullReferrerRequest_PullReferrerType{
-				PullReferrerType: ociutils.PublicKeyArtifactMediaType,
-			},
+			ReferrerType: &publicKeyType,
```
ramizpolic, line 91: we need to define a type schema for objects first, before we start with type associations. IMO this is an internal detail (how OCI stores things to enable signing/verification) instead of high-level API referrer type management (how we map type -> object). 

first, i would define a generic schema for type associations for our referrer objects, for:
- PublicKey, object defined in `sign/v1/public_key.proto`, with type association `agntcy.dir.sign.v1.PublicKey`
- Signature, object defined in `sign/v1/signature.proto`, with type association `agntcy.dir.sign.v1.Signature`

This enables us to handle specific objects with a common schema, for example when we pull/push referrer with type `agntcy.dir.sign.v1.Signature`, then we can encode/decode the data via `Signature` object.
We should avoid using the OCI types on the high level API because this is only relevant within the service implementation (aka signing service RPCs).

```suggestion
			// Use the Dir Signing Service versioned PublicKey object type using proto message mapping
			// ReferrerType: "agntcy.dir.sign.v1.PublicKey" 
			ReferrerType: (&signv1.PublicKey{})).ProtoReflect().Descriptor().FullName(),
```

The actual artifact for the public key and signature within OCI can be remapped internally, for example:
- PublicKey referrer type "agntcy.dir.sign.v1.PublicKey" on the Dir Store API
- Internal artifact type on the OCI referrer: "application/vnd.agntcy.dir.publickey.v1+pem"

We do not query for `application/vnd.agntcy.dir.publickey.v1+pem` on the high-level API, but our own schema instead `agntcy.dir.sign.v1.PublicKey`. The High Level Referrer Type <=> OCI artifact type is mapped internally inside the implementation.
We can handle this data one in two ways:
- when we pull/push `agntcy.dir.sign.v1.Signature`, we only create internally "application/vnd.dev.cosign.simplesigning.v1+json". when we pull, we search for this artifact type in the OCI and change the type name when returning it to higher level API calls
- when we pull/push `agntcy.dir.sign.v1.Signature`, we store two referrer objects: "application/vnd.agntcy.dir.sign.v1.Signature+json" for our original data so that we can sync and rebuild indexes, and "application/vnd.dev.cosign.simplesigning.v1+json" for artifacts needed by ZOT/other services which we can create during rebuild phase.

ramizpolic: we are also missing the proto definition for the PublicKey object. each service needs to define the schema for the data it needs. in our case, Signing/Verification service requires both signature and publickey schemas

```proto
// sign/v1/public_key.proto
message PublicKey {
	string key = 1;
}
```

utils/referrer/referrer.go
```diff
@@ -0,0 +1,159 @@
...
+const (
+	// PublicKeyArtifactMediaType defines the media type for public key blobs.
+	PublicKeyArtifactMediaType = "application/vnd.agntcy.dir.publickey.v1+pem"
+
+	// SignatureArtifactType defines the media type for cosign signature layers.
+	SignatureArtifactType = "application/vnd.dev.cosign.simplesigning.v1+json"
+
+	// ReferrerArtifactMediaType defines the media type for referrer blobs.
+	DefaultReferrerArtifactMediaType = "application/vnd.agntcy.dir.referrer.v1+json"
+)
```
ramizpolic: these are internal OCI artifact types which is slightly different from ReferrerType.
we can simplify the encoding/decoding flow with a generic handler for these ops. We only need to define a generic interface for referrer objects to support casting and operations:

```go
// Defined in: api/types/referrer.go

// Defines an interface for referrer objects
//
// TODO: Implement the interface for Signature and PublicKey types in
// api/store/v1/signature.go and api/store/v1/public_key.go
type ReferrerObject interface {
	// Loads the object from struct data
	// Example for signature: decoder.ProtoToStruct[signv1.Signature](ref.GetData())
	// https://github.com/agntcy/oasf-sdk/blob/25639f0bc13156dd093a1ea432fe6dc2a9a8028d/pkg/decoder/object.go#L32
	UnmarshalReferrer(corev1.ReferrerObject) error

	// Exports the object into a generic referrer object
	// Populates only referrer type and data, and optionally annotations if possible.
	// Type: ref.ReferrerType()
	// Data: decoder.StructToProto(ref)
	// https://github.com/agntcy/oasf-sdk/blob/25639f0bc13156dd093a1ea432fe6dc2a9a8028d/pkg/decoder/object.go#L23
	MarshalReferrer() (*corev1.ReferrerObject, error)

	// Returns referrer type
	// Signature: "agntcy.dir.sign.v1.Signature"
	// PublicKey: "agntcy.dir.sign.v1.PublicKey"
	ReferrerType() string
}
```

utils/referrer/referrer.go
```diff
@@ -0,0 +1,159 @@
...
+
+// EncodePublicKeyToReferrer creates a RecordReferrer for a public key.
+func EncodePublicKeyToReferrer(publicKey string) (*corev1.RecordReferrer, error) {
+	publicKeyStruct, err := structpb.NewStruct(map[string]any{
+		"publicKey": publicKey,
+	})
```
ramizpolic: we should leverage native decoder ops we have and use explicitly typed objects rather than dynamic maps

please check https://github.com/agntcy/oasf-sdk/blob/main/pkg/decoder/object.go for whats available

cli/cmd/pull/pull.go
paralta: agree with the OCI vs API referrer type distinction and the public key message ✅ about the signature, I suggest for now we leverage only the signature created by cosign as required by zot and extend later if required

server/store/oci/types.go
```diff
@@ -0,0 +1,45 @@
...
+func ociToAPIType(ociType string) string {
+	switch ociType {
+	case SignatureArtifactType:
+		return corev1.SignatureReferrerType
+	case PublicKeyArtifactMediaType:
+		return corev1.PublicKeyReferrerType
+	default:
+		return ociType // Return the original OCI type if not found
+	}
+}
```
ramizpolic, line 45: 👍

sdk/dir-py/agntcy/dir_sdk/models/sign_v1.py
```diff
@@ -1,5 +1,6 @@
 # Copyright AGNTCY Contributors (https://github.com/agntcy)
 # SPDX-License-Identifier: Apache-2.0
 
+from agntcy.dir.sign.v1.signature_pb2 import *
```
ramizpolic, line 4: we are missing the publickey

`from agntcy.dir.sign.v1.public_key_pb2 import *`

sdk/dir-js/src/utils.ts
```diff
@@ -0,0 +1,232 @@
...
+import type { Signature } from '@buf/agntcy_dir.bufbuild_es/agntcy/dir/sign/v1/signature_pb';
+import {
+    RecordReferrer,
+    RecordReferrerSchema,
+} from '@buf/agntcy_dir.bufbuild_es/agntcy/dir/core/v1/record_pb';
+import type { JsonObject } from '@bufbuild/protobuf';
+
+// Referrer type constants
+export const SIGNATURE_REFERRER_TYPE = 'agntcy.dir.sign.v1.Signature';
+export const PUBLIC_KEY_REFERRER_TYPE = 'agntcy.dir.sign.v1.PublicKey';
```
ramizpolic, line 21: we can export these values from `models/sign_v1` package directly

sdk/dir-js/src/utils.ts
ramizpolic, line 1: we actually dont need to do any encoding/decoding ourselves, all this can be handled using structproto from golang. we only need to specify the type when constructing/deconstructing messages

sdk/dir-js/src/models/sign_v1.ts
```diff
@@ -2,3 +2,4 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export * from '@buf/agntcy_dir.bufbuild_es/agntcy/dir/sign/v1/sign_service_pb';
+export * from '@buf/agntcy_dir.bufbuild_es/agntcy/dir/sign/v1/signature_pb';
```
ramizpolic, line 5: missing public key here

`export * from '@buf/agntcy_dir.bufbuild_es/agntcy/dir/sign/v1/public_key_pb';`

sdk/dir-js/test/client.test.ts
```diff
@@ -250,21 +250,23 @@ describe('Client', () => {
...
-          options: {
-            case: 'signature',
-            value: {
-              signature: 'dGVzdC1zaWduYXR1cmU=',
-              annotations: {
-                payload: 'test-payload-data'
-              }
-            },
-          },
+          referrer: signatureReferrer,
```
ramizpolic, line 269: we wont need this, in python/javascript, if we specify the map/object in the structproto field, it works automatically

this is sufficient:

```suggestion
          referrer: {
          	// same as getting the proto message type in golang
            type: models.sign_v1.SignatureSchema.typeName,
            data: {
              signature: 'dGVzdC1zaWduYXR1cmU=',
              annotations: {
                payload: 'test-payload-data',
              },
            }
          },
```

ramizpolic: the same works in python as well. we dont need to have encoding/decoding logic defined in python and ts unlike in golang, because in these languages, it is possible to construct directly from objects instead of encoded/decoded values

sdk/dir-js/src/utils.ts
ramizpolic: this is actually not needed, in both languages we can get the types directly:
- TypeScript: `models.sign_v1.SignatureSchema.typeName => agntcy.dir.sign.v1.Signature`
- Python: `str(sign_v1.Signature.DESCRIPTOR.full_name())`

ramizpolic (approved):
LGTM, great work!

PR #492, opened 2025-10-05
feat: jwt authentication

adamtagscherer:
This PR adds the JWT authentication method besides mTLS. It adds the proper interceptors to the server and the capability to the client to authenticate with JWT.


client/config.go
```diff
@@ -24,6 +24,8 @@ var DefaultConfig = Config{
 type Config struct {
 	ServerAddress    string `json:"server_address,omitempty"     mapstructure:"server_address"`
 	SpiffeSocketPath string `json:"spiffe_socket_path,omitempty" mapstructure:"spiffe_socket_path"`
+	AuthMode         string `json:"auth_mode,omitempty"          mapstructure:"auth_mode"` // "mtls", "jwt", or "" (insecure)
```
arpad-csepi: nit: the options should be in the docs instead here as a comment

client/config.go
```diff
@@ -43,6 +45,14 @@ func LoadConfig() (*Config, error) {
 	_ = v.BindEnv("spiffe_socket_path")
 	v.SetDefault("spiffe_socket_path", "")
 
+	// Authentication mode: "mtls", "jwt", or "" (insecure)
```
arpad-csepi: same as above

client/jwt.go
```diff
@@ -0,0 +1,55 @@
...
+	return map[string]string{
+		"authorization": "Bearer " + jwtSVID.Marshal(),
+	}, nil
+}
+
+// RequireTransportSecurity indicates whether the credentials require transport security.
+func (c *jwtPerRPCCredentials) RequireTransportSecurity() bool {
+	// JWT can work over insecure connections, but TLS is recommended
+	return false
+}
```
arpad-csepi:
```suggestion
```

Not used anywhere

ramizpolic: i think this is used to implement the interface needed for `WithPerRPCCredentials`. the question is: why is the transport security disabled? does this indicate TLS or mTLS exclusion?

client/jwt.go
```diff
@@ -0,0 +1,55 @@
...
+	return &jwtPerRPCCredentials{
+		jwtSource: jwtSource,
+		audience:  audience,
+	}
+}
+
+// WithJWTAuth returns a gRPC dial option that configures JWT authentication.
+func withJWTAuth(_ context.Context, jwtSource *workloadapi.JWTSource, audience string) grpc.DialOption {
+	return grpc.WithPerRPCCredentials(newJWTCredentials(jwtSource, audience))
+}
```
ramizpolic: nit: this is used just once, it is cleaner if we use it directly in the options.go

client/options.go
```diff
@@ -55,28 +57,70 @@ func withAuth(ctx context.Context) Option {
...
+		return fmt.Errorf("failed to create JWT source: %w", err)
 	}
+
+	o.jwtSource = jwtSource
+
+	// Use insecure transport (JWT provides authentication, not transport security)
+	// In production, you may want to use TLS for transport security
+	o.authOpts = append(o.authOpts,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		withJWTAuth(ctx, jwtSource, o.config.JWTAudience),
```
ramizpolic:
```suggestion
		grpc.WithPerRPCCredentials(newJWTCredentials(jwtSource, o.config.JWTAudience),
```

client/options.go
```diff
@@ -55,28 +57,70 @@ func withAuth(ctx context.Context) Option {
...
-		return nil
+		return fmt.Errorf("failed to create JWT source: %w", err)
 	}
+
+	o.jwtSource = jwtSource
+
+	// Use insecure transport (JWT provides authentication, not transport security)
+	// In production, you may want to use TLS for transport security
+	o.authOpts = append(o.authOpts,
+		grpc.WithTransportCredentials(insecure.NewCredentials()),
```
ramizpolic: when using JWT, TLS should be supported instead of insecure

client/options.go
```diff
@@ -21,6 +22,7 @@ type options struct {
 	config     *Config
 	authOpts   []grpc.DialOption
 	authClient *workloadapi.Client
+	jwtSource  *workloadapi.JWTSource
```
ramizpolic: we do not need to store this in client options, nor do we need to close it. because the jwtSource uses the authClient, once we close authClient, it also closes jwtSource. i do not see why we would need to store this since we only need `authOpts`

server/authn/config/config.go
```diff
@@ -0,0 +1,54 @@
...
+
+	// Validate mode
+	if c.Mode != AuthModeJWT && c.Mode != AuthModeMTLS {
+		return fmt.Errorf("invalid auth mode: %s (must be 'jwt' or 'mtls')", c.Mode)
+	}
+
+	// JWT-specific validation
+	if c.Mode == AuthModeJWT && len(c.Audiences) == 0 {
+		return errors.New("at least one audience is required for JWT mode")
+	}
```
ramizpolic: nit: separate validation based on authn type

```suggestion
	if !c.Enabled {
		return nil
	}

	// Switch and validate based on type
	switch c.Mode {
		case AuthModeJWT:
			if len(c.Audiences) == 0 {
				return errors.New("at least one audience is required for JWT mode")
			}
			
			return nil

		case AuthModeMTLS:
			if c.SocketPath == "" {
				return errors.New("socket path is required")
			}
			
			return nil
			
		// unsupported mode
		default:
			return fmt.Errorf("unsupported authn mode: %s", c.Mode)
	}
```

server/authn/jwt.go
```diff
@@ -0,0 +1,158 @@
...
+//
+//nolint:containedctx // Context is required for gRPC stream wrapping
+type wrappedServerStream struct {
+	grpc.ServerStream
+	ctx context.Context
+}
+
+func (w *wrappedServerStream) Context() context.Context {
+	return w.ctx
+}
```
ramizpolic, line 158: do we need to do this wrapping? the socket already injects the data into the request about spire, you can access it via

```
// Extract SPIFFE ID from peer certificate
sid, ok := grpccredentials.PeerIDFromContext(ctx)
if !ok {
	logger.Error("mTLS authentication failed: no peer ID in context")

	return ctx, status.Error(codes.Unauthenticated, "not authenticated via mTLS")
}
```

or is this not the case for JWT?

install/charts/dir/values.yaml
```diff
@@ -16,6 +16,30 @@ apiserver:
     # listen_address: "0.0.0.0:8888"
     # healthcheck_address: "0.0.0.0:8889"
 
+    # Authentication settings (handles identity verification)
+    # Supports both mTLS (X.509-SVID) and JWT (JWT-SVID) authentication
+    authn:
```
ramizpolic, line 21: we also need to update the docker installation config

client/jwt.go
adamtagscherer: First I thought that we should not talk to the server throught TLS when using JWT because it's complicated and there is mTLS for auth + encryption, but turns out it's easy to solve so I updated the code accordingly.

server/authn/jwt.go
adamtagscherer: This is needed for both mTLS and JWT as well, because we store the SPIFFE ID in a custom context key - spiffe-id - and the authz layer then can take this ID from there. Because the `grpc.StreamHandler` expects a `ServerStream` we need this wrapper, in canse of unary streams it wouldn't be needed.

client/client.go
```diff
@@ -56,6 +57,15 @@ func New(opts ...Option) (*Client, error) {
...
 
 func (c *Client) Close() error {
+	// Close JWT source if it exists
+	if c.jwtSource != nil {
+		if err := c.jwtSource.Close(); err != nil {
+			// Log but continue to close authClient
+			_ = err
+		}
+	}
+
```
ramizpolic: this can be removed, right @adamtagscherer? since it uses authClient only

install/docker/apiserver.env
```diff
@@ -1,5 +1,9 @@
 DIRECTORY_SERVER_LISTEN_ADDRESS=0.0.0.0:8888
 DIRECTORY_SERVER_HEALTHCHECK_ADDRESS=0.0.0.0:8889
+DIRECTORY_SERVER_AUTHN_ENABLED=false
```
ramizpolic, line 3: we are missing `authz` part

client/README.md
```diff
@@ -48,25 +48,91 @@ go get github.com/agntcy/dir/client
...
+The SDK can be configured via environment variables or direct instantiation.
 
+### Environment Variables
+
+| Variable | Description | Default |
+|----------|-------------|---------|
+| `DIRECTORY_CLIENT_SERVER_ADDRESS` | Directory server address | `0.0.0.0:8888` |
+| `DIRECTORY_CLIENT_AUTH_MODE` | Authentication mode: `mtls`, `jwt`, or empty for insecure | `""` (insecure) |
+| `DIRECTORY_CLIENT_SPIFFE_SOCKET_PATH` | SPIFFE Workload API socket path | `""` |
+| `DIRECTORY_CLIENT_JWT_AUDIENCE` | JWT audience for JWT authentication | `""` |
```
ramizpolic, line 60: can you just confirm if all these env variables are the same for all client SDKs (go/ts/py)?

client/options.go
```diff
@@ -55,28 +56,79 @@ func withAuth(ctx context.Context) Option {
...
+	}
+
+	// Use TLS for transport security (server presents X.509-SVID)
+	// Client authenticates with JWT-SVID via PerRPCCredentials
+	o.authOpts = append(o.authOpts,
+		grpc.WithTransportCredentials(
+			grpccredentials.TLSClientCredentials(bundleSrc, tlsconfig.AuthorizeAny()),
+		),
+		grpc.WithPerRPCCredentials(newJWTCredentials(jwtSource, o.config.JWTAudience)),
+	)
```
ramizpolic, line 106: 👍

client/README.md
adamtagscherer: The JS and Python Client SDKs have these same env vars + the `DEFAULT_DIRCTL_PATH`.

Issue #466, opened 2025-09-30
[Bug]: Client-side verification validates against payload only


ramizpolic:
### Bug Description

In the method https://github.com/agntcy/dir/blob/39c5e514d2e43ee4fad4f59ef0c1566771d8c8ec/client/verify.go#L49, we validate the signature and public key against the data that was read from the record. A security issue happens when a user:
- Pushes a record
- Pushes malicious public key and signature for that record by providing a valid signature and public key but for some arbitrary message rather than our Record

This happens because there is no check if the signature is signing the CID. All the data that is loaded and verified comes from the payload itself.

### Expected Behavior

Have a check to verify if the signature is signing a given CID

### Affected Version

v0.3.0

### Steps to Reproduce

- Push a record
- Sign with cosign offline a blob with contents `{}`
- Upload obtained signature for Record CID
- Upload obtained public key for Record CID
- Verify client-side

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

PR #493, opened 2025-10-06
fix(dir/client): verify signature against expected payload

paralta:
https://github.com/agntcy/dir/issues/466

Add payload validation in `verifyClientSide()` to ensure signatures are cryptographically bound to the specific record CID.

### DIFF
--- a/client/verify.go
+++ b/client/verify.go
@@ -14,6 +14,7 @@ import (
 	corev1 "github.com/agntcy/dir/api/core/v1"
 	signv1 "github.com/agntcy/dir/api/sign/v1"
 	storev1 "github.com/agntcy/dir/api/store/v1"
+	cosignutils "github.com/agntcy/dir/utils/cosign"
 	ociutils "github.com/agntcy/dir/utils/oci"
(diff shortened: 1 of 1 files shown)


client/verify.go
```diff
@@ -76,6 +88,13 @@ func (c *Client) verifyClientSide(ctx context.Context, recordCID string) (bool,
 			// Get the payload from the signature annotations
 			payload := signature.GetAnnotations()["payload"]
 
+			// Verify that the payload matches the expected payload for this CID
+			if payload != string(expectedPayload) {
+				logger.Debug("Payload mismatch, skipping signature", "expected", string(expectedPayload), "got", payload)
+
+				continue
+			}
```
ramizpolic: we dont need to check the payload at all. we can just verify the signature against the computed `expectedPayload` value

tkircsi (approved):
LGTM.

client/verify.go
paralta: indeed, thanks for reviewing! updated now

ramizpolic (approved):

PR #534, opened 2025-10-13
chore: bump go, golangci-lint and libp2p

paralta:
* Fix https://github.com/agntcy/dir/security/dependabot/34 by upgrading go-libp2p from v0.38.x to v0.44.0
* Bump Go version to 1.25.2 across all modules and Dockerfiles
* Update Docker base images from Debian Bullseye to Bookworm (current stable release with newer packages and better security support)
* Upgrade golangci-lint to v2.5.0 (latest major version)
* Migrate golangci-lint configuration to v2 format and disable non-critical linters (funcorder, godoclint, embeddedstructfieldcheck, noinlineerr, revive, wsl)
* Fix linter warnings across codebase

### DIFF
--- a/.github/workflows/demo-dir.yaml
+++ b/.github/workflows/demo-dir.yaml
@@ -37,7 +37,7 @@ jobs:
       - name: Setup Go
         uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
         with:
-          go-version: '1.24.5'
+          go-version: '1.25.2'
 
       - name: Setup Taskfile
         shell: bash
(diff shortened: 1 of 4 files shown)


hub/auth/internal/browser/open.go
ramizpolic, line 1: Why is this PR updating the HUB codebase?

paralta: the hub code changed because we need to run `FIX=true task lint:go` to fix the linter issues

paralta: we should run linters on the hub code, right?

ramizpolic: seems like the hub linting in this PR is only related to cosmetic inheritance fix for structs/interfaces.

@jubarbot-cisco Could you please review if the HUB changes are okay?

cli/cmd/sync/sync.go
ramizpolic, line 151: can we please remove the `goto`? i would enforce a linter to catch these issues in the future if possible to configure.

https://itnext.io/goto-hell-1e7e32989092

paralta: let's do this outside the scope of the current PR

PR #537, opened 2025-10-13
feat(dir): expose event listener rpc

tkircsi:
The Event Service is a lightweight, real-time event streaming system that enables external clients to monitor Directory system operations via gRPC. It provides an in-memory event bus that captures key operations from all major services and delivers them to interested subscribers with configurable filtering.

### Key Characteristics

- **Real-time**: Events delivered from subscription time forward (no history)
- **Filtered**: Client-side control over event types, labels, and CIDs
- **Type-safe**: Protocol buffer enums for all event types
- **Simple**: In-memory only, no external dependencies
- **Observable**: Built-in metrics and logging
- **Testable**: Mock implementations and test helpers included

### Supported Events (9 Total)

| Service | Events | Implementation |
|---------|--------|----------------|
| **Store** | RECORD_PUSHED, RECORD_PULLED, RECORD_DELETED | Service-level (eventswrap) |
| **Routing** | RECORD_PUBLISHED, RECORD_UNPUBLISHED | Service-level (direct) |
| **Sync** | SYNC_CREATED, SYNC_COMPLETED, SYNC_FAILED | Service-level (worker) |
| **Sign** | RECORD_SIGNED | Controller-level (store controller) |

---

### The Two Patterns

We use **two different patterns** for event emission based on the integration point characteristics:

#### Pattern 1: Service-Level Emission (Wrapper/Direct)

**Used for:** Store, Routing, Sync

**Rationale:**
- **Multiple entry points**: Operations called from multiple places
- **Infrastructure concern**: Core CRUD operations
- **Transparent wrapping**: No business logic needed

#### Pattern 2: Controller-Level Emission

**Used for:** Sign

**Rationale:**
- **Single entry point**: Only via RPC controller
- **Business logic concern**: Type-specific behavior (check if signature)
- **Request inspection needed**: Must examine referrer type

### Configuration Parameter Details

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `subscriber_buffer_size` | int | 100 | Channel buffer size per subscriber. Larger values allow subscribers to fall behind temporarily without dropping events, but use more memory. |
| `log_slow_consumers` | bool | true | Enable logging when events are dropped due to full subscriber buffers. Useful for monitoring slow consumers. |
| `log_published_events` | bool | false | Enable debug logging of all published events. Very verbose - only enable for debugging. |


client/events.go
```diff
@@ -0,0 +1,46 @@
...
+//	    resp, err := stream.Recv()
+//	    if err != nil {
+//	        break
+//	    }
+//	    event := resp.GetEvent()
+//	    fmt.Printf("Event: %s - %s\n", event.Type, event.ResourceId)
+//	}
+func (c *Client) ListenToEvents(ctx context.Context, req *eventsv1.ListenRequest) (eventsv1.EventService_ListenClient, error) {
+	//nolint:wrapcheck
+	return c.Listen(ctx, req)
```
ramizpolic: lets reuse our streaming package here to have a unified stream-processing client usage across methods. this is introducing yet another way to handle streams inside client package. IMO this would be better


```suggestion
func (c *Client) ListenStream(ctx context.Context, req *eventsv1.ListenRequest) (streaming.StreamResult[eventsv1.ListenResponse], error) {
	// handle stream processing
```

server/events/safe.go
ramizpolic, line 1: I think we can remove this and have explicit check for bus creation in the original event bus implementation.
IMO this duplicates code and adds maintance costs with little benefits

server/events/mock_interface_test.go
ramizpolic, line 1: What is the purpose of `testing_test`?

server/events/bus.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+}
+
+// Publish broadcasts an event to all matching subscribers.
+// Events are validated before publishing and delivered only to subscribers
+// whose filters match the event.
+//
+// If a subscriber's channel is full (slow consumer), the event is dropped
+// for that subscriber and a warning is logged (if configured).
+func (b *EventBus) Publish(event *Event) {
```
ramizpolic, line 58: If the `Publish` method is blocking due to any reasons (like bus full, cant deliver msg, slow deliveries, etc), all other methods that are publishing events will also block (ie our whole API). Can you confirm if this is the case?

ramizpolic (changes requested):

client/events.go
tkircsi: Agree. I need to implement ServerStream in our streaming package.

server/events/mock_interface_test.go
tkircsi: This is just testing the Mock implementation functionality. Other tests in the package relies on the Mock bus implementation.

server/events/safe.go
tkircsi: I wouldn't recommend this. Removing nil-safe bus generate 140+ or more nil check in the code as is. The `safe.go` is 138 lines itself. Removing this requires to add nil check in the test also. It follows the  Go's "make the zero value useful" principle.

server/events/bus.go
tkircsi: Nice catch! We need to Publish events asynchronously!

server/events/safe.go
ramizpolic: in that case, i would reconsider if we need certain methods/logic. my concern is that the current logic here is unnecessarily complex. stripping the bus from event builder would help remove that dependency

tkircsi: Ok. I can make EventBuilder independent and remove the builder from the SafeEventBus. They are completely separated. Only the bus/safebus have Publish method. I agree, it is better separation - Builder constructs, Bus publishes. Is that what you meant?

server/events/mock_interface_test.go
ramizpolic: we could rename to `mock_interface_test` so it captures area better

server/events/safe.go
tkircsi: fixed

client/events.go
tkircsi: fixed

server/events/bus.go
tkircsi: fixed

server/events/mock_interface_test.go
tkircsi: fixed

Issue #462, opened 2025-09-29
[Epic] Enable sync of heterogeneous registries


lgecse:
Extend DIR’s current sync capability (limited to DIR-to-DIR) to support synchronization with external registries.  

**Top priority:** integrate with the [MCP registry](https://github.com/modelcontextprotocol/registry).  
**Optional:** consider support for future A2A or other registries.  

**Value for users:**
- Unified view of agents across different registries  
- Easier integration with external ecosystems  
- Reduced duplication of registration and discovery work  

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

PR #544, opened 2025-10-15
feat(dir/importer): mcp registry importer

paralta:
This PR introduces an importer system that enables importing records from external registries into DIR, starting with support for the Model Context Protocol (MCP) registry.

* **MCP Importer Implementation:** HTTP client with streaming support, concurrent worker pool, and converter to transform MCP servers into OASF format
* **Import Command:** New dirctl import CLI command with support for filters, limits, dry-run mode, and progress reporting
* **Testing:** End-to-end tests for import command and unit tests for core components


client/client.go
```diff
@@ -64,3 +64,8 @@ func (c *Client) Close() error {
 
 	return nil
 }
+
+// StoreClient returns the store service client for direct access.
+func (c *Client) StoreClient() storev1.StoreServiceClient {
+	return c.StoreServiceClient
+}
```
tkircsi: Instead of exporting the `StoreServerClient`, I'd pass the whole client and in the client I'd implement the `Push`, `PushStream`, `PushBatch` following the implementation pattern of the `Pull`, `PullStream`, `PullBatch`. After that I'd use `PushStream` in the importer.

paralta: I am fine with using the whole client. Implementing the `PushStream` is a great idea, but it is outside the scope of this PR.

paralta:
@tkircsi thanks for reviewing! The pipeline pattern would indeed provide better separation of concerns and flexibility. It would however add significant complexity.. Let me look into it 👀

paralta:
@tkircsi let me know your thoughts on the latest version 😄

importer/mcp/transformer.go
```diff
@@ -0,0 +1,98 @@
...
+			response.Server.Name, response.Server.Version, err)
+	}
+
+	return record, nil
+}
+
+// convertToOASF converts an MCP server response to OASF format.
+// Note: This is a simplified conversion. Future versions will use OASF-SDK
+// for full schema validation and metadata extraction.
+func (t *Transformer) convertToOASF(response mcpapiv0.ServerResponse) (*corev1.Record, error) {
```
ramizpolic, line 52: nit: we can use native types here to construct a specific record instead of using free-form objects



```suggestion
func (t *Transformer) convertToOASF(response mcpapiv0.ServerResponse) (*corev1.Record, error) {
	record := corev1.New(&typesv1alpha1.Record{
		Name: server.Name,
		// fill record data
	}
	
	// update record data
	
	// return record
	return record, nil
```

importer/mcp/fetcher.go
```diff
@@ -0,0 +1,196 @@
...
+			Timeout: 30 * time.Second, //nolint:mnd
+		},
+		filters: filters,
+		limit:   limit,
+	}
+}
+
+// Fetch retrieves servers from the MCP registry and sends them to the output channel.
+func (f *Fetcher) Fetch(ctx context.Context) (<-chan interface{}, <-chan error) {
+	outputCh := make(chan interface{})
```
tkircsi:
```suggestion
    // Use buffered channel to allow fetcher to work ahead of transformers
	outputCh := make(chan interface{}, 50)
```
A buffered channel has internal storage for N items. Sends only block when the buffer is full. This is a minimal backpressure control. We can define a constant for the buffer size.

importer/mcp/fetcher.go
```diff
@@ -0,0 +1,196 @@
...
+		if !slices.Contains(supportedFilters, key) {
+			close(outputCh)
+
+			errCh <- fmt.Errorf("unsupported filter: %s", key)
+
+			close(errCh)
+
+			return outputCh, errCh
+		}
+	}
```
tkircsi: I'd rather move this validation into the constructor function. It is strange to return closed channels and channel creation in this case is wasteful.

importer/mcp/fetcher.go
```diff
@@ -0,0 +1,196 @@
...
+	return outputCh, errCh
+}
+
+// listServersPage fetches a single page of servers from the MCP registry.
+func (f *Fetcher) listServersPage(ctx context.Context, cursor string) ([]mcpapiv0.ServerResponse, string, error) {
+	// Build URL with query parameters
+	u, err := url.Parse(f.baseURL + "/servers")
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to parse base URL: %w", err)
+	}
```
tkircsi: Move this to the constructor function and store `baseURL` as `*url.URL` instead of `string`. No need to parse it every call.

importer/mcp/fetcher.go
```diff
@@ -0,0 +1,196 @@
...
+	// Create request
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u.String(), nil)
+	if err != nil {
+		return nil, "", fmt.Errorf("failed to create request: %w", err)
+	}
+
+	req.Header.Set("Accept", "application/json")
+
+	// TODO: Implement retry logic for transient failures
+	// Execute request
```
tkircsi, line 162: Consider using github.com/hashicorp/go-retryablehttp

importer/config/config.go
```diff
@@ -4,18 +4,44 @@
...
+
+	// FUTURE:RegistryTypeA2A represents the Agent-to-Agent protocol registry.
+	// RegistryTypeA2A RegistryType = "a2a".
+)
+
+// ClientInterface defines the interface for the DIR client used by importers.
+// This allows for easier testing and mocking.
+type ClientInterface interface {
+	Push(ctx context.Context, record *corev1.Record) (*corev1.RecordRef, error)
+}
```
tkircsi, line 32:
```suggestion
// ClientInterface defines the interface for the DIR client used by importers.
// This allows for easier testing and mocking.
type ClientInterface interface {
    // Streaming push (for bulk operations)
    PushStream(ctx context.Context, recordsCh <-chan *corev1.Record) (streaming.StreamResult[corev1.RecordRef], error)
}
```

importer/pipeline/pipeline.go
```diff
@@ -0,0 +1,268 @@
...
+type Transformer interface {
+	// Transform converts a source record to a target format.
+	Transform(ctx context.Context, source interface{}) (*corev1.Record, error)
+}
+
+// Pusher is an interface for pushing records to the destination (DIR).
+type Pusher interface {
+	// Push sends a record to the destination and returns any error.
+	Push(ctx context.Context, record *corev1.Record) (*corev1.RecordRef, error)
+}
```
tkircsi, line 33:
```suggestion
// Pusher is an interface for pushing records to the destination (DIR).
type Pusher interface {
    // Push receives a channel of records and pushes them to the destination.
    // Returns channels for results, errors, and completion.
    Push(ctx context.Context, inputCh <-chan *corev1.Record) (
        <-chan *corev1.RecordRef, 
        <-chan error,
    )
}```

File: importer/pipeline/pusher.go [hunk 9]
```diff
@@ -0,0 +1,29 @@
...
+	return &ClientPusher{
+		client: client,
+	}
+}
+
+// Push sends a record to DIR using the client.
+func (p *ClientPusher) Push(ctx context.Context, record *corev1.Record) (*corev1.RecordRef, error) {
+	//nolint:wrapcheck
+	return p.client.Push(ctx, record)
+}
```
tkircsi, line 74:
```suggestion
// Push sends records to DIR using the client's streaming API.
func (p *ClientPusher) Push(ctx context.Context, inputCh <-chan *corev1.Record) (
    <-chan *corev1.RecordRef,
    <-chan error,
) {
    refCh := make(chan *corev1.RecordRef)
    errCh := make(chan error)

    go func() {
        defer close(refCh)
        defer close(errCh)

        // Use the streaming API!
        streamResult, err := p.client.PushStream(ctx, inputCh)
        if err != nil {
            errCh <- fmt.Errorf("failed to create push stream: %w", err)
            return
        }

        // Process stream results
        for {
            select {
            case ref, ok := <-streamResult.ResCh():
                if !ok {
                    // Results channel closed
                    continue
                }
                select {
                case refCh <- ref:
                case <-ctx.Done():
                    return
                }

            case err, ok := <-streamResult.ErrCh():
                if !ok {
                    // Error channel closed
                    continue
                }
                select {
                case errCh <- err:
                case <-ctx.Done():
                    return
                }

            case <-streamResult.DoneCh():
                // Stream completed
                return

            case <-ctx.Done():
                return
            }
        }
    }()

    return refCh, errCh
}```

File: importer/pipeline/pipeline.go [hunk 10]
```diff
@@ -0,0 +1,268 @@
...
+	return &Pipeline{
+		fetcher:     fetcher,
+		transformer: transformer,
+		pusher:      pusher,
+		config:      config,
+	}
+}
+
+// Run executes the pipeline.
+func (p *Pipeline) Run(ctx context.Context) (*Result, error) {
```
tkircsi, line 74:
```
func (p *Pipeline) Run(ctx context.Context) (*Result, error) {
    result := &Result{}

    // Stage 1: Fetch records
    fetchedCh, fetchErrCh := p.fetcher.Fetch(ctx)

    // Stage 2: Transform records (still fan-out here)
    transformedCh, transformErrCh := p.runTransformStage(ctx, fetchedCh, result)

    // Stage 3: Push records (uses streaming!)
    var refCh <-chan *corev1.RecordRef
    var pushErrCh <-chan error
    
    if !p.config.DryRun {
        refCh, pushErrCh = p.pusher.Push(ctx, transformedCh)
        
        // Track successful pushes
        go func() {
            for ref := range refCh {
                if ref != nil {
                    result.mu.Lock()
                    result.ImportedCount++
                    result.mu.Unlock()
                }
            }
        }()
    }

    // Collect errors from all stages
    var wg sync.WaitGroup
    wg.Add(3)

    // Collect fetch errors
    go func() {
        defer wg.Done()
        for err := range fetchErrCh {
            if err != nil {
                result.mu.Lock()
                result.Errors = append(result.Errors, fmt.Errorf("fetch error: %w", err))
                result.mu.Unlock()
            }
        }
    }()

    // Collect transform errors
    go func() {
        defer wg.Done()
        for err := range transformErrCh {
            if err != nil {
                result.mu.Lock()
                result.Errors = append(result.Errors, err)
                result.mu.Unlock()
            }
        }
    }()

    // Collect push errors
    go func() {
        defer wg.Done()
        for err := range pushErrCh {
            if err != nil {
                result.mu.Lock()
                result.FailedCount++
                result.Errors = append(result.Errors, err)
                result.mu.Unlock()
            }
        }
    }()

    wg.Wait()

    return result, nil
}
```

importer/pipeline/pipeline.go
```diff
@@ -0,0 +1,268 @@
...
+type Pusher interface {
+	// Push sends a record to the destination and returns any error.
+	Push(ctx context.Context, record *corev1.Record) (*corev1.RecordRef, error)
+}
+
+// Config contains configuration for the pipeline.
+type Config struct {
+	FetcherWorkers     int
+	TransformerWorkers int
+	PusherWorkers      int
```
tkircsi: We can remove it. streaming handles it!

importer/pipeline/pipeline.go
```diff
@@ -0,0 +1,268 @@
...
+		wg.Wait()
+		close(outputCh)
+		close(errCh)
+	}()
+
+	return outputCh, errCh
+}
+
+// runPushStage runs the push stage with concurrent workers.
+func (p *Pipeline) runPushStage(ctx context.Context, inputCh <-chan *corev1.Record, result *Result) <-chan error {
```
tkircsi: We can remove this. New Pusher using PushStream and PushStream uses a single bidirectional gRPC stream to pipeline all records (achieving ~8x lower latency per record and 80% less memory) instead of spawning N workers that each make individual RPC calls with full request/response overhead.

tkircsi (commented):
Some comments on Pusher

importer/pipeline/pusher.go
ramizpolic: I would avoid checking context on each select statement, and only have one select to check context alongside other RW chans

importer/mcp/transformer.go
```diff
@@ -0,0 +1,96 @@
...
+		{
+			Name: "natural_language_processing/analytical_reasoning/problem_solving",
+		},
+	}
+
+	record := &typesv1alpha1.Record{
+		Name:          server.Name,
+		Version:       server.Version,
+		Description:   server.Description,
+		SchemaVersion: "0.7.0",
```
tkircsi: Could we add the SchemaVersion to the Transformer struct as DefaultSchemaVersion instead of a string literal? Or just define it as a constant.

importer/mcp/transformer.go
```diff
@@ -0,0 +1,96 @@
...
+		},
+	}
+
+	// Skills (required, provide default placeholder)
+	// Use a random skill from the list of skills to satisfy the validation.
+	skills := []*typesv1alpha1.Skill{
+		{
+			Name: "natural_language_processing/analytical_reasoning/problem_solving",
+		},
+	}
```
tkircsi, line 87: Same as SchemaVersion

importer/mcp/transformer.go
paralta: this is just a temporary converter until we get a proper one from the OASF-SDK. Let's change this Schema Version but keep the rest as is

cli/cmd/import/import.go
```diff
@@ -0,0 +1,137 @@
...
+`,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		return runImport(cmd)
+	},
+}
+
+func init() {
+	// Initialize factory and register importers once at startup
+	factory = types.NewFactory()
+	mcp.Register(factory)
```
ramizpolic: The importers should be automatically register themselves as part of implementation. I see this logic being done in the following way:

```go
// mcp importer implementation
func init() {
	types.RegisterFactory(mcp)
}

// creating an importer from factory
// method to be implemented, currently: factory.Create()
mcpImporter := types.NewImporter(mcpImporterConfig)
```

cli/cmd/import/import.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Add flags
+	Command.Flags().StringVar(&registryType, "type", "", "Registry type (mcp, a2a)")
```
ramizpolic, line 53: we can still register here the `cfg.RegistryType` without having a dedicated var for it, which we can validate below if the cfg.RegistryType is supported by config and available as importer

cli/cmd/import/import.go
```diff
@@ -0,0 +1,137 @@
...
+		return fmt.Errorf("import failed: %w", err)
+	}
+
+	// Print summary
+	printSummary(cmd, result)
+
+	return nil
+}
+
+func printSummary(cmd *cobra.Command, result *types.ImportResult) {
```
ramizpolic, line 105: i think the import summary should be split in two separate summaries:
- import stats summary - collects useful stats about full imports
- import summary - collects each item summary, e.g. `{item, status}` where we can collect all items based on success. we can do this in a separate issue/PR to be able to easily support retry mechanisms

Both summaries can be handled as JSON, for example

- `dirctl import --type=mcp --url=https://registry.modelcontextprotocol.io` - prints stats summary RAW by default
- `dirctl import --type=mcp --url=https://registry.modelcontextprotocol.io --json` - prints stats as JSON
- `dirctl import --type=mcp --url=https://registry.modelcontextprotocol.io --status status.json` - stores import summary as JSON for each item

importer/config/config.go
```diff
@@ -4,18 +4,45 @@
...
+	RegistryType RegistryType      // Registry type identifier
+	RegistryURL  string            // Base URL of the registry
+	Filters      map[string]string // Registry-specific filters
+	Limit        int               // Number of records to import (default: 0 for all)
+	Concurrency  int               // Number of concurrent workers (default: 5)
+	DryRun       bool              // If true, preview without actually importing
+
+	// Client is the DIR client for pushing records.
+	// This should be provided by the CLI from the already initialized client.
+	Client ClientInterface
```
ramizpolic: IMO we can remove this from config in general and pass it as part of context. With client in config, the config becomes an unserialisable struct and we wont be able to preserve it between runs.
This would require to move the getting and setting client from context from CLI to Client so that we can reuse it.

We could also extend the factory interface to pass client and share it with implementations. I just dont think that config is the right place from which to share the client.

importer/mcp/importer.go
```diff
@@ -0,0 +1,64 @@
...
+	"github.com/agntcy/dir/importer/types"
+)
+
+// Importer implements the Importer interface for MCP registry using a pipeline architecture.
+type Importer struct {
+	registryURL string
+}
+
+// NewImporter creates a new MCP importer instance.
+func NewImporter(cfg config.Config) (types.Importer, error) {
```
ramizpolic: for example, here we can set the factory to receive client as part of constructor rather than passing client as part of config

```
func NewImporter(client, config)
```

importer/mcp/register.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"github.com/agntcy/dir/importer/config"
+	"github.com/agntcy/dir/importer/types"
+)
+
+// Register registers the MCP importer with the factory.
+func Register(factory *types.Factory) {
+	factory.Register(config.RegistryTypeMCP, NewImporter)
+}
```
ramizpolic: we can register our MCP to a reusable/default factory defined in `importer/types/factory.go`

```suggestion
func init() {
	types.RegisterFactory(config.RegistryTypeMCP, NewImporter)
}
```

importer/pipeline/pipeline.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+// Result contains the results of the pipeline execution.
+type Result struct {
+	TotalRecords  int
+	ImportedCount int
+	FailedCount   int
+	Errors        []error
+	mu            sync.Mutex
+}
```
ramizpolic, line 48: Nice, result can also be extended with new fields in the future to enable retries by having something like `FailedIDs`

importer/pipeline/pipeline.go
```diff
@@ -0,0 +1,237 @@
...
+	return &Pipeline{
+		fetcher:     fetcher,
+		transformer: transformer,
+		pusher:      pusher,
+		config:      config,
+	}
+}
+
+// Run executes the pipeline.
+func (p *Pipeline) Run(ctx context.Context) (*Result, error) {
```
ramizpolic, line 74: IMO i do not think we should have `if/else` blocks and control mechanisms in this method depending on if its dry-run or not. I think a better approach is that we have a dedicated `Pipeline` which is used if we want a dry-run mode. This pipeline can have fetcher defined without other methods. This removes the need to control the behaviour based on the usage mode under this method, which makes it easier to extend/maintain in the future.

importer/pipeline/pipeline.go
```diff
@@ -0,0 +1,237 @@
...
+		}()
+	}
+
+	wg.Wait()
+
+	return result, nil
+}
+
+// runTransformStage runs the transformation stage with concurrent workers.
+func (p *Pipeline) runTransformStage(ctx context.Context, inputCh <-chan interface{}, result *Result) (<-chan *corev1.Record, <-chan error) {
```
ramizpolic: i am not sure i see benefit of running transformers concurrently or paralellizing this. could you elaborate what benefits we get with this approach?

importer/pipeline/pusher.go
```diff
@@ -0,0 +1,74 @@
...
+		client: client,
+	}
+}
+
+// Push sends a record to DIR using the client.
+func (p *ClientPusher) Push(ctx context.Context, inputCh <-chan *corev1.Record) (<-chan *corev1.RecordRef, <-chan error) {
+	refCh := make(chan *corev1.RecordRef)
+	errCh := make(chan error)
+
+	go func() {
```
ramizpolic, line 30: the current signature and implementation of this method can cause racing conditions on when to end the listening part (what was processed first, error or result). IMO we can just return the streaming result the way we have it and have the caller of clientpusher handle the errors. 

This is not doing anything more than just abstracting the streaming abstraction. I would consider fully removing the Pusher interface as the exact same behaviour is provided by client itself

cli/cmd/import/import.go
paralta: Let's address this in a separate PR, yes. Regarding the import summary, it would be good to discuss what would be the use case where this is needed before implementing. The sync feature does not provide us this type of report either.

importer/pipeline/pipeline.go
paralta: We already report the errors for the failed items, we can add for example the name+version of the MCP server to that log

importer/pipeline/pipeline.go
paralta: For now the MCP server to OASF record transformation is quite straightforward, but in the future it might require an LLM for skill mapping and other metadata enrichment, where this concurrent transformation will be useful.

importer/pipeline/pusher.go
tkircsi: But how would the clientpusher handle the errors? How does it know how to handle the error?

In my opinion, it is worth keeping the Pusher interface definition. This interface is defined by Pipeline because that is what it expects. It is a separate issue that it is identical or close to the Client interface.

importer/pipeline/pipeline.go
tkircsi: With dry-run we need fetcher and transformer, isn't it?

importer/pipeline/pipeline.go
tkircsi: I agree with @paralta . And even it is a simple method it should be able to accept channels and return channel to combine into the pipeline.

importer/types/importer.go
ramizpolic, line 1: consider renaming to `importer.go`

importer/types/factory.go
ramizpolic, line 1: i would suggest making the factory registration logic more extendible for new importer implementations by exposing all the struct-based interface package wide. 

```go
// registered under `importer/types/factory/factory.go` to avoid cycling imports
// exposed as a package

var importers = map[config.RegistryType]ImporterFunc{}

func Register(registryType config.RegistryType, fn ImporterFunc) {
	importers[registryType] = fn
	// we can also panic here to prevent duplications at compile-time
}

// Create creates a new Importer instance for the given client and configuration.
func Create(client config.ClientInterface, cfg config.Config) (Importer, error) {
	// create from our local importers
}

// other methods
```

this would enable implementations to quickly register

```go
// part of mcp importer implementation
func init() {
	factory.Register(mcp, New)
}
```

cli/cmd/import/import.go
ramizpolic: link https://github.com/agntcy/dir/issues/462

importer/config/config.go
ramizpolic: consider moving this to importer types package rather than config. same for registryType.

importer/types/factory.go
paralta: updated now! thanks for reviewing 😄

importer/pipeline/pusher.go
tkircsi: Errors are handled(collected) in the pipeline `Run` method and results also. There is no race condition, 4 goroutines are running and controlled by a WaitGroup. When all finished, Run returns the *Result.

tkircsi: Pusher's Push streams the inital error if happens and exits or streams the error sent by the server.

importer/config/config.go
ramizpolic: please consider avoiding type declaration in configs, we can do this as part of types

importer/types/factory/factory.go
```diff
@@ -0,0 +1,79 @@
...
+	mu.RLock()
+	defer mu.RUnlock()
+
+	_, exists := importers[registryType]
+
+	return exists
+}
+
+// Reset clears all registered importers. This is primarily useful for testing.
+func Reset() {
```
ramizpolic, line 74: consider making this method private

importer/config/config.go
tkircsi: agree

tkircsi (approved):
Nice work!

PR #547, opened 2025-10-15
ci(security): container security scanning workflow

muscariello:
Adds a container security scanning workflow.

Key features:
- Pulls explicit tagged versions (no dynamic inference) for: dir-apiserver, dir-ctl, zot, spire-server, spire-agent.
- Scans with Trivy (os + library vulns) focusing on severities: CRITICAL,HIGH,MEDIUM (unfixed ignored).
- Publishes SARIF to GitHub code scanning + raw SARIF artifacts per image.
- Provides a markdown job summary enumerating per-image approximate critical/error counts.

Images matrix:
- dir-apiserver: latest
- dir-ctl: latest
- zot: latest
- spire-server: 1.9.6
- spire-agent: 1.9.6


.github/workflows/container-security-scan.yml
```diff
@@ -0,0 +1,117 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - image: dir-apiserver
+            repo: ghcr.io/${{ github.repository_owner }}/dir-apiserver
+            version: latest
+          - image: dir-ctl
+            repo: ghcr.io/${{ github.repository_owner }}/dir-ctl
+            version: latest
```
ramizpolic, line 30: on push and pull request events, these checks will not run on the images produced by that event, but rather whatever latest was pushed. this is because the images are only pushed to `ghcr` after the `CI` pipeline has finished after merging to main (in PRs, we pass it as build artifacts without pushing to ghcr).

If we want to make this work with push and PR events, then we need to wire this into `CI` pipeline properly. For the time being, I would suggest only cron events to trigger this pipeline.

muscariello: No I think we need to have a nightly scan. This is not ready yet. I am triggering the build in PR just to see the output while working on it.

PR #553, opened 2025-10-16
fix(ci): resolve secret inheritance issues in GA

ramizpolic:
## Overview

Fix the issues with secret inheritance in GA pipelines by explicitly passing required secrets.

**NOTE:** `GITHUB_ACTION` is not required for passing as it's available by default in every workflow.


arpad-csepi (approved):

.github/workflows/reusable-release-sdk.yaml
```diff
@@ -28,6 +28,13 @@ on:
         type: boolean
         description: "Make a python SDK release."
         default: false
+    secrets:
+      PYPI_API_TOKEN:
+        description: "PyPI API token for publishing Python SDK"
+        required: false
```
arpad-csepi:
```suggestion
        required: true
```

.github/workflows/reusable-release-sdk.yaml
```diff
@@ -28,6 +28,13 @@ on:
...
+      NPMJS_TOKEN:
+        description: "NPM.js token for publishing JavaScript SDK" 
+        required: false
```
arpad-csepi:
```suggestion
        required: true
```

.github/workflows/reusable-release.yaml
```diff
@@ -14,6 +14,13 @@ on:
         required: true
         type: string
         description: "Release tag for all components."
+    secrets:
+      PYPI_API_TOKEN:
+        description: "PyPI API token for publishing Python SDK"
+        required: false
```
arpad-csepi:
```suggestion
        required: true
```

Because `release` job has the `sdk` job as needs

.github/workflows/reusable-release.yaml
```diff
@@ -14,6 +14,13 @@ on:
...
+      NPMJS_TOKEN:
+        description: "NPM.js token for publishing JavaScript SDK"
+        required: false
```
arpad-csepi:
```suggestion
        required: true
```

Same as above

PR #569, opened 2025-10-20
refactor(sdk): use local generated proto stubs

arpad-csepi:


Taskfile.yml
```diff
@@ -131,21 +131,51 @@ tasks:
       - task: deps:bufbuild
     # NOTE(ramizpolic): This allows Taskfile YAML parsing to accept '{' as a starting command token.
     # In translation, this is interpreted as a regular multi-line shell script.
+    cmds:
+      - '{{.BUFBUILD_BIN}} dep update'
+      - '{{.BUFBUILD_BIN}} generate'
+      - task: api:gen:javascript
+
+  api:gen:javascript:
```
paralta: I would just rename this task here or add this as part of the task above because the `api:gen:javascript` naming is a bit misleading (it only creates a package.json file, it is not generating the stubs)

arpad-csepi: Thanks the comment, I renamed to `api:gen:fix:javascript`, might be a bit better. If you have better name for it, let me know.

paralta: looks good, thanks!

ramizpolic:
The generated models need to be uploaded, just like for Golang. Otherwise, using and releasing Python/TS SDK without generated models will produce broken packages.

PR #576, opened 2025-10-21
feat(dir/install): zot configuration and authentication in helm chart

paralta:
This PR improves Zot configuration and authentication handling in the Helm chart, moving from hostPath volumes to PersistentVolumeClaims and streamlining credential management.

* Replace hostPath volumes with PersistentVolumeClaim for shared Zot configuration storage between apiserver and zot containers
* Embed Zot configuration and htpasswd authentication directly in Helm values with proper secret file support
* Simplify Taskfile deployment by removing deploy:zot:config helper task and generating bcrypt credentials inline using --set-file

### DIFF
--- a/install/charts/dir/values.yaml
+++ b/install/charts/dir/values.yaml
@@ -172,7 +172,7 @@ apiserver:
             },
             "accessControl": {
               "adminPolicy": {
-                "users": ["apiserver"],
+                "users": ["{{ .Values.config.store.oci.auth_config.username }}"],
                 "actions": ["read", "create", "update", "delete"]
               },
               "repositories": {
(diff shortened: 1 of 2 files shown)


install/charts/dir/values.yaml
```diff
@@ -146,35 +146,73 @@ apiserver:
...
+            "address": "0.0.0.0",
+            "port": "5000",
+            "auth": {
+              "htpasswd": {
+                "path": "/secret/htpasswd"
+              }
+            },
+            "accessControl": {
+              "adminPolicy": {
+                "users": ["apiserver"],
```
tkircsi: Just a question, because I'm not sure. What is the case if the `HTPASSWD_USERNAME` is not the default `apiserver` (Taskfile)?

tkircsi (approved):
LGTM. Just one small question.

install/charts/dir/values.yaml
paralta: good point, let me make it configurable here 👍

PR #577, opened 2025-10-22
fix(client): fix Critical Resource Leaks and Context Handling in Client Package

### LINKED ISSUE #575: [Feature]: Fix Critical Resource Leaks and Context Handling in Client Package
### Problem Statement

The client package has **six critical resource management issues** that will cause memory leaks, file descriptor exhaustion, connection leaks, goroutine blocks, and server crashes in production:

1. **gRPC connection never closed** - Every client instance leaks a connection
2. **Context not supported** - Cannot cancel/timeout client initialization
3. **SPIFFE sources never closed** - JWT/x509 sources leak goroutines and connections
4. **Partial cleanup in x509 auth** - x509Src leaked on error path
5. **Blocked goroutine leaks** - Goroutines in routing.go block forever if consumer stops reading
6. **Nil pointer panics** - Missing config validation causes server crashes

Issues 1-3 cause progressive resource exhaustion. **Issue 5 is the most critical** as blocked goroutines are unrecoverable and happen without authentication. Issue 6 can crash the entire server.

### Solution 1: Store and Close gRPC Connection

**File**: `client/client.go`

- Add `conn *grpc.ClientConn` field to `Client` struct
- Store the connection when creating client in `New()`
- Close the connection in `Close()` method
- Collect and return all close errors

### Solution 2: Accept Context in New()

**File**: `client/client.go`

- Change `New()` signature from `New(opts ...Option)` to `New(ctx context.Context, opts ...Option)`
- Pass provided context to `withAuth()` instead of `context.Background()`
- Update all callers to pass context
- Remove `//nolint:contextcheck` suppressions

### Solution 3: Store and Close SPIFFE Sources

**Files**: `client/options.go` and `client/client.go`

- Add fields to `options` struct: `bundleSrc`, `x509Src`, `jwtSource` (all `io.Closer`)
- Store sources in `setupJWTAuth()` and `setupX509Auth()`
- Add corresponding fields to `Client` struct
- Pass sources from options to Client in `New()`
- Close all sources in `Close()` method with proper error collection

### Solution 4: Fix X.509 Auth Error Path

**File**: `client/options.go`

- In `setupX509Auth()`, add `_ = x509Src.Close()` in the error path when `NewBundleSource` fails
- Ensures x509Src is cleaned up on error, matching JWT auth pattern

### Solution 5: Fix Blocked Goroutines

**File**: `client/routing.go`

- In `List()` method: Wrap `resCh <- obj` in `select` with `case <-ctx.Done()` check
- In `SearchRouting()` method: Wrap `resCh <- obj` in `select` with `case <-ctx.Done()` check
- Add error logging when context cancelled
- Matches the correct pattern used in `store.go` `PullReferrer()`

### Solution 6: Add Config Validation

**File**: `client/options.go` or `client/client.go`

- Add `if o.config == nil` check at start of `withAuth()` or in `New()` after loading options
- Return clear error: `"config is required: use WithConfig() or WithEnvConfig()"`
- Prevents nil pointer dereference panics

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

tkircsi:
This PR implements comprehensive resource management fixes for the DIR client library and adds proper context support. It addresses 6 critical issues that could cause resource leaks, goroutine leaks, and panics in production environments.

### Issue 1: gRPC Connection Resource Leak
**Problem:** The gRPC connection created by `grpc.NewClient()` was never stored in the `Client` struct, preventing it from being closed in `Client.Close()`. This caused file descriptor leaks.

**Solution:** Store the gRPC connection in the `Client` struct and close it in `Client.Close()` with proper error aggregation.

**Impact:** Prevents file descriptor exhaustion in long-running applications.

---

### Issue 2: Client Context Support (Breaking Change)
**Problem:** `client.New()` hardcoded `context.Background()` for initialization, preventing callers from controlling timeouts, cancellation, or adding tracing.

**Solution:** Changed `client.New()` signature to accept `context.Context` as the first parameter.

**Breaking Change:** Yes - all callers must now pass a context.

**Impact:** 
- Enables initialization timeout control
- Supports graceful shutdown during startup
- Allows distributed tracing
- Better production observability

---

### Issue 3: SPIFFE Sources Resource Leaks
**Problem:** SPIFFE sources (bundle, X.509, JWT) were created but never stored or closed, leading to goroutine and connection leaks.

**Solution:** Store all SPIFFE sources in the `Client` struct and close them in `Client.Close()` with proper ordering (sources before auth client before connection).

**Impact:** Prevents goroutine leaks from SPIFFE background watchers.

---

### Issue 4: Partial Cleanup Leak in X.509 Auth
**Problem:** In `setupX509Auth()`, if `NewBundleSource` failed after `NewX509Source` succeeded, the X.509 source was not closed on the error path.

**Solution:** Added explicit cleanup of `x509Src` on the error path.

**Impact:** Prevents resource leaks during initialization failures.

---

### Issue 5: Blocked Goroutine Leaks
**Problem:** In `routing.go`, the `List()` and `SearchRouting()` methods spawned goroutines that lacked `select` with `ctx.Done()`. If consumers stopped reading from channels, goroutines would block permanently.

**Solution:** Added `select` statements with `ctx.Done()` checks to all streaming goroutines, matching the pattern used in other methods.

**Impact:** Prevents unrecoverable goroutine leaks. This was the most critical issue.

---

### Issue 6: Nil Pointer Panics
**Problem:** `o.config` was not validated before dereferencing in `withAuth()`, causing server crashes if no config was provided.

**Solution:** Added config validation with clear error messages.

**Impact:** Prevents production crashes from misconfiguration.

Closes #575

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -5,6 +5,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"net"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


paralta (approved):
LGTM 👍

client/client.go
```diff
@@ -40,29 +47,66 @@ func New(opts ...Option) (*Client, error) {
...
+		if err := c.x509Src.Close(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to close X.509 source: %w", err))
+		}
+	}
+
+	if c.bundleSrc != nil {
+		if err := c.bundleSrc.Close(); err != nil {
+			errs = append(errs, fmt.Errorf("failed to close bundle source: %w", err))
+		}
+	}
```
ramizpolic, line 92: we already have `authClient` that we manage. all the other sources (jwt, bundle, x509) are created from the authClient, and closing the authClient closes also all the underlying resources (all the sources). IMO we can remove the other dependency sources from tracking

client/routing.go
```diff
@@ -48,7 +48,14 @@ func (c *Client) List(ctx context.Context, req *routingv1.ListRequest) (<-chan *
...
 			// Stream ListResponse directly (no legacy wrapper)
-			resCh <- obj
+			// Use select to prevent goroutine leak if consumer stops reading
+			select {
+			case resCh <- obj:
+			case <-ctx.Done():
+				logger.Error("context cancelled while receiving list response", "error", ctx.Err())
+
+				return
+			}
```
ramizpolic, line 58: We create our stream with the passed context in the method. Once the context closes, so does the stream. Calling `stream.Recv` on a stream with expired context will return an error and we will exit. 

This means that the context check here is not needed as its already handled by the stream itself. Same for other methods.

client/client.go
tkircsi: @ramizpolic  I don't find evidence in go-spiffe source v2.5.0 that closing the workload.Client closes also the resources created with it. I've just found, it closes the gRPC connection.

```go
// Close closes the client.
func (c *Client) Close() error {
	return c.conn.Close()
}
```
So, closing client makes sources stop working is true, but they may still have goroutines/resources running. I think, explicit cleanup prevents potential resource leaks.

client/routing.go
tkircsi: @ramizpolic  Blocking on `stream.Recv()` handled by stream, but blocking on channel send  not handled by stream. `resCh` is created by the function itself, so the function is the owner of that channel.

PR #591, opened 2025-10-28
test(dir): enable e2e coverage

paralta:
This PR implements code coverage tracking for both unit and end-to-end tests.

- Added coverage-instrumented build target and deployment infrastructure (Dockerfile, docker-bake, Kubernetes templates with `GOCOVERDIR` and coverage volumes)
- Implemented E2E test coverage collection with automated processing tasks to extract, merge, and split coverage profiles from pods
- Updated CI workflows to build coverage images, collect combined unit and E2E coverage data, and upload to Codecov with test-type-specific flags

### DIFF
--- a/.github/workflows/reusable-test-e2e.yaml
+++ b/.github/workflows/reusable-test-e2e.yaml
@@ -33,10 +33,13 @@ jobs:
         include:
           - task: test:e2e:local
             label: "Local"
+            coverage-files: ".coverage/e2e/local-cli.out,.coverage/e2e/local-client.out,.coverage/e2e/server.out"
           - task: test:e2e:network
             label: "Network"
+            coverage-files: ".coverage/e2e/network.out,.coverage/e2e/server.out"
           - task: test:e2e:spire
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable-test-e2e.yaml
ramizpolic, line 38: you can specify the coverage files for each test here by adding a new property (requires to be added for each matrix item):

```suggestion
            label: "Network"
            coverage-files: "file1.json,file2.json"
```

.github/workflows/reusable-build.yaml
ramizpolic, line 1: can you confirm if the releases will not push the coverage images to ghcr.io?

paralta: We only run `coverage-image` when `inputs.push` is false (line 134 `if: ${{ inputs.build_coverage_images && !inputs.push }}`) and it only saves images as local tar artifacts rather than pushing them to any registry.

PR #593, opened 2025-10-28
feat(dir/server): implement Application-Layer Rate Limiting for gRPC

### LINKED ISSUE #565: [Feature]: Implement Application-Layer Rate Limiting for gRPC
### Problem Statement

Currently, the DIR gRPC server has no application-layer rate limiting. This exposes the system to abuse, resource exhaustion, and denial of service attacks.

### Proposed Solution

Implement **two-tier rate limiting** using a hybrid approach:

1. **Application-layer (This Issue)**: Fine-grained, per-client, per-method limits
2. **Infrastructure-layer (Separate Issue)**: Coarse-grained, network-level protection

Use **go-grpc-middleware v2 rate limiting interface** with **custom rate limiter implementation** that supports:

#### 1. Local Rate Limiting
**Technology**: `golang.org/x/time/rate` (Token Bucket)
**Scope**: Single-instance rate limiting
**Use Case**: Development, testing, single-server deployments

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

tkircsi:
This PR implements comprehensive application-layer rate limiting for the DIR gRPC server using the battle-tested `go-grpc-middleware/v2` interceptor framework with a custom limiter implementation that supports per-client and per-method rate limiting.

## Problem Solved

The DIR gRPC server had no rate limiting protection, exposing it to:
- Denial of Service (DoS) attacks
- Resource exhaustion from abusive or buggy clients
- Unfair resource distribution among clients
- Unpredictable costs and capacity issues

### Core Features

1. **Per-Client Rate Limiting**
   - Each authenticated client (identified by SPIFFE ID) gets independent rate limits
   - Prevents one client from monopolizing server resources
   - Fair resource allocation across all clients

2. **Global Rate Limiting**
   - Fallback limits for unauthenticated clients
   - Protects against anonymous abuse

3. **Per-Method Rate Limiting**
   - Optional method-specific overrides for expensive operations
   - Fine-grained control over resource-intensive endpoints
   - Example: Search operations can have stricter limits than simple reads

4. **Token Bucket Algorithm**
   - Smooth rate limiting with burst support
   - Allows temporary traffic spikes above sustained rate
   - Better user experience than strict fixed-rate limiting
  
Closes #565


server/middleware/ratelimit/config/config.go
```diff
@@ -0,0 +1,176 @@
...
+
+	// DefaultPerClientBurst is the default burst capacity for per-client rate limiters.
+	DefaultPerClientBurst = 1500
+)
+
+// Config defines rate limiting configuration for the gRPC server.
+// It supports global rate limiting for unauthenticated clients,
+// per-client rate limiting for authenticated clients (identified by SPIFFE ID),
+// and optional per-method overrides for fine-grained control.
+type Config struct {
```
ramizpolic, line 32: nit: consider using uint types for values that MUST >= 0 and to avoid < 0 checks

server/middleware/ratelimit/server.go
```diff
@@ -0,0 +1,57 @@
...
+
+	return []grpc.ServerOption{
+		grpc.ChainUnaryInterceptor(
+			grpc_ratelimit.UnaryServerInterceptor(limiter),
+		),
+		grpc.ChainStreamInterceptor(
+			grpc_ratelimit.StreamServerInterceptor(limiter),
+		),
+	}, nil
+}
```
ramizpolic, line 57: nit: consider moving as a method on rate limiter (since its already created)


```suggestion
func (l *ClientLimiter) GetServerOptions() []grpc.ServerOption {
	return []grpc.ServerOption{
		grpc.ChainUnaryInterceptor(
			grpc_ratelimit.UnaryServerInterceptor(l),
		),
		grpc.ChainStreamInterceptor(
			grpc_ratelimit.StreamServerInterceptor(l),
		),
	}
}
```

server/middleware/ratelimit/config/config.go
tkircsi: @ramizpolic This is a good idea, but the pros and cons must be considered. In the case of "int", Viper loading is successful at -100, and a user-friendly error message can be returned during validation. In the case of "uint", there is indeed no need for < 0 checking, but there will be a parsing error during Viper loading, where we return an error message that is not entirely user-friendly. WDYT?

tkircsi: Error messages example:
int: Error: invalid rate limit config: global_burst must be non-negative, got: -100
uint: Error: failed to load configuration: json: cannot unmarshal number -100 into Go struct field Config.GlobalBurst of type uint

server/middleware/ratelimit/server.go
tkircsi: Looking at the codebase, there are two different patterns used:
1. Service Pattern (Stateful) - authn, authz
2. Middleware Pattern (Stateless) - recovery, logging, ratelimit

I think, the current approach has better encapsulation also

```go
// server.go doesn't need to know about ClientLimiter
rateLimitOpts, err := ratelimit.ServerOptions(&cfg.RateLimit)
```

with method:
```go
// Exposes implementation detail (ClientLimiter)
limiter, err := ratelimit.NewClientLimiter(&cfg.RateLimit)
serverOpts = append(serverOpts, limiter.GetServerOptions()...)
```

IMO, the `ClientLimiter` is an internal implementation detail. The package should expose a simple `ServerOptions()` function, not force users to understand the internal `ClientLimiter` type.

PR #595, opened 2025-10-29
fix(dir): update hub apis

mtrinell:
Copied over new hub APIs with `RecordDM` renaming. Does not affect anything as only dir `Record` is used in hub subcommand.


jnapper7 (approved):
LGTM

PR #602, opened 2025-10-30
chore(dir): security fixes

paralta:
This PR addresses Dependabot and CodeQL security alerts by moving sensitive credentials to Kubernetes secrets, updating vulnerable dependencies, and removing unused code.

* Helm Charts: Moved OCI and sync authentication credentials from values.yaml to Kubernetes secrets with environment variable injection in deployment templates
* JavaScript SDK: Upgraded Vite from 7.1.5 to 7.1.12 to address CVE-2025-62522
* Python SDK: Removed unused SPIFFE/TLS imports from client.py

### DIFF
--- a/install/charts/dir/apiserver/templates/secret.yaml
+++ b/install/charts/dir/apiserver/templates/secret.yaml
@@ -1,7 +1,7 @@
 # Copyright AGNTCY Contributors (https://github.com/agntcy)
 # SPDX-License-Identifier: Apache-2.0
 
-{{- if or .Values.privKey (and .Values.syncAuthSecret (and .Values.syncAuthSecret.username .Values.syncAuthSecret.password)) (and .Values.ociAuthSecret (and .Values.ociAuthSecret.username .Values.ociAuthSecret.password)) -}}
+{{- if .Values.secrets -}}
 ---
 apiVersion: v1
 kind: Secret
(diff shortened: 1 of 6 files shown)


install/charts/dir/apiserver/templates/secret.yaml
```diff
@@ -1,7 +1,7 @@
 # Copyright AGNTCY Contributors (https://github.com/agntcy)
 # SPDX-License-Identifier: Apache-2.0
 
-{{- if .Values.privKey -}}
+{{- if or .Values.privKey (and .Values.syncAuthSecret (and .Values.syncAuthSecret.username .Values.syncAuthSecret.password)) (and .Values.ociAuthSecret (and .Values.ociAuthSecret.username .Values.ociAuthSecret.password)) -}}
```
ramizpolic: can you please add and document these details in the values file? i would suggest moving to a dedicated struct:

```
secrets:
	<secret-name>: <secret-value>
```

and check with 


```suggestion
{{- if .Values.secrets -}}
```

ramizpolic: also consider generating some random value with helm chart if certain values are not set

paralta: I've implemented the suggested `secrets` refactoring, but had to revert since CodeQL flags any empty password strings in configuration files as security vulnerabilities 😢

PR #612, opened 2025-11-03
chore(cli/hub): add hub sign/verify commands back

jubarbot-cisco:


hub/cmd/sign/sign.go
```diff
@@ -0,0 +1,214 @@
...
+
+		_ = presenter.PrintMessage(cmd, "signature", "Record is", "signed")
+		_ = presenter.PrintMessage(cmd, "signature", "Signature", signature)
+		_ = presenter.PrintMessage(cmd, "signature", "Public Key", publicKey)
+
+		return nil
+	},
+}
+
+func sign(recordCID string) (string, string, error) {
```
jnapper7, line 118: Did you need to add your own sign function?

jubarbot-cisco: Ideally no, but that would have involved a small refactoring on the cli side. We'll do it later.

ramizpolic (approved):

PR #620, opened 2025-11-04
fix(dir/server): fix oci e2e concurrent issues and healthcheck service

tkircsi:
- **Fixed OCI store race condition**: Added retry logic with exponential backoff (3 retries, 50ms-500ms delays) to handle "not found" errors when tagging manifests during concurrent push operations, preventing intermittent failures under high load
- **Fixed rate limit test assertion**: Changed error rate threshold from `< 0.05` to `<= 0.05` to properly allow up to 5% transient errors in concurrent request tests
- **Fixed healthcheck initialization**: Modified `Start()` to immediately set `SERVING` status when no readiness checks are registered, resolving unit test failures caused by 5-second ticker delays
- **Updated CI codecov config**: Uncommented `fail_ci_if_error` setting in GitHub Actions workflow to use default behavior for coverage upload failures

### DIFF
--- a/.github/workflows/reusable-test.yaml
+++ b/.github/workflows/reusable-test.yaml
@@ -63,7 +63,6 @@ jobs:
         with:
           files: .coverage/unit/api.out,.coverage/unit/client.out,.coverage/unit/server.out,.coverage/unit/cli.out,.coverage/unit/hub.out,.coverage/unit/importer.out,.coverage/unit/mcp.out,.coverage/unit/utils.out
           flags: unit
-          # fail_ci_if_error: false
           verbose: true
           token: ${{ secrets.CODECOV_TOKEN }}
 


.github/workflows/reusable-test.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
         with:
           files: .coverage/unit/api.out,.coverage/unit/client.out,.coverage/unit/server.out,.coverage/unit/cli.out,.coverage/unit/hub.out,.coverage/unit/importer.out,.coverage/unit/mcp.out,.coverage/unit/utils.out
           flags: unit
-          fail_ci_if_error: false
+          # fail_ci_if_error: false
```
arpad-csepi:
```suggestion
```

We can remove it as `fail_ci_if_error` default is false
Source: https://github.com/codecov/codecov-action?tab=readme-ov-file#usage

PR #624, opened 2025-11-04
feat(importer): use oasf-sdk/translator

### LINKED ISSUE #572: [Feature]: Leverage OASF-SDK Record Conversion in Importer
### Problem Statement

The current simplified conversion in `importer/mcp/transformer.go` is temporary and needs to be replaced with the proper MCP server translation implementation.

**Critical challenge - Non-deterministic LLM conversion**: When using LLM-based transformation (via the MCP server tools), converting the same MCP server multiple times will produce different OASF records with different content identifiers (CIDs). This means we cannot rely on CID-based deduplication and must implement source-based tracking to prevent duplicate imports of the same MCP server.

### 1. Integrate MCP Server Tools for Conversion

Replace the `convertToOASF` function to use the MCP server tools for proper schema-validated conversion with LLM-based transformation.

### 2. Implement Source-Based Deduplication

Since LLM transformations generate different CIDs for the same source, track MCP servers by source identifier (e.g., `mcp://<registry>/<server-name>@<version>`) instead of CID. Check for existing records before import and provide options to skip or force re-import.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

akijakya:
## Changes

- **Use oasf-sdk translator**: Leverage `oasf-sdk/translator.MCPToRecord()` for converting MCP Registry entries to OASF 0.8.0 records
- **Deduplication**: Skip importing records that already exist (by `name@version`), add `--force` flag to override
- **Debug diagnostics**: Add `--debug` flag to show detailed output including MCP source and generated OASF records for failed imports
- **Graceful error handling**: Server sends `ERROR:` prefixed CIDs for validation/storage failures instead of aborting, allowing subsequent records to continue processing
- **Enricher updates**: Configure enricher to use `dirctl mcp serve` with `agntcy_oasf_get_schema_skills` tool for LLM-based skill selection

## Testing

```bash
# Test basic import with deduplication
./bin/dirctl import --type=mcp --url=https://registry.modelcontextprotocol.io/v0.1 --limit=10

# Test with debug output to see detailed diagnostics
./bin/dirctl import --type=mcp --url=https://registry.modelcontextprotocol.io/v0.1 --limit=5 --debug

# Test force mode to reimport existing records
./bin/dirctl import --type=mcp --url=https://registry.modelcontextprotocol.io/v0.1 --limit=5 --force

# Test with enrichment (requires OpenAI API key)
./bin/dirctl import --type=mcp --url=https://registry.modelcontextprotocol.io/v0.1 --limit=3 --enrich --debug
```

Closes #572

### DIFF
--- a/importer/pipeline/pusher.go
+++ b/importer/pipeline/pusher.go
@@ -9,7 +9,6 @@ import (
 	"errors"
 	"fmt"
 	"os"
-	"strings"
 	"sync"
 
 	corev1 "github.com/agntcy/dir/api/core/v1"
@@ -202,12 +201,7 @@ func (p *ClientPusher) Push(ctx context.Context, inputCh <-chan *corev1.Record) 
 				continue
 			}
 
-			// Check if server returned an ERROR: CID (validation/storage failure)
-			if ref != nil && strings.HasPrefix(ref.GetCid(), "ERROR:") {
-				p.handleServerRejection(ref, record, mcpSourceJSON)
-			}
-
-			// Send reference (success or ERROR: CID)
+			// Send reference (success)
 			select {
 			case refCh <- ref:
 			case <-ctx.Done():
@@ -297,19 +291,6 @@ func (p *ClientPusher) handlePushError(err error, record *corev1.Record, mcpSour
 	select {
 	case errCh <- err:
 	case <-ctx.Done():
-	}
-}
-
-// handleServerRejection handles server rejections (ERROR: CID responses).
-func (p *ClientPusher) handleServerRejection(ref *corev1.RecordRef, record *corev1.Record, mcpSourceJSON string) {
-	logger.Debug("Server rejected record", "cid", ref.GetCid())
-
-	// Print detailed debug output if debug flag is set
-	if p.debug && mcpSourceJSON != "" {
-		// Extract error message from CID
-		errorMsg := strings.TrimPrefix(ref.GetCid(), "ERROR:")
-		errorMsg = strings.TrimSpace(errorMsg)
-		p.printPushFailure(record, mcpSourceJSON, errorMsg)
 	}
 }
 
--- a/server/controller/store.go
+++ b/server/controller/store.go
@@ -65,37 +65,17 @@ func (s storeCtrl) Push(stream storev1.StoreService_PushServer) error {
 			recordName, recordVersion := extractRecordInfo(record)
 
 			// Log validation error with record details
-			storeLogger.Warn("Record validation failed, skipping record",
+			storeLogger.Warn("Record validation failed",
 				"name", recordName,
 				"version", recordVersion,
 				"errors", validationErrors)
 
-			// Send back a RecordRef with error message encoded in Cid field
-			// Format: "ERROR: <validation errors>"
-			// This allows the client to display the actual error without closing the stream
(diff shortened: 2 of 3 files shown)


server/controller/store.go
```diff
@@ -374,3 +403,28 @@ func (s storeCtrl) pullRecordFromStore(ctx context.Context, recordRef *corev1.Re
 
 	return record, nil
 }
+
+// extractRecordInfo extracts name and version from a record for logging.
+func extractRecordInfo(record *corev1.Record) (string, string) {
```
ramizpolic, line 388: we have an adapter interface which can be constructed from the request to fetch natively name and version without reimplementing

server/controller/store.go
```diff
@@ -61,12 +61,41 @@ func (s storeCtrl) Push(stream storev1.StoreService_PushServer) error {
...
+				"version", recordVersion,
+				"errors", validationErrors)
+
+			// Send back a RecordRef with error message encoded in Cid field
+			// Format: "ERROR: <validation errors>"
+			// This allows the client to display the actual error without closing the stream
+			errorMsg := fmt.Sprintf("ERROR: %v", validationErrors)
+			if err := stream.Send(&corev1.RecordRef{Cid: errorMsg}); err != nil {
+				return status.Errorf(codes.Internal, "failed to send error response: %v", err)
+			}
```
ramizpolic: lets not do this, this will cause a lot of issue on the API layer. if the CID is returned, it indicates success. lets explore alternatives here (e.g. by extending the API with PushResponse { cid, error, message } fields

akijakya: Yeah, this is hacky... The problem that needs to be solved, that if a record fails validation, and only an error would be returned as before, it would close the entire stream, making the processing of remaining records in the batch to stop, hanging the import. 

Two solutions comes to mind:
1. Add an error message field to the proto message (essentially creating a PushResponse message instead of the current core.v1.RecordRef), so we can print out why saving the record failed
2. Return an empty CID signaling something went wrong, in this case we cannot print out what the problem was but at least the process doesn't hang

I'd opt for the 1. but that is a bigger, breaking change I'd put in another PR, wdyt?

importer/pipeline/pusher.go
```diff
@@ -5,27 +5,118 @@ package pipeline
...
+			return nil, fmt.Errorf("failed to build existing records cache: %w", err)
+		}
+
+		if debug {
+			fmt.Fprintf(os.Stderr, "[DEDUP] Cache built with %d existing MCP records\n", len(p.existingRecords))
+			os.Stderr.Sync()
+		}
+	}
+
+	return p, nil
```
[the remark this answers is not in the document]
akijakya: This shouldn't be an issue, because we are not expecting parallel import runs, and receiving duplicates from the MCP registry, but implementing a robust system that makes the cache up-to-date during an import would add a lot of overhead, slowing the process down.

server/controller/store.go
akijakya: @tkircsi, @adamtagscherer what is your opinion on this issue?

akijakya: Sorry guys, I realized this is a non-issue now that we push one-by-one, removed the hack and resolving the issue. The PushStream method is still problematic, but that can be solved in another PR.

PR #649, opened 2025-11-10
feat(sdk): add tls grpc auth method

arpad-csepi:


install/charts/dir/apiserver/values.yaml
```diff
@@ -5,13 +5,13 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-nameOverride: ""
-fullnameOverride: ""
+nameOverride: ''
+fullnameOverride: ''
```
ramizpolic: can we please avoid these changes in the future? i think vim may be overriding these automatically

install/charts/dir/apiserver/templates/clusterspiffeids.yaml
```diff
@@ -15,6 +15,8 @@ spec:
     - k8s:sa:{{ include "chart.serviceAccountName" . }}
   spiffeIDTemplate: {{ "spiffe://{{ .TrustDomain }}/ns/{{ .PodMeta.Namespace }}/sa/{{ .PodSpec.ServiceAccountName }}" }}
   autoPopulateDNSNames: true
+  dnsNameTemplates:
+    {{ .Values.spire.dnsNameTemplates }}
```
ramizpolic: can you confirm if this works as well?
```suggestion
    {{- toYaml .Values.spire.dnsNameTemplates | nindent 4 }}
```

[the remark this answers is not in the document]
ramizpolic: @arpad-csepi

arpad-csepi: Working with multiple dns values, thanks!

PR #673, opened 2025-11-12
fix(dir): sign and verify options

mtrinell:
Base options (like `server-address`) to sign and verify subcommands were not actually passed (but overwritten by NewHubOptions).


thelasttoto (approved):
LGTM

PR #686, opened 2025-11-13
fix(brew): use brew updater after release is public

arpad-csepi:


arpad-csepi:
> @arpad-csepi do you know why the CI pipelien doesnt run when github-bot opens a PR? This is a separate issue, but the pipeline never runs and we have to recreate the PR under personal name for CI to run

We are opening the PR with the CI using the `GITHUB_TOKEN` and this behavior is expected to avoid cilcular CI calls.
I modified the brew workflow to only open a PR if the brew test passes. Do we need to run the CI to be able to merge the brew changes? The other tests in the CI are not useful for these hash changes.

Source: https://github.com/marketplace/actions/add-commit#the-commit-from-the-action-is-not-triggering-ci

arpad-csepi:
> > @arpad-csepi do you know why the CI pipelien doesnt run when github-bot opens a PR? This is a separate issue, but the pipeline never runs and we have to recreate the PR under personal name for CI to run
> 
> We are opening the PR with the CI using the `GITHUB_TOKEN` and this behavior is expected to avoid cilcular CI calls. I modified the brew workflow to only open a PR if the brew test passes. Do we need to run the CI to be able to merge the brew changes? The other tests in the CI are not useful for these hash changes.
> 
> Source: https://github.com/marketplace/actions/add-commit#the-commit-from-the-action-is-not-triggering-ci

Managed to find a workaround to use PAT token as GTIHUB_TOKEN for the CI by closing the reopening the brew update PR by hand. In this way the CI is triggered and the we are able to merge the changes as the Status check will be success.
Added a disclaimer to the automated PR description as a guidence. WDYT @ramizpolic , is it acceptable solution? The other solution can be to create an org account and use that account PAT token for commits.

.github/workflows/reusable-brew-update.yaml
```diff
@@ -17,6 +20,10 @@ on:
         type: boolean
         default: false
         description: Run even if there are no changes.
+      run-with-fake-diff:
+        type: boolean
+        default: false
+        description: Run and fake the diff for test the workflow.
```
ramizpolic: what is the difference between `run-without-diff` and `run-with-fake-diff`?

if needed, i suggest replacing both vars with a single `run-force` to force brew updates regardless of the states

arpad-csepi: Those are used for testing, removed as not needed anymore

PR #692, opened 2025-11-17
ci(dir): add workflow dispatch for jobs

arpad-csepi:


.github/workflows/ci.yaml
ramizpolic, line 1: can we only make the release actions callable with the workflow_dispatch instead of the whole CI pipeline?

also, this makes the release run on every CI push, so we would be releasing for each PR merge

arpad-csepi: Removed all workflow_dispatch input expect the release one. The if condition for the release is checking the git ref for tag as well (as before), so we will not releasing with each PR.

.github/workflows/ci.yaml
```diff
@@ -8,8 +8,14 @@ on:
...
     tags:
-      - 'v*.*.*'
+      - "v*.*.*"
   pull_request:
+  workflow_dispatch:
+    inputs:
+      run-release:
+        description: "Run release manually"
+        default: "false"
+        required: false
```
ramizpolic, line 18: can we also specify custom version to release in addition to this? e.g. if we have `run-release, release-version`, then we can create custom releases?

if we run it without specifying the version, the pipeline will fail since `${{ github.ref_name }}` will not resolve to a tag. if it does resolve to a tag, it will resolve to an existing tag and replace it, effectively removing a previous release.

PR #711, opened 2025-11-20
feat(mcp, dir/server): add OASF API validation

### LINKED ISSUE #608: [Feature]: Add `schema_url` env to MCP server conf
### Problem Statement

OASF-SDK 0.0.9 introduced validation via the OASF API when a schema server URL is included in the call, but the MCP server currently cannot be configured with a `schema_url` env.

### Proposed Solution

Add `schema_url` param to the validation tool call of the MCP server if the server is configured with that env var.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

### LINKED ISSUE #609: [Feature]: Add `schema_url` to oasf-sdk validation call in server, api
### Problem Statement

In OASF-SDK 0.0.9 if a `schema_url` is included in the validation call, it uses the OASF API validator, but this is not yet used in the server and api packages.

### Proposed Solution

Add `schema_url` if the server is configured with an OASF server address to the validation call.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

akijakya:
Added OASF API validation by providing a schema URL to the MCP server by configuring SCHEMA_URL env, and set to be the default validation method for the directory server using `https://schema.oasf.outshift.com/`, that can be overwritten with the DIRECTORY_SERVER_OASF_SCHEMA_URL env.

Fixed test records to be valid against API validator and e2e tests accordingly. API validation can be disabled by setting DIRECTORY_SERVER_DISABLE_API_VALIDATION to `true`.

Also updated OASF-SDK to 0.0.14 so the validator now uses context and the API validator can be configured to not use strict validation by setting DIRECTORY_SERVER_STRICT_VALIDATION to `false` (allowing warnings).

These configurations must be made before starting the directory server, and cannot be changed per request. [Here](https://github.com/agntcy/dir/issues/727) is my proposal on how to make this more flexible.

Also added instructions to the readmes on how to try it out with a locally deployed OASF multi-version server instance.

Closes #608 
Closes #609

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -627,9 +627,9 @@ tasks:
           {{ if .DIRECTORY_SERVER_RATELIMIT_PER_CLIENT_RPS }}--set apiserver.config.ratelimit.per_client_rps="{{ .DIRECTORY_SERVER_RATELIMIT_PER_CLIENT_RPS }}"{{ end }} \
           {{ if .DIRECTORY_SERVER_RATELIMIT_PER_CLIENT_BURST }}--set apiserver.config.ratelimit.per_client_burst="{{ .DIRECTORY_SERVER_RATELIMIT_PER_CLIENT_BURST }}"{{ end }} \
           {{ if eq .E2E_COVERAGE_ENABLED "true" }}--set-json 'apiserver.extraEnv=[{"name":"GOCOVERDIR","value":"/tmp/coverage"}]' --set apiserver.coverageVolume=true{{ end }} \
-          {{ if .DIRECTORY_SERVER_SCHEMA_URL }}--set apiserver.config.schema_url="{{ .DIRECTORY_SERVER_SCHEMA_URL }}"{{ end }} \
-          {{ if .DIRECTORY_SERVER_DISABLE_API_VALIDATION }}--set apiserver.config.disable_api_validation="{{ .DIRECTORY_SERVER_DISABLE_API_VALIDATION }}"{{ end }} \
-          {{ if ne .DIRECTORY_SERVER_STRICT_VALIDATION nil }}--set apiserver.config.strict_validation="{{ .DIRECTORY_SERVER_STRICT_VALIDATION }}"{{ end }} \
+          {{ if .DIRECTORY_SERVER_OASF_API_VALIDATION_SCHEMA_URL }}--set apiserver.config.schema_url="{{ .DIRECTORY_SERVER_OASF_API_VALIDATION_SCHEMA_URL }}"{{ end }} \
+          {{ if .DIRECTORY_SERVER_OASF_API_VALIDATION_DISABLE }}--set apiserver.config.disable_api_validation="{{ .DIRECTORY_SERVER_OASF_API_VALIDATION_DISABLE }}"{{ end }} \
+          {{ if ne .DIRECTORY_SERVER_OASF_API_VALIDATION_STRICT_MODE nil }}--set apiserver.config.strict_validation="{{ .DIRECTORY_SERVER_OASF_API_VALIDATION_STRICT_MODE }}"{{ end }} \
           --namespace {{ .HELM_NAMESPACE }} \
           --create-namespace \
           --install \
@@ -1352,7 +1352,7 @@ tasks:
               enabled: true
               trust_domain: {{ .DIR_TRUST_DOMAIN }}
           extraEnv:
-            - name: DIRECTORY_SERVER_DISABLE_API_VALIDATION
+            - name: DIRECTORY_SERVER_OASF_API_VALIDATION_DISABLE
               value: "true"
           spire:
             enabled: true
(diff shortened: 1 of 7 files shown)


mcp/server/server.go
```diff
@@ -18,6 +20,14 @@ import (
 //
 //nolint:maintidx // Function registers all MCP tools and prompts, complexity is acceptable
 func Serve(ctx context.Context) error {
+	// Configure schema URL from environment variable if set
+	schemaURL := os.Getenv("SCHEMA_URL")
```
ramizpolic:
```suggestion
	schemaURL := os.Getenv("OASF_VALIDATOR_SCHEMA_URL")
```

nit: may be better if we use an explicit name and usage for the schema url

server/config/config.go
```diff
@@ -276,6 +285,12 @@ func LoadConfig() (*Config, error) {
 	_ = v.BindEnv("listen_address")
 	v.SetDefault("listen_address", DefaultListenAddress)
 
+	//
+	// OASF Validation configuration
+	//
+	_ = v.BindEnv("schema_url")
+	v.SetDefault("schema_url", DefaultSchemaURL)
```
ramizpolic: I am not sure if we should explicitly enable API-based validation. I think we should let the users do this on their own, also because to disable it, we need to explicitly set SCHEMA_URL to empty, which IMO is an anti-pattern

README.md
```diff
@@ -166,6 +166,61 @@ cd install/docker
 docker compose up -d
 ```
 
+## Server Configuration
+
+The Directory server supports configuration via environment variables, YAML configuration files, or both. Environment variables follow the `DIRECTORY_SERVER_` prefix convention.
+
+### OASF Validation Configuration
+
+The server includes built-in OASF record validation with support for API-based validation:
```
ramizpolic: It may be better to move the documentation around server to DOCS website and to `server` package rather than doing it in the main readme.

Taskfile.yml
```diff
@@ -17,7 +17,7 @@ vars:
   IMAGE_REPO: '{{ .IMAGE_REPO | default "ghcr.io/agntcy" }}'
   IMAGE_TAG: "{{ .IMAGE_TAG | default .COMMIT_SHA }}"
   IMAGE_BAKE_ENV: "IMAGE_REPO={{.IMAGE_REPO}} IMAGE_TAG={{.IMAGE_TAG}}"
-  IMAGE_BAKE_OPTS: '{{ .IMAGE_BAKE_OPTS | default "" }}'
+  IMAGE_BAKE_OPTS: '{{ .IMAGE_BAKE_OPTS | default "--allow=fs.read=.." }}'
```
ramizpolic: could you provide some notes around this change?

Taskfile.yml
```diff
@@ -606,6 +606,11 @@ tasks:
 
         {{ .HELM_BIN }} dependency build {{ .HELM_CHART_PATH }}
 
+        # Build extraEnv JSON array
+        EXTRA_ENV_JSON='[]'
+        {{ if eq .E2E_COVERAGE_ENABLED "true" }}EXTRA_ENV_JSON=$(echo "$EXTRA_ENV_JSON" | jq '. += [{"name":"GOCOVERDIR","value":"/tmp/coverage"}]'){{ end }}
+        {{ if ne .DIRECTORY_SERVER_SCHEMA_URL nil }}EXTRA_ENV_JSON=$(echo "$EXTRA_ENV_JSON" | jq '. += [{"name":"DIRECTORY_SERVER_SCHEMA_URL","value":"{{ .DIRECTORY_SERVER_SCHEMA_URL }}"}]'){{ end }}
```
ramizpolic: could you provide some notes around this change? why is it needed? 

can we not handle it like we do for other configs?

mcp/server/server.go
akijakya: Maybe just OASF_SCHEMA_URL then? We might need it for other purposes than just validation in the future...

server/config/config.go
akijakya: It depends on what we want users to use as default. I'd argue since the API validator is the superior one (can validate content with regex, can return more differentiated results with the use of errors and warnings etc.), it should be the default one. What if instead of setting a value empty for disabling it, we add another env e.g. OASF_API_VALIDATON_DISABLED, which if set to true will disable it?

Taskfile.yml
akijakya: Sorry my mistake, I left it from when I tested it with local oasf-sdk code with package replacement, good catch!

Taskfile.yml
akijakya: Its to be sure DIRECTORY_SERVER_SCHEMA_URL="" overrides the default url and not taken as nil. If we change how to disable API validation, this can be reverted!

mcp/server/server.go
paralta: agree with this one! same for the other variables like for example `DISABLE_API_VALIDATION` to `OASF_VALIDATION_API_DISABLE` and `STRICT_VALIDATION` to `OASF_VALIDATION_STRICT_MODE`

paralta (approved):
LGTM 👍  just this comment here https://github.com/agntcy/dir/pull/711#discussion_r2571647785

mcp/server/server.go
akijakya: I renamed the variables, @paralta check the last commit if you like them or not, thanks!

PR #734, opened 2025-11-26
fix(dir): add cleanup and sudo to spire task

arpad-csepi:
This PR is fixing build and cleanup issues with the spire task and also provide a solution to reduce the evalated permission scope by using sudo only for the kind cloud loadbalancer on MacOS.

Source:
https://github.com/kubernetes-sigs/cloud-provider-kind/blob/506a3b224a53c7dec2608584958714b1efa606b1/cmd/app.go#L80-L83

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -1246,6 +1246,7 @@ tasks:
     desc: Run end-to-end tests for SPIRE deployment
     cmds:
       # Run SPIRE deployment
+      - defer: { task: test:spire:cleanup }
       - task: test:spire
       # Run SDK tests
       - task: sdk:deps:javascript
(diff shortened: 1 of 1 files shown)


tkircsi (approved):

Taskfile.yml
```diff
@@ -719,7 +719,7 @@ tasks:
           --timeout "15m"
 
       - |
-        {{ .HELM_BIN }} upgrade --install spire spire \
+        sudo {{ .HELM_BIN }} upgrade --install spire spire \
```
ramizpolic: not sure why we need sudo here, its a simple helm command

arpad-csepi: Sorry for that, wrong change was commited. I need to test where we need evelated permission, as `task test:spire` needs it to run successfully.

muscariello (approved):
LGTM

Taskfile.yml
```diff
@@ -1246,6 +1246,8 @@ tasks:
     desc: Run end-to-end tests for SPIRE deployment
     cmds:
       # Run SPIRE deployment
+      # - defer: { task: test:spire:cleanup }
```
akijakya: Why is this commented out?

arpad-csepi: By mistake when testing, locally works the modifications to the spire task but CI needs to be ironed out

Taskfile.yml
```diff
@@ -1246,6 +1246,8 @@ tasks:
     desc: Run end-to-end tests for SPIRE deployment
     cmds:
       # Run SPIRE deployment
+      - defer: { task: test:spire:cleanup }
+      - task: build
```
akijakya: During a CI run the `task build` step is already done earlier, locally it can be convenient but I'm not sure it worth it if it adds minutes to the CI job...

arpad-csepi: Good point, removed the task build.

PR #735, opened 2025-11-26
fix(client): add retry logic for X509-SVID availability in SPIRE auth

tkircsi:
- Add exponential backoff retry when fetching X509-SVID during client initialization to handle SPIRE entry sync delays. 
- This fixes authentication failures in CronJobs and short-lived workloads where pods start before SPIRE agents have synced workload entries.
 - Retry up to 10 times with exponential backoff (500ms → 10s) to wait for SPIRE entry sync before establishing gRPC connection.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -3,6 +3,9 @@
 
 version: "3"
 
+env:
+  GOWORK: off
+
 vars:
   ## Version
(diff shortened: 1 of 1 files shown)


arpad-csepi (approved):
LGTM

PR #743, opened 2025-11-27
feat(importer): prioritize dedup checker

### LINKED ISSUE #714: [Feature] Check for duplicates before enrichment in importer
In the current importer implementation, the transformation step including the conversion and enrichment happens before extracting the name and version from the created record to check whether it is a duplicate of an existing record or not. The problem is that the enrichment step takes a long time (invoking and waiting for the result of the MCP server), so it would be better if the duplication check would happen before the enrichment step.

akijakya:
This PR changes the order of steps so the duplication check precedes the transformation/enrichment, saving time for the import process.

Also adds the duplicate checker for the `dry-run` command so the expected summary would be accurate.

Fixes #714

### DIFF
--- a/importer/mcp/importer.go
+++ b/importer/mcp/importer.go
@@ -50,9 +50,19 @@ func (i *Importer) Run(ctx context.Context, cfg config.Config) (*types.ImportRes
 	// Create and run the appropriate pipeline based on dry-run mode
 	var pipelineResult *pipeline.Result
 
+	//nolint:nestif // Complexity is acceptable for pipeline setup with different modes
 	if cfg.DryRun {
-		// Use dry-run pipeline (fetch and transform only)
-		p := pipeline.NewDryRun(fetcher, transformer, pipelineConfig)
+		// Create duplicate checker for accurate dry-run preview (unless --force is set)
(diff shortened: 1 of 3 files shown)


importer/mcp/importer.go
```diff
@@ -55,17 +55,20 @@ func (i *Importer) Run(ctx context.Context, cfg config.Config) (*types.ImportRes
...
+		if !cfg.Force {
+			dupChecker, err = pipeline.NewMCPDuplicateChecker(ctx, i.client, cfg.Debug)
+			if err != nil {
+				return nil, fmt.Errorf("failed to create duplicate checker: %w", err)
+			}
 		}
 
-		p := pipeline.New(fetcher, transformer, pusher, pipelineConfig)
+		// Create full pipeline with optional duplicate checker
+		p := pipeline.New(fetcher, transformer, pusher, dupChecker, pipelineConfig)
```
paralta:
```suggestion
		p := pipeline.New(fetcher, duplicateChecker, transformer, pusher, pipelineConfig)
```

it would be neat if the constructor parameter order matched the actual pipeline execution flow

PR #746, opened 2025-11-28
feat(dir): add job to fetch latest release version for container security scan workflow

paralta:
This PR adds a dynamic version fetching job to the container security scan workflow that automatically retrieves the latest release version from the GitHub API instead of using hardcoded versions. This ensures the security scans always run against the most recent release without requiring manual updates to the workflow file when new versions are released.


arpad-csepi (approved):
LGTM :shipit:

PR #747, opened 2025-12-01
refactor(dir): local search

adamtagscherer:
## Local Search Enhancements

- Added support for `created_at`, `authors`, `schema_version`, and `module_id`
- Refactored `dirctl search` into subcommands:
  - `dirctl search cids` - returns only CIDs (efficient for piping)
  - `dirctl search records` - returns full record data
- Multiple values per filter field now use OR logic
- Refactored search methods to use `streaming.StreamResult` pattern

### DIFF
--- a/e2e/local/02_search_test.go
+++ b/e2e/local/02_search_test.go
@@ -196,6 +196,79 @@ var _ = ginkgo.Describe("Search functionality for OASF 0.8.0 records", func() {
 				gomega.Expect(output).To(gomega.ContainSubstring(recordCID))
 			})
 		})
+
+		ginkgo.Context("comparison operators", func() {
+			ginkgo.It("finds record with version >= v3.0.0", func() {
+				output := cli.Search().
+					WithVersion(">=v3.0.0").
+					ShouldSucceed()
+				gomega.Expect(output).To(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("finds record with version <= v5.0.0", func() {
+				output := cli.Search().
+					WithVersion("<=v5.0.0").
+					ShouldSucceed()
+				gomega.Expect(output).To(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("finds record with version > v3.0.0", func() {
+				output := cli.Search().
+					WithVersion(">v3.0.0").
+					ShouldSucceed()
+				gomega.Expect(output).To(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("does not find record with version < v4.0.0", func() {
+				output := cli.Search().
+					WithVersion("<v4.0.0").
+					ShouldSucceed()
+				gomega.Expect(output).NotTo(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("finds record with version =v4.0.0", func() {
+				output := cli.Search().
+					WithVersion("=v4.0.0").
+					ShouldSucceed()
+				gomega.Expect(output).To(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("finds record with schema-version >= 0.7.0", func() {
+				output := cli.Search().
+					WithSchemaVersion(">=0.7.0").
+					ShouldSucceed()
+				gomega.Expect(output).To(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("does not find record with schema-version > 0.8.0", func() {
+				output := cli.Search().
+					WithSchemaVersion(">0.8.0").
+					ShouldSucceed()
+				gomega.Expect(output).NotTo(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("finds record with created-at >= 2025-01-01", func() {
+				output := cli.Search().
+					WithCreatedAt(">=2025-01-01").
+					ShouldSucceed()
+				gomega.Expect(output).To(gomega.ContainSubstring(recordCID))
+			})
+
+			ginkgo.It("does not find record with created-at < 2025-01-01", func() {
+				output := cli.Search().
+					WithCreatedAt("<2025-01-01").
+					ShouldSucceed()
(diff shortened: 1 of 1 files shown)


client/search.go
```diff
@@ -9,13 +9,16 @@ import (
...
 
+	corev1 "github.com/agntcy/dir/api/core/v1"
 	searchv1 "github.com/agntcy/dir/api/search/v1"
 )
 
-func (c *Client) Search(ctx context.Context, req *searchv1.SearchRequest) (<-chan string, error) {
-	stream, err := c.SearchServiceClient.Search(ctx, req)
+// SearchCIDs searches for record CIDs matching the given request.
+// Returns a channel that streams CIDs as they are received.
+func (c *Client) SearchCIDs(ctx context.Context, req *searchv1.SearchRequest) (<-chan string, error) {
```
tkircsi: Please, refacto client to use StreamResult pattern like other streaming methods ie: events/listen

```go
func (c *Client) SearchCIDs(ctx context.Context, req *searchv1.SearchRequest) (streaming.StreamResult[string], error) {
    stream, err := c.SearchServiceClient.SearchCIDs(ctx, req)
    if err != nil {
        return nil, fmt.Errorf("failed to create search CIDs stream: %w", err)
    }

    // Use ProcessServerStream helper
    return streaming.ProcessServerStreamString(ctx, stream)
}
```

cli/cmd/search/cids.go
```diff
@@ -0,0 +1,188 @@
...
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		return runCidsCommand(cmd)
+	},
+}
+
+func init() {
+	registerFlags(cidsCmd)
+}
+
+func runCidsCommand(cmd *cobra.Command) error {
```
tkircsi, line 49: Update it to handle errors properly and use the proper pattern

```go
func runCidsCommand(cmd *cobra.Command) error {
    // ... setup code ...
    
    result, err := c.SearchCIDs(cmd.Context(), &searchv1.SearchRequest{
        Limit:   &opts.Limit,
        Offset:  &opts.Offset,
        Queries: queries,
    })
    if err != nil {
        return fmt.Errorf("failed to search: %w", err)
    }

    results := make([]interface{}, 0)

    for {
        select {
        case cid := <-result.ResCh():
            if cid != "" {
                results = append(results, cid)
            }
        case err := <-result.ErrCh():
            return fmt.Errorf("error receiving CID: %w", err)
        case <-result.DoneCh():
            return presenter.PrintMessage(cmd, "record CIDs", "Record CIDs found", results)
        case <-cmd.Context().Done():
            return cmd.Context().Err()
        }
    }
}
```

tkircsi (commented):
Please use the existing streaming patterns.

client/routing.go
```diff
@@ -63,7 +63,7 @@ func (c *Client) List(ctx context.Context, req *routingv1.ListRequest) (<-chan *
 }
 
 func (c *Client) SearchRouting(ctx context.Context, req *routingv1.SearchRequest) (<-chan *routingv1.SearchResponse, error) {
-	stream, err := c.RoutingServiceClient.Search(ctx, req)
+	stream, err := c.Search(ctx, req)
```
tkircsi, line 66: Are you sure, this is correct? `SearchRouting` is about the remote search in the local BadgerDB.

adamtagscherer: No, you are absolutely right.

adamtagscherer: Actually I get the following linter error if I don't remove the `RoutingServiceClient` from the call. If we call `c.Search` it will correctly call `RoutingServiceClient`'s `Search` method because `RoutingServiceClient` is embedded into `Client`. IDK why it's not failing on `main` already.

```bash
Running golangci-lint in ./client
client/routing.go:66:19: QF1008: could remove embedded field "RoutingServiceClient" from selector (staticcheck)
	stream, err := c.RoutingServiceClient.Search(ctx, req)
	                 ^
1 issues:
* staticcheck: 1
task: Failed to run task "lint:go": exit status 1
```

tkircsi: Ah, you're right. This is because there is no longer any conflict between the interfaces of the embedded components. Local Search no longer has a Search method that conflicts with Routing Search.

server/controller/search.go
```diff
@@ -45,7 +48,43 @@ func (c *searchCtlr) Search(req *searchv1.SearchRequest, srv searchv1.SearchServ
...
+		if err != nil {
+			searchLogger.Warn("Failed to pull record from store", "cid", cid, "error", err)
+
+			continue
+		}
+
+		if err := srv.Send(&searchv1.SearchRecordsResponse{Record: record}); err != nil {
 			return fmt.Errorf("failed to send record: %w", err)
 		}
 	}
```
tkircsi:
```suggestion
// Pull full records from store and stream them
for _, cid := range recordCIDs {
	// Check if client disconnected before expensive store.Pull
	if err := srv.Context().Err(); err != nil {
		return err  // or return nil if you want graceful termination
	}

	record, err := c.store.Pull(srv.Context(), &corev1.RecordRef{Cid: cid})
	if err != nil {
		searchLogger.Warn("Failed to pull record from store", "cid", cid, "error", err)
		continue
	}

	if err := srv.Send(&searchv1.SearchRecordsResponse{Record: record}); err != nil {
		return fmt.Errorf("failed to send record: %w", err)
	}
}
```

cli/cmd/search/cids.go
```diff
@@ -0,0 +1,193 @@
...
+		case <-result.DoneCh():
+			return presenter.PrintMessage(cmd, "record CIDs", "Record CIDs found", results)
+		case <-cmd.Context().Done():
+			return cmd.Context().Err()
+		}
+	}
+}
+
+// buildQueriesFromFlags builds API queries.
+func buildQueriesFromFlags() []*searchv1.RecordQuery {
```
tkircsi: Move `buildQueriesFromFlags()` to options.go since it operates on opts and used by both the `cids.go` and `records.go`. This makes the relationship clear: `options.go` owns both the flag definitions and the query builder.

PR #749, opened 2025-12-01
ci(dir): avoid no space left in runner

arpad-csepi:


.github/workflows/clear-docker-cache.yaml
```diff
@@ -0,0 +1,15 @@
...
+on:
+  workflow_dispatch:
+jobs:
+  clear-docker-cache:
+    name: Clear Docker Cache
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clear Github Action runner cache
+        run: |
+          gh cache delete --all --succeed-on-no-caches --repo "${{ github.repository }}"
```
tkircsi, line 15:
```suggestion
jobs:
  clear-docker-cache:
    name: Clear Docker Cache
    runs-on: ubuntu-latest
    steps:
      - name: Clear Github Action runner cache
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          echo "Listing current caches..."
          gh cache list --repo "${{ github.repository }}" || true
          echo "Deleting all caches..."
          gh cache delete --all --repo "${{ github.repository }}" || echo "No caches to delete"
```

.github/workflows/reusable-test-e2e.yaml
```diff
@@ -112,6 +96,12 @@ jobs:
           done
           docker images
 
+      - name: Load compiled dirctl binary
+        run: |
+          mkdir ./bin
+          cp tmp/artifacts/dirctl ./bin/dirctl
+          chmod +x ./bin/dirctl
+
```
tkircsi, line 104: Can we download `dirctl` directly into the final destination?

```
- name: Download CLI artifact
  uses: actions/download-artifact@634f93cb2916e3fdff6788551b99b062d0335ce0 # v5.0.0
  with:
    name: dirctl-bin
    path: ./bin/

- name: Make binary executable
  run: chmod +x ./bin/dirctl
```

tkircsi: the same in `reusable-test-sdk.yaml`

.github/workflows/reusable-build.yaml
```diff
@@ -31,20 +31,56 @@ on:
...
+        run: |
+          task cli:compile
+          mv ./bin/dirctl /tmp/artifacts/dirctl
+
+      - name: Upload CLI artifact
+        uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        with:
+          name: dirctl-bin
+          path: /tmp/artifacts/dirctl
+          retention-days: 1
```
tkircsi: Is 1 day sufficient? Consider re-running failed jobs might need artifacts or debug workflows might reference older runs. WDYT about 3-7 days?

.github/workflows/clear-docker-cache.yaml
arpad-csepi: Do you have a reason why we want to list the caches before deleting them? I do not see value in it as we cannot restore it and named semi-randomly.

The delete sub(sub?) command covers basic information about how many cache is deleted and prints out to the stdout. The `--succeed-on-no-caches` prevents the error on exit when there is no cache, so we do not need `||` to handle errors.

Source: https://github.com/cli/cli/blob/c8ab18323ee47721cba857bf82d55a21143566cb/pkg/cmd/cache/delete/delete.go#L202

.github/workflows/reusable-test-e2e.yaml
arpad-csepi: We can download `dirctl` directly to the final destination but right now, the `Download artifacts` step is downloading all artifacts like docker images, binaries, etc. in only one step.
I do not see much difference between these 2 case:
- Download all artifacts and move it to the right place (1 download, 1 move+chmod step)
- Download specific artifacts to the right place (2 download step, 1 chmod step)

So I prefer to keep in this way as a but more simpler for the artifact download step and 1 step shorter the solution.

.github/workflows/reusable-build.yaml
arpad-csepi: The artifacts are used for only 1 CI run as a way to transport data between workflows. Those are not used as a traditional cache. So the `dirctl` binary will be built every time when the CI is triggered, just now will be only one compilation instead of multiple spread across multiple workflows.

The Github Action cache is used by the `setup-go`, so the binary compile time is short due to the packages for it is already will be loaded and does not need to download each time.

So 1 day is sufficient for the artifacts IMO, as we need only for 1 specific CI run and not reused across multiple CI runs (just the cache). If we want we can extend the retention days to a week. This enables us to rerun a specific failed CI job which relies on the artifact, but this is applies for only the `Rerun only failed job` case.

.github/workflows/reusable-test-e2e.yaml
tkircsi: There's a minor consideration. Downloading all artifacts uses more disk space temporarily. Given the PR is about "no space left" issues, downloading only what's needed per workflow might be slightly better. But I'm fine with your approach too.

.github/workflows/reusable-build.yaml
tkircsi: What happens if I re-run the failed jobs after 3 days the CI was failed? I think, the workflow will fail because `dirctl-bin` artifact no longer exists. WDYT?

.github/workflows/clear-docker-cache.yaml
tkircsi: Our existing gh CLI usage in `reusable-brew-update.yaml` explicitly sets `GH_TOKEN` at the job level with a helpful comment. Since this is our only other workflow using gh CLI, the new cache deletion workflow should follow the same pattern for consistency. 
I'm fine with your `--succeed-on-no-caches` approach. The `list` has only some diagnostic value like: what caches exist (keys, sizes, branches), show workflow/branch/dependency info... (optional)

.github/workflows/reusable-build.yaml
arpad-csepi: That's correct, this is what I describe above more or less. In this case we can restart the whole CI to regenerate the artifacts, but I will extend the retention to 7 days so we can re-run the failed jobs within a week.

.github/workflows/clear-docker-cache.yaml
arpad-csepi: The `reusable-brew-update.yaml` sets `GH_TOKEN` explicitly for a reason. That workflow needs to be able to create a PR under a different user/bot account, so the CI will check the PR and anybody can approve and merge it.
I rather not follow the same pattern there as the default value of `GH_TOKEN` in the most cases are good for us.

.github/workflows/reusable-test-e2e.yaml
arpad-csepi: You are right however all the artifacts are used by all the affected workflows right now, so we do not gain disk space with downloading separatly. Later we can change that and group artifacts more effectively to avoid downloading every single stuff one by one.

.github/workflows/reusable-build.yaml
arpad-csepi: Done

PR #759, opened 2025-12-04
refactor(dir): remove search subcommands

adamtagscherer:
## Summary

Refactors `dirctl search` from subcommand-based (`search cids` / `search records`) to flag-based approach (`--format cid` / `--format model`) to maintain backwards compatibility and avoid breaking changes in dependent projects.

## Changes

- **CLI**: Merged `cids` and `records` subcommands into single `search` command with `--format` flag
  - `--format cid` (default) - returns only CIDs
  - `--format model` - returns full record data

## Usage

```bash
dirctl search --name "web*"

dirctl search --name "web*" --format model --output json

dirctl search --version ">=1.0.0" --created-at ">=2024-01-01"
```

### DIFF
--- a/cli/README.md
+++ b/cli/README.md
@@ -268,7 +268,7 @@ Search for records in the directory. Use `--format` to control output type.
 
 **Format options:**
 - `--format cid` (default) - Return only record CIDs (efficient for piping)
-- `--format model` - Return full record data
+- `--format record` - Return full record data
 
 **Examples:**
 ```bash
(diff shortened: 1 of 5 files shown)


paralta (approved):
LGTM 👍 though the format naming `model` might be misleading since we do not use this term anywhere else. Maybe just keep the prev subcommand name `record`, WDYT?

tkircsi (approved):

PR #761, opened 2025-12-05
chore(deps): update zot and fix codeql workflow warning

paralta:
This PR updates the Zot registry dependency and fixes a CodeQL workflow configuration issue.

**Changes:**
* Fix CodeQL workflow by moving the Initialize CodeQL step after Go setup to prevent tool interference warnings
* Update Zot across all configurations (Taskfile, go, helm, docker-compose, container security scan)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -31,6 +31,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Setup Go
+        uses: actions/setup-go@44694675825211faa026b3c33043df3e48a5fa00 # v6.0.0
+        with:
+          go-version: "1.25.2"
+
       - name: Setup Taskfile
         shell: bash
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -190,6 +162,32 @@ jobs:
 
         echo "Go environment prepared for autobuild"
 
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v4 # v4.0.0
```
arpad-csepi:
```suggestion
      uses: github/codeql-action/init@cf1bb45a277cb3c205638b2cd5c984db1c46a412 # v4.31.7
```
Just a nit to pin the version. This way is more align with the rest of the workflows, not braking hopefully seemingly randomly because of an update. Or if we would like to keep the `v4` tag, then please remove the comment next to it as not correct.

PR #785, opened 2025-12-16
feat(dir/server): validation configuration

### LINKED ISSUE #784: Default schema URL in the config instead of the code
Currently the default OASF API endpoint is hardcoded in the code. Instead, it should be made default through configuration (Helm chart, mcphost.json).

akijakya:
This PR changes how the default OASF schema url is handled: it must come from the config instead of being hardcoded in the code. If a schema URL is not provided, and api validation is not disabled from the configuration, an error is returned about setting providing an OASF URL.

Closes #784

### DIFF
--- a/sdk/examples/example-js/example.js
+++ b/sdk/examples/example-js/example.js
@@ -1,15 +1,16 @@
 // Copyright AGNTCY Contributors (https://github.com/agntcy)
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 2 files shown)


paralta (commented):
Validation config and test changes look good to me 👍 However, I believe the Taskfile changes lead to overly complex e2e test tasks with significant code duplication. I would suggest trying to create a new task that takes in the validation mode as a variable

PR #793, opened 2026-01-06
feat(cli): dirctl validate

### LINKED ISSUE #767: [Feature]: `dirctl validate` for client-side record validation
### Problem Statement

There is no way other than using the MCP server to validate a record locally before pushing.

It would also be valuable if the whole directory could be re-validated with a `dirctl validate` command, ensuring the node's content is compatible with 
- another OASF instance
- the JSON schema validator
- the currently used OASF instance that added new validation rules/fixed existing ones

### Proposed Solution

Implement a `validate` command with flags 
- to provide a schema url
- disable strict API validation
- disable API validation so the schema validator would be used
- specify a server instance for a content re-validation
- pass a record json that would be validated locally

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

akijakya:
This PR adds `validate` command to dirctl to either validate a local json file, or the whole content of the directory  with different settings. Latter is done in batches, a test I ran that validated 3214 records finished in 10:17 mins.

```
dirctl validate --help
Validate OASF record JSON against the OASF schema. The JSON can be provided
as a file path or piped from stdin (e.g., from dirctl pull).

You must specify either --url for API-based validation or --disable-api for
embedded schema validation.

Usage examples:

1. Validate a file using embedded schemas (no API calls):
   dirctl validate record.json --disable-api

2. Validate a file with API-based validation:
   dirctl validate record.json --url https://schema.oasf.outshift.com

3. Validate a file with non-strict mode (more permissive, only works with --url):
   dirctl validate record.json --url https://schema.oasf.outshift.com --disable-strict

4. Validate JSON piped from stdin:
   cat record.json | dirctl validate --disable-api

5. Validate a record pulled from directory:
   dirctl pull <cid> | dirctl validate --disable-api

Note: You must specify either --url (for API validation) or --disable-api
(for embedded schema validation).

Usage:
  dirctl validate [<file>] [flags]

Flags:
      --disable-api      Disable API-based validation (use embedded schemas instead, required if --url is not specified)
      --disable-strict   Disable strict validation mode (more permissive validation, only works with --url)
  -h, --help             help for validate
      --url string       OASF schema URL for API-based validation (required if --disable-api is not specified)
```

Closes #767

### DIFF
--- a/cli/cmd/validate/options.go
+++ b/cli/cmd/validate/options.go
@@ -3,17 +3,12 @@
 
 package validate
 
-import (
-	"github.com/agntcy/dir/cli/presenter"
-)
-
 var opts = &options{}
 
 type options struct {
 	DisableAPI    bool
 	DisableStrict bool
 	SchemaURL     string
-	ValidateAll   bool
 }
 
 func init() {
@@ -25,9 +20,4 @@ func init() {
 		"Disable strict validation mode (more permissive validation, only works with --url)")
 	flags.StringVar(&opts.SchemaURL, "url", "",
 		"OASF schema URL for API-based validation (required if --disable-api is not specified)")
-	flags.BoolVar(&opts.ValidateAll, "all", false,
-		"Validate all records in the directory instance (requires connection to Directory server)")
-
-	// Add output format flags
-	presenter.AddOutputFlags(Command)
 }
--- a/cli/README.md
+++ b/cli/README.md
@@ -343,63 +343,42 @@ Verify record signatures.
 dirctl verify record.json signature.sig --key public.key
 ```
 
-#### `dirctl validate [<file>|--all] [flags]`
-Validate OASF record JSON files locally or all records in a directory instance.
+#### `dirctl validate [<file>] [flags]`
+Validate OASF record JSON from a file or stdin.
 
-This command can validate a single JSON file locally (without requiring a connection
-to a Directory server) or validate all records in a directory instance (requires
-connection to Directory server).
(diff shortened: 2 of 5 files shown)


ramizpolic:
We should reconsider flag usage, for example:
- default case, no flags -> runs embedded schema validation
- `--api` -> runs api-based validation
- `--strict` -> runs api-based validation in strict mode

i think following default + enable approach is more intuitive than enforcing certain checks which need to be explicitly disabled

cli/cmd/validate/options.go
```diff
@@ -0,0 +1,23 @@
...
+
+func init() {
+	flags := Command.Flags()
+
+	flags.BoolVar(&opts.DisableAPI, "disable-api", false,
+		"Disable API-based validation (use embedded schemas instead, required if --url is not specified)")
+	flags.BoolVar(&opts.DisableStrict, "disable-strict", false,
+		"Disable strict validation mode (more permissive validation, only works with --url)")
+	flags.StringVar(&opts.SchemaURL, "url", "",
+		"OASF schema URL for API-based validation (required if --disable-api is not specified)")
```
ramizpolic, line 22: imo we do not need to reflect the config options set against the backend since this is our CLI approach. i still think that the flags are a bit unintuitive to work with and explain due to disable-first approach.

```bash
dirctl validate # api validation
dirctl validate --url <my-url> # api validation
dirctl validate --strict # api validation with strict mode
dirctl validate --local # static, json schema validation
```

PR #800, opened 2026-01-09
feat(release): migrate js sdk to npm trusted publishers and decouple sdk releases from ci

paralta:
Updates the SDK release process to use [npm Trusted Publishers](https://docs.npmjs.com/trusted-publishers) for OIDC-based authentication and makes SDK releases a manual step separate from the main release workflow.

- Replace `NODE_AUTH_TOKEN` with OIDC trusted publishing for npm (adds `id-token: write` permission)
- Remove SDK job from automated release workflow; now triggered manually via `workflow_dispatch`
- Remove `NPMJS_TOKEN` and `PYPI_API_TOKEN` secrets from CI workflows
- Update `RELEASE.md` with manual SDK release instructions
- Clean up `Taskfile.yml` and `.npmrc` to remove token references

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -141,8 +141,6 @@ jobs:
     with:
       image_repo: ghcr.io/agntcy
       release_tag: ${{ github.ref_name }}
-    secrets:
-      PYPI_API_TOKEN: ${{ secrets.PYPI_API_TOKEN }}
 
   integration:
     name: Run integration tests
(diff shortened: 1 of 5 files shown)


paralta:
@ramizpolic The SDK release workflow was moved to manual execution because version conflicts (when a version has already been pushed) cause the entire CI pipeline to fail and block. Once the SDK release process is more reliable at handling these cases, we can re-integrate it into the automated CI workflow.

PR #803, opened 2026-01-12
feat(dir): domain ownership verification

adamtagscherer:
This PR adds domain ownership verification for OASF records. Publishers can prove they control the domain in their record's name field by publishing their signing public key via DNS TXT record or well-known file.

**Features Implemented**
**Name verification via JWKS** - Uses standard RFC 7517 /.well-known/jwks.json
**Name verification via DNS TXT** - Uses _oasf.<domain> TXT records
**Protocol prefixes** - dns:// or http:// or https:// to specify verification method
**Automatic verification on sign** - Best-effort verification after dirctl sign
**New dirctl naming commands** - verify and check subcommands
**Persistent storage** - Verification results stored as OCI referrers

Future Work
- Implement ListVerifiedAgents (requires domain index in DB)
- Re-verification based on TTL
- Name resolution (dirctl pull name instead of cid)

### DIFF
--- a/server/go.mod
+++ b/server/go.mod
@@ -58,8 +58,8 @@ require (
 	github.com/lestrrat-go/option v1.0.1 // indirect
 	github.com/moby/term v0.5.2 // indirect
 	github.com/pkg/browser v0.0.0-20240102092130-5ac0b6a4141c // indirect
+	github.com/segmentio/asm v1.2.1 // indirect
 	github.com/sigstore/cosign/v3 v3.0.4 // indirect
-	github.com/segmentio/asm v1.2.0 // indirect
 	github.com/sigstore/rekor-tiles/v2 v2.0.1 // indirect
 	github.com/sigstore/timestamp-authority/v2 v2.0.4 // indirect
 	github.com/tiendc/go-deepcopy v1.7.1 // indirect
@@ -223,7 +223,7 @@ require (
 	github.com/shibumi/go-pathspec v1.3.0 // indirect
 	github.com/sigstore/protobuf-specs v0.5.0 // indirect
 	github.com/sigstore/rekor v1.4.3 // indirect
-	github.com/sigstore/sigstore v1.10.3 // indirect
+	github.com/sigstore/sigstore v1.10.3
 	github.com/sigstore/sigstore-go v1.1.4 // indirect
 	github.com/sirupsen/logrus v1.9.4-0.20230606125235-dd1b4c2e81af // indirect
 	github.com/sourcegraph/conc v0.3.1-0.20240121214520-5f936abd7ae8 // indirect
--- a/server/go.sum
+++ b/server/go.sum
@@ -696,8 +696,8 @@ github.com/sassoftware/relic/v7 v7.6.2 h1:rS44Lbv9G9eXsukknS4mSjIAuuX+lMq/FnStgm
 github.com/sassoftware/relic/v7 v7.6.2/go.mod h1:kjmP0IBVkJZ6gXeAu35/KCEfca//+PKM6vTAsyDPY+k=
 github.com/secure-systems-lab/go-securesystemslib v0.9.1 h1:nZZaNz4DiERIQguNy0cL5qTdn9lR8XKHf4RUyG1Sx3g=
 github.com/secure-systems-lab/go-securesystemslib v0.9.1/go.mod h1:np53YzT0zXGMv6x4iEWc9Z59uR+x+ndLwCLqPYpLXVU=
-github.com/segmentio/asm v1.2.0 h1:9BQrFxC+YOHJlTlHGkTrFWf59nbL3XnCoFLTwDCI7ys=
-github.com/segmentio/asm v1.2.0/go.mod h1:BqMnlJP91P8d+4ibuonYZw9mfnzI9HfxselHZr5aAcs=
+github.com/segmentio/asm v1.2.1 h1:DTNbBqs57ioxAD4PrArqftgypG4/qNpXoJx8TVXxPR0=
+github.com/segmentio/asm v1.2.1/go.mod h1:BqMnlJP91P8d+4ibuonYZw9mfnzI9HfxselHZr5aAcs=
 github.com/sergi/go-diff v1.0.0/go.mod h1:0CfEIISq7TuYL3j771MWULgwwjU+GofnZX9QAmXWZgo=
 github.com/sergi/go-diff v1.4.0 h1:n/SP9D5ad1fORl+llWyN+D6qoUETXNZARKjyY2/KVCw=
 github.com/sergi/go-diff v1.4.0/go.mod h1:A0bzQcvG0E7Rwjx0REVgAGH58e96+X0MeOfepqsbeW4=


arpad-csepi (approved):

proto/agntcy/dir/naming/v1/naming_service.proto
```diff
@@ -0,0 +1,98 @@
...
+service NamingService {
+  // VerifyDomain performs domain ownership verification for a signed record.
+  // This should be called after signing a record to verify and store the domain ownership proof.
+  rpc VerifyDomain(VerifyDomainRequest) returns (VerifyDomainResponse);
+
+  // CheckDomainVerification checks if a record has verified domain ownership.
+  rpc CheckDomainVerification(CheckDomainVerificationRequest) returns (CheckDomainVerificationResponse);
+
+  // ListVerifiedAgents lists all agents that have verified domain ownership for a given domain.
+  rpc ListVerifiedAgents(ListVerifiedAgentsRequest) returns (ListVerifiedAgentsResponse);
```
ramizpolic: lets provide this functionality through local search service to avoid reimplementation.
apart from storing the information inside the manifest, we should also cache the results in the sqlite table (e.g. `verified_at`, `verification_status`, etc)

the relevant thing is to be able to search for records with a name "abc.com" that are verified, which is perfect usecase for local search

proto/agntcy/dir/naming/v1/naming_service.proto
```diff
@@ -0,0 +1,98 @@
...
+
+// NamingService provides methods to check domain verification status
+// and list verified agents for a domain.
+service NamingService {
+  // VerifyDomain performs domain ownership verification for a signed record.
+  // This should be called after signing a record to verify and store the domain ownership proof.
+  rpc VerifyDomain(VerifyDomainRequest) returns (VerifyDomainResponse);
+
+  // CheckDomainVerification checks if a record has verified domain ownership.
+  rpc CheckDomainVerification(CheckDomainVerificationRequest) returns (CheckDomainVerificationResponse);
```
ramizpolic: naming service should abstract the naming schema and only provide a way to initiate or inspect the name verification state.

i think the better approach here for RPCs is:

```
Verify() => provide CID to verify the name
GetVerificationInfo()
```

the name itself can support `dns://` or `wellknown://` (not sure about what you support with the current approach)
the get verification result can accept either a name or a record CID and return the details such as

```
message GetVerificationInfoRequest {
	oneof id {
		string record_cid = 1; // to check specific record name info
		string name = 2; // to check specific name
	}
}

message NameVerificationInfo {
	string name = 1;
	string status = 2; // verified/not-verified
	string verified_at = 3; // timestamp of the last verification state
}
```

api/core/v1/referrer_types.go
```diff
@@ -10,4 +10,7 @@ const (
 
 	// SignatureReferrerType is the type for Signature referrers.
 	SignatureReferrerType = "agntcy.dir.sign.v1.Signature"
+
+	// DomainVerificationReferrerType is the type for DomainVerification referrers.
+	DomainVerificationReferrerType = "agntcy.dir.naming.v1.DomainVerification"
```
ramizpolic:
```suggestion
	DomainVerificationReferrerType = "agntcy.dir.naming.v1.Verification"
```

can we do it more generally for all kinds of naming verification data?

proto/agntcy/dir/naming/v1/domain_verification.proto
```diff
@@ -0,0 +1,24 @@
...
+
+syntax = "proto3";
+
+package agntcy.dir.naming.v1;
+
+import "google/protobuf/timestamp.proto";
+
+// DomainVerification represents the result of verifying that a record's
+// signing key is authorized by the domain claimed in the record's name.
+message DomainVerification {
```
ramizpolic, line 22: can we generalize the verification details of a name

```
message VerificationInfo {
	oneof info {
		DomainVerification domain = 1;
		// whatever else comes later
	}
}
```

proto/agntcy/dir/naming/v1/name_verification.proto
```diff
@@ -0,0 +1,24 @@
...
+// signing key is authorized by the domain claimed in the record's name.
+message DomainVerification {
+  // The domain that was verified (e.g., "cisco.com").
+  string domain = 1;
+
+  // The verification method used: "dns" or "wellknown".
+  string method = 2;
+
+  // The ID of the domain key that matched the signing key.
+  string matched_key_id = 3;
```
ramizpolic: nit: use `key_id` and describe what the key_id is here as it is a bit unclear if its the hash of the public key or something else

server/controller/naming.go
```diff
@@ -0,0 +1,295 @@
...
+func NewNamingController(store types.StoreAPI, verifier *verification.Verifier) namingv1.NamingServiceServer {
+	return &namingCtrl{
+		store:    store,
+		verifier: verifier,
+	}
+}
+
+// VerifyDomain performs domain ownership verification for a signed record.
+func (n *namingCtrl) VerifyDomain(ctx context.Context, req *namingv1.VerifyDomainRequest) (*namingv1.VerifyDomainResponse, error) {
+	namingLogger.Debug("VerifyDomain request received", "cid", req.GetCid())
```
ramizpolic: here we need to first check if the domain verification info is cached or not (from sqlite).
if not, once we are done with the verification, we should cache it to sqlite

server/naming/dns/resolver.go
ramizpolic, line 1: can we put the naming related code changes to `server/naming` package and structure it in the following way

```
/server/naming
	/config - config for naming service and related providers
	/dns - dns verifier + resolver
		/config	- dns provider config
	/atprotocol - future when other configs are added
	/provider.go - creates the necessary provider based on the config to be used in services (can call multiple other ones based on the naming request, e.g. if name is "dns://" to be verified, then it calls the dns verifier)
```

server/verification/types.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// MethodNone indicates no verification was possible.
+	MethodNone VerificationMethod = "none"
+)
+
+// DNSRecordPrefix is the subdomain prefix for OASF DNS TXT records.
+const DNSRecordPrefix = "_oasf."
+
+// WellKnownPath is the path for the OASF well-known file.
+const WellKnownPath = "/.well-known/oasf.json"
```
ramizpolic: lets check if there is an existing `/.well-known/` for storing/using public keys.
defining our own well known format is rather hard to maintain and difficult to integrate (high complexity). we should focus on using existing standards for this.
if we can simplify here by using some existing system or providing a service that can do that out-of-box, it would be better than passing the responsibility for users to maintain this

check into https://www.dennisokeeffe.com/blog/2025-04-09-well-known-jwks and https://auth0.com/docs/secure/tokens/json-web-tokens/json-web-key-sets if we can leverage this instead of declaring our own.

proto/agntcy/dir/naming/v1/naming_service.proto
adamtagscherer: I removed the procedure, and also created a separate Issue to search for all verified records that belong to a specific domain.
I don't want to make this PR bigger, so let's do that separately.

https://github.com/agntcy/dir/issues/810

proto/agntcy/dir/naming/v1/naming_service.proto
adamtagscherer: Added the protocol prefix to the name field, and also renamed the procedures. Since there is still no SQL DB changes, and I want to do it in a separate PR that will be added in a follow up PR.

https://github.com/agntcy/dir/issues/810

server/controller/naming.go
adamtagscherer: Added to the follow up issue, because it needs SQL DB changes: https://github.com/agntcy/dir/issues/810

cli/cmd/naming/naming.go
```diff
@@ -0,0 +1,54 @@
...
+
+Protocol prefixes (required for verification):
+- dns://domain/path - verify using DNS TXT records
+- wellknown://domain/path - verify using well-known file
+
+Records without a protocol prefix will not be verified.
+
+Verification methods:
+1. DNS TXT record: _oasf.<domain> with format "v=oasf1; k=<type>; p=<key>"
+2. Well-known file: https://<domain>/.well-known/oasf.json
```
ramizpolic: my concern here is that this has nothing to do with OASF. this is DIR naming system specific data. Lets change this to reflect that.

ramizpolic: well-known approach should have a different type prefix perhaps: `https://` and `http://` maybe that will check into wellknowns

ramizpolic (approved):
LGTM

PR #808, opened 2026-01-13
feat(runtime): add runtime process discovery service

ramizpolic:
## Summary

This PR introduces an MVP runtime service that enables process discovery and lifecycle management for agent processes. Created as a shim-layer to expose a common API over any supported runtime. Supports **docker, containerd, kubernetes**. 

This directory contains the runtime discovery components that watch container runtimes
(Docker, Kubernetes) for workloads and provide a gRPC API for querying them.

## Architecture

The system is split into two independent components:

```
┌──────────────────────────────────────────────────────────────────────────────┐
│                          Runtime Discovery System                            │
└──────────────────────────────────────────────────────────────────────────────┘

┌─────────────────────────────┐         ┌──────────────────────────────────────┐
│         Discovery           │         │              Server                  │
│    (runtime/discovery)      │         │          (runtime/server)            │
│                             │         │                                      │
│  ┌─────────────────────┐    │         │    ┌─────────────────────────┐       │
│  │   Runtime Adapters  │    │         │    │       gRPC API          │       │
│  │   - Docker          │    │         │    │     /ListWorkloads      │       │
│  │   - Kubernetes      │    │         │    └─────────────────────────┘       │
│  └─────────────────────┘    │         │               ▲                      │
│            │                │         │               │                      │
│            ▼                │         └──────────────────────────────────────┘
│  ┌─────────────────────┐    │                         │                      
│  │     Resolvers       │    │                         │                      
│  │   - A2A             │    │                         │                      
│  │   - OASF            │    │                         │                      
│  └─────────────────────┘    │                         │                      
│            │                │                         │                      
│            ▼                │                         │                      
│  ┌─────────────────────┐    │                         │                      
│  │    Store Writer     │    │                         │                      
│  └─────────────────────┘    │                         │                      
└────────────│────────────────┘                         │                      
             │ write/manage                             │
             ▼                                          │
  ┌──────────────────────────────┐                      │
  │ Storage Backend              │       read/watch     │
  │  - etcd (distributed)        │──────────────────────┘
  │  - Kubernetes CRDs (native)  │
  └──────────────────────────────┘
```

### Discovery (`runtime/discovery/`)

The discovery component is responsible for:
- **Watching runtimes** for workloads with the `org.agntcy/discover=true` label. Supported runtimes:
  - **Docker**: Watches Docker daemon for labeled containers
  - **Kubernetes**: Watches Kubernetes API for labeled pods/services
  - Extensible architecture allows adding more runtimes in the future
- **Resolving workload metadata** using configurable resolvers:
  - **A2A resolver**: Extracts A2A agent card from workloads with `org.agntcy/type: a2a` label
  - **OASF resolver**: Resolves OASF records from Directory for workloads with `org.agntcy/agent-record` label
  - Extensible architecture allows adding more resolvers in the future
- **Writing workloads** to the storage backend (etcd or CRDs)

The storage backend can be used to expose discovered workloads to other components (e.g., clients/servers) without coupling them directly and to reduce attack surface.

In non-Kubernetes environments, etcd is recommended as the storage backend for better portability.
In Kubernetes environments, CRDs can be used for a more native experience to ensure clients can access workloads via the Kubernetes API.

```
Runtime Discovery --------> Runtime Server (for querying workloads via gRPC)
      │
      └-------------------> CRD (for querying workloads via Kubernetes API)
```

### Server (`runtime/server/`)

The server component provides a **gRPC API** for querying discovered workloads

### Docker Compose

```bash
# Deploy the stack
docker compose -f runtime/install/docker/docker-compose.yml up -d

# Deploy example workloads
docker compose -f runtime/install/examples/docker-compose.yml up -d

# Add discovery to all networks (required for resolvers to work)
docker network connect examples_team-a runtime-discovery
docker network connect examples_team-b runtime-discovery
docker restart runtime-discovery

# Query the API
grpcurl -plaintext localhost:8080 agntcy.dir.runtime.v1.DiscoveryService/ListWorkloads

# Cleanup
docker compose -f runtime/install/docker/docker-compose.yml down
docker compose -f runtime/install/examples/docker-compose.yml down
```

### Kubernetes

The Helm chart supports both CRD and etcd storage backends.

#### Setup KIND Cluster

```bash
# Create cluster
kind create cluster --name runtime

# Build images
docker build -t ghcr.io/agntcy/dir-runtime-discovery:latest -f runtime/discovery/Dockerfile .
docker build -t ghcr.io/agntcy/dir-runtime-server:latest -f runtime/server/Dockerfile .

# Load images into KIND
kind load docker-image ghcr.io/agntcy/dir-runtime-discovery:latest --name runtime
kind load docker-image ghcr.io/agntcy/dir-runtime-server:latest --name runtime
```

#### Deploy Example Workloads

```bash
kubectl apply -f runtime/install/examples/k8s.workloads.yaml
```

#### Deploy with CRD Storage

```bash
# Install the chart with CRD storage (default)
helm install runtime runtime/install/chart/

# Wait for pods
kubectl wait --for=condition=ready pod -l app.kubernetes.io/component=discovery --timeout=60s
kubectl wait --for=condition=ready pod -l app.kubernetes.io/component=server --timeout=60s

# Query the gRPC API
kubectl port-forward svc/runtime-server 8080:8080 &
grpcurl -plaintext localhost:8080 agntcy.dir.runtime.v1.DiscoveryService/ListWorkloads

# Query the Kubernetes API to see discovered workloads
kubectl get dw
```

#### Deploy with etcd Storage

```bash
# Install the chart with etcd storage
helm install runtime runtime/install/chart/ \
  --set etcd.enabled=true \
  --set discovery.config.store.type=etcd \
  --set server.config.store.type=etcd

# Wait for pods
kubectl wait --for=condition=ready pod -l app.kubernetes.io/component=etcd --timeout=60s
kubectl wait --for=condition=ready pod -l app.kubernetes.io/component=discovery --timeout=60s
kubectl wait --for=condition=ready pod -l app.kubernetes.io/component=server --timeout=60s

# Query the gRPC API
kubectl port-forward svc/runtime-server 8080:8080 &
grpcurl -plaintext localhost:8080 agntcy.dir.runtime.v1.DiscoveryService/ListWorkloads
```

#### Cleanup

```bash
kind delete cluster --name runtime
```

## Workload Labels

Workloads are discovered based on labels. The discovery component watches for workloads with specific labels and processes their metadata.

### Discovery Labels

| Label | Runtime | Description |
|-------|---------|-------------|
| `org.agntcy/discover=true` | Kubernetes | Marks a pod/service for discovery |
| `org.agntcy/discover=true` | Docker | Marks a container for discovery |

### Resolver Labels

| Label/Annotation | Description |
|------------------|-------------|
| `org.agntcy/agent-type: a2a` | Enables A2A resolver - fetches A2A agent card from workload |
| `org.agntcy/agent-record: <fqdn>` | Enables OASF resolver - resolves record from Directory (e.g., `my-agent:v1.0.0`) |

### Workload Services

Discovered workloads have a `services` field that holds metadata extracted by resolvers:

```json
{
  "id": "4467371c-84fd-4683-ab30-93895d78bab7",
  "name": "service-a2a",
  "hostname": "service-a2a",
  "runtime": "kubernetes",
  "type": "pod",
  "labels": {
    "app": "service-a2a",
    "org.agntcy/discover": "true",
    "org.agntcy/agent-type": "a2a",
    "org.agntcy/agent-record": "my-agent:1.0.0"
  },
  "addresses": [
    "10-244-0-9.team-a.pod"
  ],
  "ports": [
    "8080",
    "9999"
  ],
  "isolationGroups": [
    "team-a"
  ],
  "services": {
    "a2a": {
      "name": "My Agent",
      "description": "...",
      "capabilities": [...]
    },
    "oasf": {
      "cid": "baf123",
      "name": "my-agent",
      "record": {
        "name": "my-agent",
        "version": "1.0.0",
        "skills": [...]
      }
    }
  }
} 
```

### Discovery Component

| Environment Variable | Description | Default |
|---------------------|-------------|---------|
| `DISCOVERY_RUNTIME_TYPE` | Runtime type (`docker`, `kubernetes`) | `docker` |
| `DISCOVERY_STORE_TYPE` | Storage type (`etcd`, `crd`) | `etcd` |
| `DISCOVERY_RESOLVER_A2A_ENABLED` | Enable A2A resolver | `true` |
| `DISCOVERY_RESOLVER_OASF_ENABLED` | Enable OASF resolver | `true` |
| `DISCOVERY_WORKERS` | Number of resolver workers | `16` |

### Server Component

| Environment Variable | Description | Default |
|---------------------|-------------|---------|
| `SERVER_STORE_TYPE` | Storage type (`etcd`, `crd`) | `etcd` |
| `SERVER_HOST` | Server bind host | `0.0.0.0` |
| `SERVER_PORT` | Server listen port | `8080` |

## gRPC API

The server exposes a gRPC API defined in `proto/agntcy/dir/runtime/v1/discovery_service.proto`.

### GetWorkload

Get a specific workload by ID, name, or hostname.

```bash
grpcurl -plaintext -d '{"id": "my-service"}' \
  localhost:8080 agntcy.dir.runtime.v1.DiscoveryService/GetWorkload
```

### ListWorkloads

Stream all workloads with optional label filters. Labels support regex patterns.

```bash
# List all workloads
grpcurl -plaintext -d '{}' \
  localhost:8080 agntcy.dir.runtime.v1.DiscoveryService/ListWorkloads

# Filter by labels (supports regex)
grpcurl -plaintext -d '{"labels": {"org.agntcy/agent-type": "a2a"}}' \
  localhost:8080 agntcy.dir.runtime.v1.DiscoveryService/ListWorkloads
```

### DIFF
--- a/runtime/discovery/resolver/oasf/resolver.go
+++ b/runtime/discovery/resolver/oasf/resolver.go
@@ -102,10 +102,10 @@ func (r *resolver) Resolve(ctx context.Context, workload *runtimev1.Workload) (a
 
 	// Return the record data along with its validity
 	return map[string]any{
-		"cid":          recordRef.GetCid(),
-		"name":         nameVersion,
-		"verification": verified,
-		"record":       record.GetData().AsMap(),
+		"cid":      recordRef.GetCid(),
+		"name":     nameVersion,
+		"verified": verified,
+		"record":   record.GetData().AsMap(),
 	}, nil
 }
 


proto/agntcy/dir/runtime/v1/discovery_service.proto
```diff
@@ -0,0 +1,42 @@
...
+  // Labels are applied in an AND fashion; only workloads matching all labels are returned.
+  // Supports regular expression syntax for label values.
+  map<string, string> labels = 1;
+
+  // Optional runtime types to filter workloads.
+  // Only workloads matching one of the specified runtime types are returned.
+  // optional string runtime_type = 2;
+
+  // Reachability filter. If set, only workloads reachable from the specified workload ID are returned.
+  // optional string reachable_from_workload_id = 3;
```
tkircsi, line 41: Are they for future features?

proto/agntcy/dir/runtime/v1/workload.proto
```diff
@@ -0,0 +1,64 @@
...
+  string name = 2;
+
+  // Hostname is the hostname of the workload.
+  string hostname = 3;
+
+  // Runtime is the type of the runtime (e.g., docker, containerd).
+  string runtime = 4;
+
+  // WorkloadType is the type of workload (e.g., process, container, pod).
+  string workload_type = 5;
```
tkircsi: Can we define enum types for them?

proto/agntcy/dir/runtime/v1/workload.proto
```diff
@@ -0,0 +1,64 @@
...
+  map<string, string> labels = 6;
+
+  // Annotations are additional metadata.
+  map<string, string> annotations = 7;
+
+  // Addresses are the network addresses of the workload.
+  repeated string addresses = 8;
+
+  // Ports are the exposed ports.
+  repeated string ports = 9;
```
tkircsi, line 38: What are the valid formats of the port?

runtime/discovery/runtime/docker/adapter.go
```diff
@@ -0,0 +1,250 @@
...
+		}
+	}
+
+	return workloads, nil
+}
+
+// WatchEvents watches Docker events and sends workload events to the channel.
+//
+//nolint:wrapcheck
+func (d *adapter) WatchEvents(ctx context.Context, eventChan chan<- *types.RuntimeEvent) error {
```
tkircsi, line 100: If `WatchEvents` returns an error, it exits. The main loop doesn't restart it. Compare with K8s adapter `watchPods` which has infinite retry

runtime/discovery/runtime/k8s/adapter.go
```diff
@@ -0,0 +1,402 @@
...
+func (k *adapter) ListWorkloads(ctx context.Context) ([]*runtimev1.Workload, error) {
+	pods, err := k.clientset.CoreV1().Pods(k.namespace).List(ctx, metav1.ListOptions{
+		LabelSelector: fmt.Sprintf("%s=%s", k.labelKey, k.labelValue),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to list pods: %w", err)
+	}
+
+	// Get services for address building
+	servicesByNamespace := k.getServicesByNamespace(ctx)
```
tkircsi, line 96: Can we cache it and invalidate the cache at pod events?

runtime/discovery/resolver/a2a/resolver.go
```diff
@@ -0,0 +1,136 @@
...
+//
+//nolint:nosprintfhostport
+func (r *resolver) Resolve(ctx context.Context, workload *runtimev1.Workload) (any, error) {
+	// Build list of URLs to try
+	var urls []string
+
+	for _, addr := range workload.GetAddresses() {
+		for _, port := range workload.GetPorts() {
+			for _, path := range r.paths {
+				urls = append(urls, fmt.Sprintf("http://%s:%s%s", addr, port, path))
```
tkircsi, line 72: I'm not sure, so I ask it. These URLs can be only `http` ?

runtime/store/etcd/store.go
```diff
@@ -0,0 +1,209 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to delete workload from etcd: %w", err)
+	}
+
+	logger.Info("deregistered workload", "workload", workloadID)
+
+	return nil
+}
+
+// PatchWorkload performs a full update of an existing workload in storage.
```
tkircsi: Name suggests "patch" , but implementations do full replacement(update). Same for k8s

runtime/store/crd/store.go
```diff
@@ -0,0 +1,303 @@
...
+	_, err = s.client.Resource(s.gvr).Namespace(s.namespace).Create(ctx, obj, metav1.CreateOptions{})
+	if err != nil {
+		// If already exists, update it
+		if strings.Contains(err.Error(), "already exists") {
+			//nolint:wrapcheck
+			return s.PatchWorkload(ctx, workload)
+		}
+
+		return fmt.Errorf("failed to create CR: %w", err)
+	}
```
tkircsi, line 101: use typed errors 
```go
import "k8s.io/apimachinery/pkg/api/errors"

if errors.IsAlreadyExists(err) {
    return s.PatchWorkload(ctx, workload)
}

if errors.IsNotFound(err) {
    return nil
}
```

runtime/store/crd/store.go
```diff
@@ -0,0 +1,303 @@
...
+//nolint:mnd
+func crName(workloadID string) string {
+	name := strings.ToLower(workloadID)
+	name = strings.ReplaceAll(name, "_", "-")
+	name = strings.ReplaceAll(name, ":", "-")
+	name = "dw-" + name
+
+	if len(name) > 63 {
+		name = name[:63]
+	}
```
tkircsi, line 315: Possible name collision at very long names. add hash suffix.
```suggestion
    if len(name) > 63 {
        hash := fmt.Sprintf("%x", sha256.Sum256([]byte(workloadID)))[:8]
        name = name[:50] + "-" + hash
    }
```

runtime/server/cmd/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Start server in goroutine
+	go func() {
+		logger.Info("gRPC server listening", "address", cfg.Addr())
+
+		if err := grpcServer.Serve(listener); err != nil {
+			logger.Error("Server error", "error", err)
+			return
+		}
+	}()
```
tkircsi, line 75: If `Serve()` errors, the goroutine exits silently and `main` thread waits forever.

runtime/store/types/store.go
```diff
@@ -0,0 +1,49 @@
...
+	// PatchWorkload performs a full update of an existing workload in storage.
+	PatchWorkload(ctx context.Context, workload *runtimev1.Workload) error
+
+	// Close closes the storage connection.
+	Close() error
+}
+
+// StoreReader is the interface for reading workload data from storage.
+// Used by the server component to rebuild database state.
+type StoreReader interface {
```
tkircsi, line 32: Checking the code base, I think, add this method will be useful. At some places using `ListWorkloads` just to find one Workload.
`GetWorkload(ctx context.Context, id string) (*runtimev1.Workload, error)`

runtime/server/grpc/discovery.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		return status.Errorf(codes.Internal, "failed to list workloads: %v", err)
+	}
+
+	// Apply label filters if provided
+	for _, workload := range workloads {
+		if err := stream.Send(workload); err != nil {
+			return status.Errorf(codes.Internal, "failed to send workload: %v", err)
+		}
+	}
```
tkircsi, line 59:
```suggestion
for _, workload := range workloads {
    select {
    case <-ctx.Done():
        return status.Error(codes.Canceled, "client disconnected")
    default:
    }
    
    if err := stream.Send(workload); err != nil {
        return status.Errorf(codes.Internal, "failed to send workload: %v", err)
    }
}```

File: runtime/install/chart/templates/etcd-deployment.yaml [hunk 13]
```diff
@@ -0,0 +1,107 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
+{{- if .Values.etcd.enabled }}
+apiVersion: apps/v1
+kind: Deployment
```
tkircsi, line 6: For prod it would be better to use StatefulSet instead of Deployment and maybe it worths to consider using bitnami etcd chart or something similar.

proto/agntcy/dir/runtime/v1/discovery_service.proto
ramizpolic: Yes, I did not want to add them in the current implementation because these features can be added separately.

proto/agntcy/dir/runtime/v1/workload.proto
ramizpolic: I wanted to avoid adding them right away in the current PR, but yes, we can add these as a an enum later. WDYT to do it separately?

runtime/server/cmd/main.go
```diff
@@ -0,0 +1,81 @@
...
+	if err != nil {
+		logger.Error("Failed to initialize database", "error", err)
+		return
+	}
+	defer db.Close()
+
+	// Create gRPC server
+	grpcServer := grpc.NewServer()
+	discoveryServer := grpcserver.NewDiscovery(db)
+	runtimev1.RegisterDiscoveryServiceServer(grpcServer, discoveryServer)
```
tkircsi, line 51: It would worh to add grpc health check using standard library, than using only TCP probe.

```go
import (
    "google.golang.org/grpc/health"
    "google.golang.org/grpc/health/grpc_health_v1"
)
```



```suggestion
// Create gRPC server with health checks
grpcServer := grpc.NewServer()

// Health check service
healthServer := health.NewServer()
grpc_health_v1.RegisterHealthServer(grpcServer, healthServer)

// Check database connectivity
if db.Count(context.Background()) >= 0 {
    healthServer.SetServingStatus("", grpc_health_v1.HealthCheckResponse_SERVING)
    healthServer.SetServingStatus("agntcy.dir.runtime.v1.DiscoveryService", 
                                   grpc_health_v1.HealthCheckResponse_SERVING)
} else {
    healthServer.SetServingStatus("", grpc_health_v1.HealthCheckResponse_NOT_SERVING)
}

// Register discovery service
discoveryServer := grpcserver.NewDiscovery(db)
runtimev1.RegisterDiscoveryServiceServer(grpcServer, discoveryServer)
```

requires helm chart change 
Kubernetes 1.24+
```yaml
livenessProbe:
  grpc:
    port: 8080
  initialDelaySeconds: 10
  periodSeconds: 10
  
readinessProbe:
  grpc:
    port: 8080
  initialDelaySeconds: 5
  periodSeconds: 5
```

For Kubernetes 1.24< use `grpc-health-probe`

proto/agntcy/dir/runtime/v1/workload.proto
tkircsi: It's fine for me.

runtime/discovery/runtime/k8s/adapter.go
ramizpolic: No need for this now, we can add optimizations later

runtime/discovery/resolver/a2a/resolver.go
ramizpolic: Yes, probing for a well-known on a pod-level should only flow through HTTP. although HTTPS can also work, it may not be available.

runtime/store/crd/store.go
ramizpolic: name collisions are fine and do not need any special handling because the main identifiers are ID-based, which cannot collide

runtime/server/cmd/main.go
ramizpolic: we can do this separately, for now only a probe is enough because the server does nothing except expose simple API over storage

runtime/install/chart/templates/etcd-deployment.yaml
ramizpolic: i think its okay for now as-is, id optimize separately

runtime/store/types/store.go
ramizpolic: this is only done on the server side, and IMO its okay because the more important part is getting many vs getting one. once the other feature is there (get by isolation group), this will be more relevant and can be added then

proto/agntcy/dir/runtime/v1/workload.proto
ramizpolic: updated docs

runtime/discovery/runtime/docker/adapter.go
ramizpolic: i have added a check if watchevents fails to restart the app so that the app itself handles the retries through restarts. both k8s and docker can fail and are not handling retries.

runtime/server/cmd/main.go
ramizpolic: added a case to handle this

runtime/server/database/db.go
```diff
@@ -0,0 +1,111 @@
...
+	ids, err := d.store.ListWorkloadIDs(ctx)
+	if err != nil {
+		return 0
+	}
+
+	return len(ids)
+}
+
+func (d *Database) Get(ctx context.Context, id string) (*v1.Workload, error) {
+	results, err := d.store.ListWorkloads(ctx)
```
muscariello: fetching all workloads to find a single one by ID will scale poorly (O(N)). 
We should add a GetWorkload(ctx, id) method to the StoreReader interface and implement efficient single-key lookups in the etcd/crd backends.
For k8s we cannot assume this number to be small in general.

A suggestion would be to Implement efficient single-key lookup in etcd and crds.

runtime/utils/struct.go
```diff
@@ -0,0 +1,33 @@
...
+package utils
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func InterfaceToStruct(input any) (*structpb.Struct, error) {
```
muscariello, line 13: This is doing Marshal + Unmarshal which is expensive.

proto/agntcy/dir/runtime/v1/workload.proto
```diff
@@ -0,0 +1,80 @@
...
+  // Specified as strings to accommodate various formats (e.g., "80/tcp", "443").
+  repeated string ports = 9;
+
+  // IsolationGroups are the network/namespace isolation groups.
+  repeated string isolation_groups = 10;
+
+  // Services holds service/protocol-specific metadata.
+  // These values can be used by discovery resolvers
+  // to enrich the discovery information about the workload.
+  WorkloadServices services = 11;
```
muscariello, line 46: If a resolver fails (e.g., A2A probe timeout), the error is logged but not exposed here. Clients won't know if discovery is partial or failed.

maybe add a repeated string status_conditions or similar field to allow reporting resolver errors to the end user

runtime/discovery/runtime/docker/adapter.go
```diff
@@ -0,0 +1,250 @@
...
+	// Extract networks
+	var (
+		addresses       []string
+		isolationGroups []string
+		ports           []string
+	)
+
+	if inspect.NetworkSettings != nil {
+		for netName := range inspect.NetworkSettings.Networks {
+			addresses = append(addresses, name+"."+netName)
```
muscariello, line 217: This implicitly requires the consumer to be on the same Docker network or have specific DNS configuration. This dependency should be documented.

runtime/utils/struct.go
ramizpolic: There is no other way to convert a generic struct into a proto struct without doing marshal/unmarshal this way. This is showcased with a test. 

Note that this happens maximum once per resolver per workload, does not happen again, and is not large enough to cause issues.

runtime/discovery/runtime/docker/adapter.go
ramizpolic: This is how docker networking handles workload resolution, I do not think that we need to do anything here as its a common behaviour on docker side. Similar to how internal traffic resolution works for k8s namespaced services/pods.

runtime/server/database/db.go
ramizpolic: Resolved in latest commit

proto/agntcy/dir/runtime/v1/workload.proto
ramizpolic: Resolved with latest commit by using a `{"error": "<error messages>"}` in case resolvers fail for a given service.

Issue #346, opened 2025-09-02
[Epic] Reusable Reconciler service


ramizpolic:
### Problem Statement

Currently in the directory API, several services require post-processing tasks that reimplement Reconciler mechanisms. For example, post-sync operations need to loop through the things and perform indexing, load public keys, validate, etc. Other services may also require reconciliation in the future, such as routing, naming, garbage collection, etc.

### Proposed Solution

Define a reusable Reconciler service that allows registration of reusable interfaces so that reconciler can perform logic centrally under `server/reconciler` with a reusable `Reconciler` interface. The Reconciler would expose simple primitives to the worker jobs like DBs, APIs, etc.

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

PR #809, opened 2026-01-13
refactor(dir): move signature verification from client to server

paralta:
This PR refactors the signature verification logic by moving it from the client side to the server side, simplifying the client and centralising verification logic.

### Changes:
* Simplified client Verify method to a thin wrapper that calls the server's `SignService.Verify` RPC
* Added server-side verification in `server/store/oci/signatures.go` supporting both Zot GraphQL API and OCI referrers fallback

### Validation Results:
* GitHub Container Registry (GHCR) signature and verification successfully tested
* Docker Hub signature and verification successfully tested


server/store/oci/signatures.go
```diff
@@ -155,8 +160,44 @@ func (s *store) convertCosignSignatureToReferrer(blobDesc ocispec.Descriptor, da
...
+// VerifySignature verifies a record signature using the appropriate method
+// based on the configured registry type.
+//
+// For Zot registries: Uses Zot's GraphQL API as a fast path for verification.
+// If Zot verification fails, falls back to standalone verification.
+//
+// For other registries (GHCR, DockerHub): Uses standalone verification
+// which retrieves signatures and public keys from OCI referrers.
+func (s *store) VerifySignature(ctx context.Context, recordCID string) (bool, error) {
+	switch s.config.GetType() {
```
ramizpolic: caching the verification state in the sqlite would definitely be useful in the future to avoid recompute.
this can be an additional feature to add later (cache verification status). the relates closely to the work @adamtagscherer is doing on naming verification because it allows us to quickly gather things for the local side:

- get all records that are signed (signature verification)
- get all records that are named (naming verification)

both verification services can be part of the worker service that runs alongside Dir node (implementation details that we can discuss later)

ramizpolic: remember that Zot already caches the verification status itself through the graphql

paralta: Thanks for the review! I agree that we should have the caching logic that Zot already provides to enable filtering trusted records. However, without a Reconciler service in place, we can't implement this properly. I'll add a comment to the Reconciler epic (#346) to track exposing this functionality.

ramizpolic: I am not sure if that's the case. Once we verify the record, we can cache the result into the DB and check if we have validation cached before re-verifying.

The reconciler is only responsible for doing work on items which dont have cached verification result, and can be done later. However, caching the results now and adding it right away simplifies the work that needs to be done once the reconciler work begins.

server/store/oci/signatures.go
ramizpolic, line 1: nit: i would suggest that all the backend code for signing/verification to be part of a separate package like `server/signing`. imo, services such as routing/naming/storage/signing/etc should be properly decoupled

server/store/oci/signatures.go
paralta: We could cache the trust status when signing/verifying (also post-sync) and return this cached value on subsequent verify calls, but without a reconciler there's no way to revoke trust if for example the signature or public key referrers are removed. This would follow Zot's pattern of using a scheduled task that periodically re-validates signatures and updates their validity status.

server/store/oci/signatures.go
paralta: Sounds good as well, I followed this path because of how signatures are so tightly coupled with the storage's OCI referrers

paralta: I agree with the separation of services suggested, let me work on these changes ⌛

paralta: updated

server/types/signing.go
```diff
@@ -0,0 +1,27 @@
...
+	// Verify verifies a record signature.
+	// Returns true if the signature is valid and trusted.
+	Verify(ctx context.Context, recordCID string) (bool, error)
+
+	// PushSignature attaches a signature to a record using cosign.
+	PushSignature(ctx context.Context, recordCID string, referrer *corev1.RecordReferrer) error
+
+	// UploadPublicKey uploads a public key to the registry for signature verification.
+	// For Zot registries, this uploads to the _cosign directory.
+	UploadPublicKey(ctx context.Context, referrer *corev1.RecordReferrer) error
```
ramizpolic: should these values have the actual datatypes from the signing service?

they are only converted to OCI artifacts under the hood (in the actual implementation).
this also means that they dont need to be implemented as part of OCI since we can decide for example to cache it to sqlite vs to OCI (just an example)


```suggestion
	LinkSignature(ctx context.Context, recordCID string, signature *signv1.Signature) error

	// UploadPublicKey uploads a public key to the registry for signature verification.
	// For Zot registries, this uploads to the _cosign directory.
	LinkPublicKey(ctx context.Context, pubKey *signv1.PublicKey) error
```

id also suggest to use a common interface for both signature/publickey like `LinkX` or `PushX` because they do the same thing -- create a linkage between a record and the datamodel

server/store/oci/signatures.go
ramizpolic: nicely done @paralta

server/types/signing.go
paralta: Indeed, these are not signing types just methods to support the sign service.

I think the common interface might not be adequate here because, while the `PushSignature` is building the signature and leveraging cosign to attach the signature (not just pushing the OCI signature referrer), `UploadPublicKey ` is using zot to upload the public keys https://zotregistry.dev/v2.0.1/articles/verifying-signatures/?h=public+keys#what-is-needed-for-verifying-signatures (after we create the public key referrers).

paralta: > Indeed, these are not signing types just methods to support the sign service.

After exploring the separation, I think it overcomplicates the store controller which needs all three "signing" related methods. I suggest we keep them as part of `SigningAPI` for now.

The alternative would be to move these two methods back to the store service.

paralta: @ramizpolic please have a look at the latest changes and let me know what you think 👀

PR #813, opened 2026-01-14
release(dir): prepare v0.6.1

paralta:


arpad-csepi (approved):
Please update REAMDE.md as well to use the new versions in the example

paralta:
> Please update REAMDE.md as well to use the new versions in the example

@arpad-csepi nice catch!

PR #826, opened 2026-01-19
feat(dir): re-verification service & verification storage

adamtagscherer:
**Overview**

Implements automatic re-verification of domain ownership with persistent storage in SQLite. Records are periodically re-verified to ensure domain ownership remains valid over time.

**Why SQLite Instead of OCI Referrers?**
Previously, verification results were stored as OCI referrers attached to records. Moved to SQLite for several reasons:

- **Verifications are mutable:** OCI referrers are designed for immutable metadata. Verifications change state (verified → expired → re-verified), which doesn't fit the append-only referrer model.
- **Efficient querying:** Finding expired verifications requires querying by timestamp. SQLite: WHERE updated_at < ? is O(log n) with an index. OCI: would require walking all referrers of all records.
- **Operational vs. permanent data:** Signatures and public keys are permanent record metadata (belong in OCI). Verification status is operational/cache data (belongs in a database).
- **Unique constraint:** One verification per CID is enforced by SQLite. OCI referrers have no such constraint, leading to duplicate verification records.

**Features**
- Re-verification Service
- Background scheduler-worker pattern
- Configurable TTL (default: 1week): how long before a verification expires
- Configurable check interval (default: 1h): how often to scan for expired verifications
- Always enabled (no feature flag)

**Verification Storage**
- One verification record per CID (unique constraint)
- Tracks: method, key_id, status, error message
- Uses updated_at for TTL expiration checks
- Error message cleared when verification succeeds after failure

### DIFF
--- a/server/controller/store.go
+++ b/server/controller/store.go
@@ -257,7 +257,7 @@ func (s storeCtrl) pushReferrer(ctx context.Context, request *storev1.PushReferr
 		}
 	}
 
-	_, err := refStore.PushReferrer(ctx, recordCID, request.GetReferrer())
+	err := refStore.PushReferrer(ctx, recordCID, request.GetReferrer())
 	if err != nil {
 		errMsg := fmt.Sprintf("failed to push referrer for record %s: %v", recordCID, err)
 
--- a/server/store/eventswrap/eventswrap.go
+++ b/server/store/eventswrap/eventswrap.go
@@ -101,31 +101,16 @@ func (s *eventsStore) IsReady(ctx context.Context) bool {
 
 // PushReferrer delegates to the source store if it supports referrer operations.
 // This is needed for signature and public key storage.
-// Returns the digest of the stored referrer manifest.
-func (s *eventsStore) PushReferrer(ctx context.Context, recordCID string, referrer *corev1.RecordReferrer) (string, error) {
+func (s *eventsStore) PushReferrer(ctx context.Context, recordCID string, referrer *corev1.RecordReferrer) error {
 	// Check if source supports referrer operations
 	referrerStore, ok := s.source.(types.ReferrerStoreAPI)
 	if !ok {
-		return "", status.Errorf(codes.Unimplemented, "source store does not support referrer operations")
+		return status.Errorf(codes.Unimplemented, "source store does not support referrer operations")
 	}
 
 	// Delegate to source (no event emitted for referrer operations)
 	//nolint:wrapcheck
 	return referrerStore.PushReferrer(ctx, recordCID, referrer)
-}
-
-// GetReferrer delegates to the source store if it supports referrer operations.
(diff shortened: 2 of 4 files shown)


server/reverification/config/config.go
```diff
@@ -0,0 +1,71 @@
...
+	DefaultSchedulerInterval = 1 * time.Hour
+
+	// DefaultWorkerCount is the default number of workers.
+	DefaultWorkerCount = 1
+
+	// DefaultWorkerTimeout is the default timeout for each verification.
+	DefaultWorkerTimeout = 30 * time.Second
+
+	// DefaultTTL is the default time-to-live for verifications.
+	DefaultTTL = 24 * time.Hour
```
ramizpolic: i would suggest the verification to be cached for longer than 24h, maybe 7 days

```go
	DefaultTTL = 7 * 24 * time.Hour
```

server/database/sqlite/naming.go
ramizpolic, line 1: can we rename this to `naming` component since it relates to both naming and verification

server/types/adapters/record.go
```diff
@@ -47,3 +47,19 @@ func (r *RecordAdapter) GetRecordData() (types.RecordData, error) {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get record data: %w", err)
+	}
+
+	if recordData == nil {
+		return "", nil
+	}
+
+	return recordData.GetName(), nil
+}
```
ramizpolic: this is not needed, when we need to access the name, we should do

```go

recordData, _ = adapter.GetRecordData()
recordData.GetName() // returns the name

```

server/types/database.go
```diff
@@ -80,3 +84,18 @@ type PublicationDatabaseAPI interface {
...
+
+type NameVerificationDatabaseAPI interface {
+	// CreateNameVerification creates a new name verification for a record.
+	CreateNameVerification(verification NameVerificationObject) error
+
+	// UpdateNameVerification updates an existing name verification for a record.
+	UpdateNameVerification(verification NameVerificationObject) error
+
+	// GetVerificationByCID retrieves the verification for a record.
+	GetVerificationByCID(cid string) (NameVerificationObject, error)
```
ramizpolic, line 99: We will need the `GetVerificationByName` as well which returns a list

ramizpolic (approved):

server/types/database.go
adamtagscherer: I will add it in a follow up PR, when we will use it immediately.

server/controller/naming.go
ramizpolic: should we keep this? users should not be able to submit verification jobs on the server-side IMO.
the controller should only respond with `verified/not-verified`. the verification process is done in the background by the worker you added

PR #831, opened 2026-01-21
feat(gui): search results UI widgets

vaesposito:
## Summary
Enhances the GUI search experience with a new search results widget that loads agent details.

https://github.com/user-attachments/assets/92a21447-ef3d-4d55-8271-6ab42ebf45aa

## Changes
- Added `SearchResultsWidget` with pagination and sorting (by name/date)
- Implemented auto-pull of agent records when search results are received
- Added `AgentRecordCard` showing name, author, version, CID with copy button
- Added `AgentDetailCard` for expanded view when "See more" is clicked
- Fixed duplicate search widgets bug
- Removed elastic scroll bounce effect
- Added smooth scroll animation for new content
- Improved MCP response parsing for `record_data`
- Added OpenAI-compatible provider support for AI Gateway
- Disabled OASF API validation for local development

## Testing
Tested locally with Docker Compose backend and 21 test agents.


vaesposito:
Buf CI check may fail because PRs from forks don't have access to the BUF_TOKEN secret.

vaesposito:
✅ Fixed and pushed! (according to cursor)

muscariello (changes requested):
@vaesposito I would now change the header of the GUI to something like 

AGNTCY Agent Directory GUI

Also I would think about using the colors of the AGNTCY website both in light and dark mode.

PR #834, opened 2026-01-21
feat(dir): dynamic policy for authz

arpad-csepi:


Taskfile.yml
```diff
@@ -1410,7 +1410,12 @@ tasks:
                 - "spiffe://{{ .DIR_TRUST_DOMAIN }}/spire/server"
             authz:
               enabled: true
-              trust_domain: {{ .DIR_TRUST_DOMAIN }}
+              enforcer_policy_file_content: |
+                p,{{ .DIR_TRUST_DOMAIN }},*
```
ramizpolic, line 1414: @arpad-csepi can you confirm that this resolution works?

install/charts/dir/values.yaml
```diff
@@ -103,9 +103,14 @@ apiserver:
     authz:
       # Enable authorization policies
       enabled: false
-      # Trust domain for this Directory server
-      # Used to distinguish internal (same trust domain) vs external requests
-      trust_domain: "example.org"
+
+      # Content of the authorization policy file
+      enforcer_policy_file_content: |
```
ramizpolic: This is user-facing, so it would be beneficial to include some documentation on how:
- The policy is defined (e.g. Policy on DIR side is enforced through prefix and regexp matching)
- Define some examples on how you can define the schema, similar to what we had before (e.g. p, <your-trust-domain>/service/admin/*, * // enable all APIs for admin services from your SPIFFE trust domain, etc)

ramizpolic: also `enforcer_policy_file_content` can simply be renamed to `enforcer_policy.csv` to make it more clear what this is

install/charts/dir/apiserver/values.yaml
```diff
@@ -46,9 +46,14 @@ config:
   authz:
     # Enable authorization policies
     enabled: false
-    # Trust domain for this Directory server
-    # Used to distinguish internal (same trust domain) vs external requests
-    trust_domain: "example.org"
+
+    # Content of the authorization policy file
+    enforcer_policy_file_content: |
```
ramizpolic: same notes here as below

install/charts/dir/apiserver/templates/deployment.yaml
```diff
@@ -220,6 +220,11 @@ spec:
             {{- with .Values.extraVolumeMounts }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
+            {{- if .Values.config.authz.enforcer_policy_file_content }}
+            - name: config-volume
+              mountPath: /etc/agntcy/dir/authz_policies.csv
+              subPath: authz_policies.csv
+            {{- end }}
```
ramizpolic: two things here:
- 1. we are missing the check if the authz is enabled
- 2. we are missing the env var to set the path for authz policy file in the deployment

install/charts/dir/values.yaml
arpad-csepi: The dot character inside a yaml key name is cumbersome to define and override, so I did not change to the recommended name.

Taskfile.yml
arpad-csepi: Yes, during development, the policies are enforced and denied when this was not correct, the Taskfile variable correctly resolved.

README.md
```diff
@@ -136,6 +136,116 @@ All release binaries are distributed via [GitHub Releases](https://github.com/ag
 
 - **JavaScript** - [npmjs.com/agntcy-dir](https://www.npmjs.com/package/agntcy-dir) - [github.com/agntcy/dir/sdk/dir-js](https://github.com/agntcy/dir/tree/main/sdk/dir-js)
 
+## Security
+
+### Authorization Policies
```
ramizpolic: lets move this to a separate `SECURITY.md` file instead of being part of main readme

install/charts/dir/values.yaml
ramizpolic: then perhaps `enforcer_policy_csv` would be better

PR #840, opened 2026-01-23
fix(gui): bundle mcp-server in macOS build and update path resolution

muscariello:
Signed-off-by: Luca Muscariello <muscariello@ieee.org>

### DIFF
--- a/gui/.metadata
+++ b/gui/.metadata
@@ -15,7 +15,22 @@ migration:
     - platform: root
       create_revision: f6ff1529fd6d8af5f706051d9251ac9231c83407
       base_revision: f6ff1529fd6d8af5f706051d9251ac9231c83407
+    - platform: android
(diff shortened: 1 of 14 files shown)


.github/workflows/gui-ci.yaml
ramizpolic, line 1: the releases for the GUI will only be available in the actions pipeline, not as part of release artifacts. 
is this okay @muscariello?

muscariello: Sorry I moved this into another PR as the branch was an old one already closed. Check #843

muscariello: I think we should have path on the run and build separately with a separate tag: gui* for instance. 

Test coverage is important for future update in case we have UX designers using code assistants.

PR #842, opened 2026-01-23
feat(dir): dirctl pull by name

adamtagscherer:
### Features:
`dirctl pull/info <name[:version][@digest]>` - resolve records by name instead of CID
`dirctl search` --verified - filter for records with verified name ownership
NamingService.Resolve RPC returning version history


proto/agntcy/dir/naming/v1/naming_service.proto
```diff
@@ -7,12 +7,20 @@ package agntcy.dir.naming.v1;
...
 // Note: Verification is performed automatically by the backend scheduler
 // for signed records with verifiable names (http://, https:// prefixes).
 service NamingService {
   // GetVerificationInfo retrieves the verification info for a record.
   rpc GetVerificationInfo(GetVerificationInfoRequest) returns (GetVerificationInfoResponse);
+
+  // Resolve resolves a record reference (name with optional version) to a single CID.
+  // Supports Docker-style references:
+  //   - "name" -> resolves to the latest version (by semver)
+  //   - "name:version" -> resolves to the specific version
```
ramizpolic:
```suggestion
  // Resolve resolves a record reference (name with optional version) to a single CID.
  // Supports Docker-style references:
  //   - "name" -> resolves to the latest version
  //   - "name:version" -> resolves to the specific version
  //   - "name:version@cid" -> resolves to a hash-verified record 
```

some notes here: resolution is also the process of confirmation of two things:
- without hash, records that are named/versioned
- with hash, hash-verification of a named/versioned record
it is important that we also support hash-suffixed lookups so that users can always know that this named agent is hash-verified (similar to containers, etc)

proto/agntcy/dir/naming/v1/naming_service.proto
```diff
@@ -32,3 +40,19 @@ message GetVerificationInfoResponse {
...
   optional string error_message = 3;
 }
+
+// ResolveRequest is the request for resolving a record reference to a CID.
+message ResolveRequest {
+  // The name of the record to resolve (e.g., "cisco.com/agent").
+  string name = 1;
+
+  // Optional version to resolve to (e.g., "v1.0.0").
+  // If not specified, resolves to the latest version by semver.
```
ramizpolic: avoid using "semver" across the code because the directory does not enforce the versioning type, the version schema can be anything (semver, timestamps, colors, etc)

proto/agntcy/dir/naming/v1/naming_service.proto
```diff
@@ -7,12 +7,20 @@ package agntcy.dir.naming.v1;
...
   // GetVerificationInfo retrieves the verification info for a record.
   rpc GetVerificationInfo(GetVerificationInfoRequest) returns (GetVerificationInfoResponse);
+
+  // Resolve resolves a record reference (name with optional version) to a single CID.
+  // Supports Docker-style references:
+  //   - "name" -> resolves to the latest version (by semver)
+  //   - "name:version" -> resolves to the specific version
+  // Returns an error if no matching record is found or if multiple records
+  // match the same name+version (ambiguous).
+  rpc Resolve(ResolveRequest) returns (ResolveResponse);
```
ramizpolic, line 25: consider adding additional method that resolves many, because that would support getting the version history of a given record

```proto
rpc ResolveMany(ResolveRequest) returns (stream ResolveResponse);
```

i think it may be sufficient to only have one method for resolution that can optionally return multiple resolved CIDs, for example

```
message ResolveResponse {
	repeated NamedRecordRef named_refs = 1; 
}

message NamedRecordRef {
	string name = 1;
	string version = 2;
	string cid = 3;
}
```

cli/util/reference/reference.go
```diff
@@ -0,0 +1,77 @@
...
+type Ref struct {
+	// Name is the record name (e.g., "cisco.com/agent").
+	Name string
+
+	// Version is the optional version (e.g., "v1.0.0").
+	Version string
+
+	// CID is set if the input is a raw CID (direct pull by content address).
+	CID string
+}
```
ramizpolic, line 28: i would consider moving this to the `core` proto package and define it from there because the named objects become reusable across the stack

```proto
// Represents a named reference to a Record
message NamedRecordRef {
	string name = 1;
	string version = 2;
	string cid = 3;
}

cli/cmd/pull/pull.go
ramizpolic, line 1: this is only available on pull, can we do the same for info command?

separate question, what options do we have on search in regards to named records?

ramizpolic: can we do `dirctl search --verified`

proto/agntcy/dir/naming/v1/naming_service.proto
adamtagscherer: I assumed that the version field of a record has to be semver, and when we pull a record just by the name e.g.: `dirctl pull my-agent.agntcy.org` then it will pull the record with the latest version, where the versions are compared assuming they are semver.

If this is not the case then
1. we need to enforce this field to be semver in the OASF JSON
2. we can pull the latest uplaoded record comparing the `created_at` field

ramizpolic: latest created_at is okay IMO

PR #847, opened 2026-01-23
feat(deps): update with OASF SDK v0.0.16

### LINKED ISSUE #815: Update with OASF SDK 0.0.15
The next version of OASF SDK will have a simplified validator, bump the SDK version and make the necessary changes in server, mcp, import, sync etc. packages where validation is used.

akijakya:
The new OASF SDK removed the embedded OASF schemas and simplified validator configurations, so packages using these methods needed to be adjusted to it. There is only one validation mode going forward, which requires an OASF URL.

Closes #815


api/core/v1/record.go
```diff
@@ -25,50 +25,23 @@ const (
 )
 
 var (
-	defaultValidator     *validator.Validator
-	configMu             sync.RWMutex
-	schemaURL            = "" // Set via SetSchemaURL() from configuration
-	disableAPIValidation = false
-	strictValidation     = true
+	defaultValidator *validator.Validator
+	configMu         sync.RWMutex
```
ramizpolic, line 29: we need to rethink this, may not be ideal to use globals in this package for validation

akijakya: would it be better if `record.Validate()` would require a Validator instance, or schema url to be passed to it?

ramizpolic: lets create an issue for this @akijakya

akijakya: https://github.com/agntcy/dir/issues/856

PR #850, opened 2026-01-25
feat(gui): add Google Analytics via Measurement Protocol

muscariello:
## Description

Adds Google Analytics 4 support to the macOS GUI app using the Measurement Protocol.

### Changes
- **Analytics Service**: Implemented `AnalyticsService` to handle GA4 events via HTTP (lightweight, no Firebase SDK).
- **Events**: Added automatic logging for:
  - App Launch
  - Message Sent (with provider/model metadata)
  - Tool Usage (recording which tools are called)
- **Remote Config**: Measurement ID and API Secret are fetched from `gui/analytics_config.json` in the repo, allowing dynamic updates without app store releases.
- **Documentation**: Added `GUI_GUIDE.md`.

### Implementation Details
- Uses `shared_preferences` to persist a random Client ID.
- Generates a fresh Session ID on each app launch.
- Falls back to default keys if remote config fails.


.github/workflows/gui-ci.yaml
```diff
@@ -68,6 +80,10 @@ jobs:
         with:
           go-version: '1.25.6'
 
+      - name: Inject Analytics Config
+        run: |
+          echo "{\"measurement_id\": \"$MEASUREMENT_ID\", \"api_secret\": \"$API_SECRET\"}" > analytics_config.json
```
ramizpolic: do we need this file or are we passing these details through env vars?

muscariello: leftover. 
We need to merge and see if the action works.

PR #876, opened 2026-01-28
feat(apiserver): update database configuration to default to PostgreSQL

paralta:
Changes the default database from SQLite to PostgreSQL for production-ready deployments. Bundles the PostgreSQL helm chart to fix CI timeout issues.

**Changes**
* Set `database.type` default to `postgres` (was `sqlite`)
* Set `postgresql.enabled` default to `true` (was `false`)
* Added `postgresql-18.2.3.tgz` chart binary to repository

### DIFF
--- a/install/charts/dir/apiserver/values.yaml
+++ b/install/charts/dir/apiserver/values.yaml
@@ -662,11 +662,11 @@ envoy-authz:
       nginx.ingress.kubernetes.io/grpc-backend: "true"
 
 # PostgreSQL subchart configuration
-# Deploy PostgreSQL alongside the apiserver.
-# Enabled by default when database.type is "postgres".
-#
-# CREDENTIALS: When enabled, postgresql.auth.* is the single source of truth.
-# The apiserver will automatically use these credentials (no need to also set secrets.postgresAuth.*).
+# Deploys PostgreSQL alongside the apiserver for dev/production use.
+# Set enabled: false when using SQLite or external PostgreSQL.
+# Note: database.type (above) controls which backend the apiserver uses;
+#       postgresql.enabled controls whether to deploy this subchart.
+# When enabled, postgresql.auth.* is the single source of truth for credentials.
 postgresql:
   enabled: true
   auth:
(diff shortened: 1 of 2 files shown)


install/charts/dir/apiserver/values.yaml
```diff
@@ -661,23 +661,21 @@ envoy-authz:
       nginx.ingress.kubernetes.io/backend-protocol: "GRPC"
       nginx.ingress.kubernetes.io/grpc-backend: "true"
 
-# PostgreSQL subchart configuration (OPTIONAL)
-# Deploy PostgreSQL alongside the apiserver for multi-instance deployments.
-# NOT installed by default. Set enabled: true to deploy.
+# PostgreSQL subchart configuration
+# Deploy PostgreSQL alongside the apiserver.
+# Enabled by default when database.type is "postgres".
```
tkircsi: Where is this logic? I can't see it in the helpers or templates.

install/charts/dir/apiserver/values.yaml
```diff
@@ -661,23 +661,21 @@ envoy-authz:
...
 #
 # CREDENTIALS: When enabled, postgresql.auth.* is the single source of truth.
 # The apiserver will automatically use these credentials (no need to also set secrets.postgresAuth.*).
 postgresql:
-  enabled: false
+  enabled: true
```
tkircsi, line 671: I'm a bit confused with `database.type` and this settings. Can you explain a bit more details how the `database`, `posgresql` and `reconciler` are relates to each other and the `apiserver`?

install/charts/dir/apiserver/values.yaml
paralta: you are right, this comment is misleading! Let me update this

install/charts/dir/apiserver/values.yaml
paralta: will also add documentation on this to clarify 👍

paralta: @tkircsi please review the last commit. I think in the long-term the SQLite support can be dropped, which would remove the need for `database.type` and simplify configuration to just `postgresql.enabled` (deploy subchart vs use external).

PR #931, opened 2026-02-10
chore(dir): add hardcoded expected cid for e2e test

arpad-csepi:


e2e/shared/utils/cid.go
```diff
@@ -110,6 +112,20 @@ func CalculateCIDFromData(data []byte) string {
 	return cid
 }
 
+// ValidateCIDAgainstExpectedCID validates that a CID string matches a hardcoded expected CID value from expected_cid_list.json.
+func ValidateCIDAgainstExpectedCID(cidString string, recordData []byte, expectedCIDs []byte) {
```
ramizpolic: do we need the MD5?

can we just compare like this:

```
ValidateRecordCID(recordFile, expectedCID string)
```

then internally, just load the record file, calculate the CID using our helper functions that we already have, and compare CIDs

e2e/shared/testdata/expected_cid_list.json
```diff
@@ -0,0 +1,30 @@
+{
+    "c32098a7f0bd1e5ee8514354cb1a92a4": {
+        "expected_cid": "baeareih47nx2vjmqxx5xlar766ufi7jbxyedv634kcgagqrmnt42xjor3u",
+        "file": "record_070.json"
+    },
+    "dc0f821ae4a4674874f6f9f30a89159b": {
+        "expected_cid": "baeareign3udxxjh3zvclci2qdkusdgs7qm5ukdnv5nzlxakfttuivria4u",
+        "file": "record_070_sync_v4.json"
+    },
```
ramizpolic:
```
{
	"file": "expected-cid"
	"record_070_sync_v4.json": "baeareign3udxxjh3zvclci2qdkusdgs7qm5ukdnv5nzlxakfttuivria4u"
}
```

e2e/shared/utils/cid.go
arpad-csepi: To store a bunch of CIDs as a list without any more information in the file makes it harder to fix/update the values in it. That's why I added the MD5 and the filename as well. The MD5 is for validating that the file content unchanged and filename is just for to us for easier maintenance of the JSON file.

We need to store the expectedCID somewhere to avoid relying the CID calculation helper function to compute the CID on the fly. This way we avoid the tests to pass when a CID calculation is changed.

We already do the CID computation and compare using the recordFile with `utils.ValidateCIDAgainstData()`.

In the test case we only do have the recordFile as the record content itself and the CID for that. We can not use any file name or similar as a key in the expected json. (We can use the record name itself if we want to but I did not find that a good option as well)

arpad-csepi: We do not have to use MD5 specifically, but a hash on the file content as a key looks a good solution to link the exact record content and expected CID together

e2e/shared/testdata/expected_cid_list.json
arpad-csepi: We using the record files as embedded data so we do not have the file name as an information in the test case, so can not use it as a key.

e2e/shared/utils/cid.go
ramizpolic: IMO this is a bit error prone. in this implementation, hashes are not computed over cannonical data, so if we change certain order or apply formatting over JSONs (without changing the values), the hash will be different, while CID the same, so we still need to update this file.

with filename -> cid, we dont have this problem.

my suggestion here is: lets just test the CID calculation for one valid `v1.0.0` record and ensure that push for that record returns `baeareiabbog2umgduqhlcb64fzt6adn34kblzvru3fdzkl75hjhwt6h3da` CID. we do not need to test all the files. this makes it much simpler and we can cleanup current implementation.

e2e/client/01_client_test.go
ramizpolic, line 87:
```suggestion
testVersions := []struct {
name                string
jsonData            []byte
expectedSkillLabels []string
expectedDomainLabel string
expectedModuleLabel string
expectedCID string
```

ramizpolic: we add here expected CID

e2e/client/01_client_test.go
```diff
@@ -141,6 +141,9 @@ var _ = ginkgo.Describe("Running client end-to-end tests using a local single no
 
 				// Validate that the returned CID correctly represents the pushed data using canonical marshaling
 				utils.ValidateCIDAgainstData(recordRef.GetCid(), canonicalData)
+
+				// Validate the returned CID is in the hardcoded expected CIDs from expected_cid_list.json
+				utils.ValidateCIDAgainstExpectedCID(recordRef.GetCid(), version.jsonData, testdata.ExpectedCIDs)
```
ramizpolic:
```suggestion
				// Validate the returned CID is in the same as the CID from testcase
				gomega.Expect(recordRef.GetCid()).To(gomega.Equal(version.ExpectedCID))
```

ramizpolic: those are the only 2 changes we need (alongside populating the CIDs for testcases). we do not need the utils, embeds, or the expect file

ramizpolic (approved):

PR #932, opened 2026-02-10
fix(gui): improve verification details parsing and add verification result widget

muscariello:
This PR updates the GUI to correctly parse ProtoJSON verification details (Type -> Oidc structure) and adds a new visual widget to the chat stream that displays verification results immediately after checking.


mcp/tools/verify_record.go
ramizpolic, line 1: You can autofix these issues with `task lint:go FIX=true`

PR #944, opened 2026-02-12
fix(dir/zot): registry url handling and validation

paralta:
Fixes sync from a remote Zot registry over HTTPS so credentials are found and authentication succeeds.

## Problem

When syncing from a remote Zot registry over HTTPS, Zot reports **"failed to list repositories: unauthorized"**. Two causes:

1. **URL scheme:** Sync config used `http://` when no scheme was present. External registries are typically served over **HTTPS** (TLS on 443). Zot was connecting with the wrong scheme.
2. **Credential key:** Credentials must be keyed so Zot's lookup finds them. Zot uses the registry URL's authority (host or host:port, no scheme) as the key.

## Changes

- **config.go** – No URL normalization/validation; URL is validated on input in `RequestRegistryCredentials`.
- **credentials.go** – Derive credential key by stripping `http://`/`https://` from the URL (matches Zot's `url.Host`). Reject empty URL.
- **server/controller/sync.go** – `RequestRegistryCredentials`: when the configured registry address has no scheme, add `http://` when insecure (internal/E2E), `https://` otherwise; validate the final URL (parse and check scheme) before returning.

## Zot sync config (URLs vs credentials)

- **Sync config `urls`** ([config](https://github.com/project-zot/zot/blob/main/pkg/extensions/config/sync/config.go), [mirroring docs](https://zotregistry.dev/v2.0.4/articles/mirroring/#basic-configuration-for-mirroring-with-sync)): full URLs with scheme (e.g. `"https://registry1:5000"`) — used by the HTTP client.
- **Credentials file keys**: Key is the registry authority without scheme (e.g. `"registry2:5000"`). We strip the scheme from the URL to produce the same key.

### DIFF
--- a/server/controller/sync.go
+++ b/server/controller/sync.go
@@ -152,36 +152,15 @@ func (c *syncCtlr) RequestRegistryCredentials(_ context.Context, req *storev1.Re
 	ociConfig := c.opts.Config().Store.OCI
 	syncConfig := c.opts.Config().Sync
 
-	// Get registry address with default fallback
-	registryAddress := ociConfig.RegistryAddress
-	if registryAddress == "" {
-		registryAddress = ociconfig.DefaultRegistryAddress
+	registryAddress, err := ociConfig.GetRegistryAddress()
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "%v", err)
 	}
 
 	// Get repository name with default fallback
 	repositoryName := ociConfig.RepositoryName
 	if repositoryName == "" {
 		repositoryName = ociconfig.DefaultRepositoryName
-	}
-
-	// Add explicit scheme when none is present so Zot sync uses the correct protocol.
-	// Internal/insecure registries (e.g. E2E, k8s :5000) need http://; external use https://.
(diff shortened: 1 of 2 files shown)


server/controller/sync.go
```diff
@@ -164,6 +164,26 @@ func (c *syncCtlr) RequestRegistryCredentials(_ context.Context, req *storev1.Re
...
+
+	// Validate the final URL so zot utils receive a valid address.
+	parsed, err := url.Parse(registryAddress)
+	if err != nil {
+		return nil, status.Errorf(codes.InvalidArgument, "invalid registry address: %v", err)
+	}
+
+	if parsed.Scheme != "http" && parsed.Scheme != "https" {
+		return nil, status.Errorf(codes.InvalidArgument, "registry address must use http or https scheme")
+	}
```
ramizpolic: nit: consider moving this part to OCIConfig struct as a method `GetRegistryAddress() (string, error)`. the method could also take into account the default if not set

paralta: agreed 👍

PR #951, opened 2026-02-17
feat(server): enable regsync atomic sync processing

ramizpolic:
This PR implements a major refactoring of the synchronization architecture, moving from Zot-based dynamic sync configuration to atomic sync processing using `regsync`. The changes decouple sync operations from the apiserver and move them to the dedicated reconciler service.

### Architecture
- **Removed Zot-based sync**: Eliminated dynamic Zot sync configuration management, including `zot-config-pvc.yaml`, `zot-config-storageclass.yaml`, and related helpers. This simplifies the architecture and reduces issues/complexity in sync management caused by dynamic Zot configuration being shared across multiple components which limits scaling and impacts production deployment.
- **Reconciler-driven sync**: All sync operations now handled atomically by the reconciler service using `regsync` binary. This ensures more reliable and consistent sync processing, as the reconciler can manage sync state and retries without relying on external dynamic configuration or sharing state across components. This also enables support for any OCI registry that `regsync` supports, not just Zot.
- **Database-only sync state**: Dropped unnecessary fields (like `requires_regsync`) from the sync database table to ensure idempotency when managing syncs. The apiserver no longer performs credential negotiation as this is the responsibility of the reconciler.
- **Removed SQLite support**: Database is now PostgreSQL-only with adapter patter to support other server-side databases in the future. This simplifies database management and ensures better performance and reliability for sync operations, as SQLite was not well-suited for concurrent access patterns (server-side usage) in processing across services. This also enables better scaling capabilities for larger deployments.
- **Stateless deployment**: The services no longer share or require state, all state is loaded and managed by the external database or OCI registry.

### Helm Chart
- Simplified database configuration - moved from separate `database` section to `config.database`
- Added `authz_policies_csv` as separate top-level config (not part of server config struct)
- Reconciler now always deployed (removed `reconciler.enabled` flag)
- Reconciler configuration via `reconciler.config.yml` file instead of environment variables
- PostgreSQL host auto-detection improved with explicit user configuration priority
- Configs for apiserver and reconciler decoupled to only define actual go-based configuration parameters

### E2E Testing
- Tests now wait for `COMPLETED` status instead of `IN_PROGRESS` with arbitrary sleep (caused flakiness)
- Fixed flaky test and removed 60-second sleep delays in sync tests

### Local Stack (Docker)
- `server:start` task now uses `docker compose` to start full stack (apiserver, reconciler, zot, postgres)
- Added `reconciler.env` configuration file for local development
- Simplified local development workflow

## Breaking Changes

- Database configuration path changed from `database.*` to `config.database.*`
- Authorization policies moved from `config.authz.enforcer_policy_csv` to `authz_policies_csv`
- Reconciler configuration structure changed (now uses config file)
- SQLite database support removed (PostgreSQL only)

## Next steps

- [ ] (high) Adjust the testbed deployments with latest changes (from feature branch)
- [ ] (high) Ensure postgresql deployment is secure and working
- [ ] (high) Test sync logic across dev/prod testbed once deployed
- [ ] (high) Fix coverage-related issues when deploying a chart in the pipeline with enabled coverage. When enabled, the chart fails to install and times out for local and network test. https://github.com/agntcy/dir/pull/951/changes/58e3010c7791a68e17468c58833ff0426025fc51
- [ ] (low) Ensure reconciler deployment works with SPIRE as its needed in the future to get the identity in order to be able to interact via client (pullReferrer, etc) once other reconciliation logic from apiserver like naming verification is moved to it

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -544,7 +544,7 @@ tasks:
     vars:
       KIND_CLUSTER_NAME: '{{ .KIND_CLUSTER_NAME | default "agntcy-cluster" }}'
       KIND_CREATE_OPTS: '{{ .KIND_CREATE_OPTS | default "" }}'
-      E2E_COVERAGE_ENABLED: '{{ .E2E_COVERAGE_ENABLED | default "false" }}'
+      E2E_COVERAGE_ENABLED: "false"
       DEPLOY_BAKE_ENV: '{{ if eq .E2E_COVERAGE_ENABLED "true" }}{{ .COVERAGE_BAKE_ENV }}{{ else }}{{ .BAKE_ENV }}{{ end }}'
       DEPLOY_BAKE_GROUP: '{{ if eq .E2E_COVERAGE_ENABLED "true" }}coverage{{ else }}default{{ end }}'
     cmds:
@@ -618,7 +618,7 @@ tasks:
       HELM_CHART_PATH: "{{ .ROOT_DIR }}/install/charts/dir"
       HELM_VALUES_PATH: "{{ .ROOT_DIR }}/install/charts/dir/values.yaml"
       # Coverage config
-      E2E_COVERAGE_ENABLED: '{{ .E2E_COVERAGE_ENABLED | default "false" }}'
+      E2E_COVERAGE_ENABLED: "false"
       DEPLOY_IMAGE_TAG: '{{ if eq .E2E_COVERAGE_ENABLED "true" }}{{ .COVERAGE_IMAGE_TAG }}{{ else }}{{ .IMAGE_TAG }}{{ end }}'
     cmds:
       # TODO: make logic idempotent so that running functional tests does not change previous contexts
@@ -680,7 +680,7 @@ tasks:
       HELM_CHART_PATH: "{{ .ROOT_DIR }}/install/charts/dir"
       HELM_VALUES_PATH: "{{ .ROOT_DIR }}/install/charts/dir/values.yaml"
       # Coverage config
-      E2E_COVERAGE_ENABLED: '{{ .E2E_COVERAGE_ENABLED | default "false" }}'
+      E2E_COVERAGE_ENABLED: "false"
       DEPLOY_IMAGE_TAG: '{{ if eq .E2E_COVERAGE_ENABLED "true" }}{{ .COVERAGE_IMAGE_TAG }}{{ else }}{{ .IMAGE_TAG }}{{ end }}'
     cmds:
       # Validate required environment variables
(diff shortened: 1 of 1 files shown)


install/charts/dir/apiserver/templates/_helpers.tpl
ramizpolic, line 1: @tkircsi please review the helm chart if i missed anything to address these use-cases
- deployed with external psql database (values set through secrets/externalsecrets)
- deployed with chart psql database (values set through ssecrets/externalsecrets)
- deployed with chart psql database (values set through server/reconciler configs)
- deployed with chart psql database (values not set at all)

feel free to patch the PR with simplifications/improvements on this (in this or separate PR)

reconciler/tasks/regsync/task.go
```diff
@@ -90,220 +64,207 @@ func (t *Task) IsEnabled() bool {
...
-			"sync_id", t.currentSyncID, "error", err)
+	// Run the worker
+	workerErr := worker.Run(ctx)
 
-		return true
-	}
+	// Remove worker from active list after completion
+	t.mu.Lock()
+	delete(t.activeWorkers, syncID)
+	t.mu.Unlock()
```
tkircsi, line 241: Use the same defer pattern as in the processSyncDeletion or add recovery mechanism.
```suggestion
	var panicErr error
	defer func() {
		t.mu.Lock()
		delete(t.activeWorkers, syncID)
		t.mu.Unlock()
		if r := recover(); r != nil {
			panicErr = fmt.Errorf("worker panicked: %v", r)
			if err := t.db.UpdateSyncStatus(syncID, storev1.SyncStatus_SYNC_STATUS_FAILED); err != nil {
				logger.Error("Failed to update sync status to FAILED after panic", "sync_id", syncID, "error", err)
			}
		}
	}()

	// Run the worker
	workerErr := worker.Run(ctx)

	if panicErr != nil {
		return panicErr
	}
```

reconciler/tasks/regsync/task.go
```diff
@@ -90,220 +64,207 @@ func (t *Task) IsEnabled() bool {
...
-	// Run the command
-	err := cmd.Run()
+	logger.Info("Processing sync deletion", "sync_id", syncID)
 
-	// Log output regardless of success/failure
-	if stdout.Len() > 0 {
-		logger.Debug("regsync stdout", "sync_id", syncID, "output", stdout.String())
+	// Update status to DELETE_PENDING before starting worker
+	if err := t.db.UpdateSyncStatus(syncID, storev1.SyncStatus_SYNC_STATUS_DELETE_PENDING); err != nil {
+		return fmt.Errorf("failed to update sync status to DELETE_PENDING: %w", err)
```
tkircsi, line 192: Isnt't the status should be `SyncStatus_SYNC_STATUS_DELETE_PENDING` here? `processPendignDeletions` already fiktered by this state.

paralta (approved):
LGTM

Just a comment on processing deletions, previously we had periodic syncs which required deletion to stop. Now, we have syncs that run once and complete once done. The current deletion will not interrupt the sync process, so the sync deletion operation is now obsolete.

PR #953, opened 2026-02-18
fix(dir): coverage test run

arpad-csepi:
This PR is contains multiple fixes and chores:
- Remove broken CI workflow
- Enable coverage tests for E2E
- Fix apiserver envoy chart determinism at build time (removed from ignored)
- Updated helm version
- Implement reconciler coverage docker image
- Limit the built coverage image numbers on CI
- Fix CI lint job warning about go cache
- Add sub-level go modules to the codecov report
- Dynamically generate go module names for codecov unit test reports
- Fix E2E coverage extract waiting for pods

### DIFF
--- a/install/charts/dir/apiserver/templates/reconciler-deployment.yaml
+++ b/install/charts/dir/apiserver/templates/reconciler-deployment.yaml
@@ -57,6 +57,9 @@ spec:
                   name: {{ include "chart.fullname" . }}
                   key: oci-password
             {{- end }}
+            {{- with .Values.extraEnv }}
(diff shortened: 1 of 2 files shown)


paralta (commented):
Looking good overall 👍 The coverage is already looking better https://app.codecov.io/gh/agntcy/dir/tree/fix%2Fcoverage-e2e

The reconciler coverage image is now built, but we never extract coverage from reconciler pods or add reconciler to the Codecov upload. I think these steps are still missing, have a look at how this is done for the `apiserver`.

PR #955, opened 2026-02-19
chore(api): remove sync deletion across codebase

ramizpolic:
This PR removes the sync deletion across codebase since syncs can only be processed once (in a given time point) with final state being either completed/failed. The syncs are performed by syncing the snapshot from remote once they are processed.


ramizpolic:
Switching in favor of https://github.com/agntcy/dir/pull/956 because without it we'd loose the capabilities to invalidate previous sync deletions (and cleanup tasks)

PR #960, opened 2026-02-19
fix(dirctl): add record check if no signature found

arpad-csepi:
This PR adds a record check before verification of the signature to ensure a clearer error message and user experience.

### DIFF
--- a/client/verify.go
+++ b/client/verify.go
@@ -6,6 +6,7 @@ package client
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	corev1 "github.com/agntcy/dir/api/core/v1"
 	signv1 "github.com/agntcy/dir/api/sign/v1"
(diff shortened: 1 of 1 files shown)


client/verify.go
```diff
@@ -34,6 +34,17 @@ func (c *Client) Verify(ctx context.Context, req *signv1.VerifyRequest) (*signv1
 		return nil, fmt.Errorf("record CID is required")
 	}
 
+	_, err := c.Lookup(ctx, req.GetRecordRef())
```
paralta, line 38: a lookup failure is not necessarily a `record not found` error, maybe we can change the record message below or check the returned error code

arpad-csepi: Correct, I made a check for the lookup error message, please check if it is good

PR #976, opened 2026-02-23
refactor(dir/reconciler): migrate reverification to reconciler name task

paralta:
This PR moves DNS ownership verification from the standalone `server/reverification` service into a reconciler task, so scheduling and execution use the shared reconciler pool.

* Change signed record storage to keep an attached signature instead of the public key.
* Add reconciler name task (`reconciler/tasks/name`) that re-verifies DNS ownership of named records and persists results, reusing the previous verification logic. Why “name” task and not “reverification” task? A separate reconciler task for signature verification (caching verification state for signed/trusted filtering) is planned and this naming avoids conflicts.
* Remove the `server/reverification package` (scheduler, worker, service, config, types) and stop starting it from `server.Server`.


reconciler/tasks/name/task.go
```diff
@@ -0,0 +1,232 @@
...
+func (t *Task) Run(ctx context.Context) error {
+	logger.Debug("Running name verification")
+
+	records, err := t.db.GetRecordsNeedingVerification(t.config.GetTTL())
+	if err != nil {
+		return fmt.Errorf("get records needing name verification: %w", err)
+	}
+
+	if len(records) == 0 {
+		logger.Debug("No records need name verification")
```
arpad-csepi:
```suggestion
		logger.Info("No records need name verification")
```

server/naming/config.go
arpad-csepi, line 1: Is this needed? Looks duplicate with `naming/config/config.go`

paralta: nice catch! we should keep only the config inside the `config` folder

PR #982, opened 2026-02-24
feat(mcp): add name verification mcp tool

adamtagscherer:
This PR:
- adds a new MCP tool to verify a record's name
- refactors the MCP server to not parse the client config and create the client instance multiple times

Screenshot 2026-02-24 at 13 57 25

### DIFF
--- a/mcp/tools/verify_name.go
+++ b/mcp/tools/verify_name.go
@@ -1,0 +1,97 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package tools
+
+import (
+	"context"
+	"fmt"
+
+	namingv1 "github.com/agntcy/dir/api/naming/v1"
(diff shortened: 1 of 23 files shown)


mcp/tools/verify_name.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	// Call the appropriate verification method
+	var resp *namingv1.GetVerificationInfoResponse
+
+	var err error
+
+	if input.CID != "" {
+		// Verify by CID
+		resp, err = t.Client.GetVerificationInfo(ctx, input.CID)
```
paralta, line 57: no optional version like we have for `GetVerificationInfoByName`?

adamtagscherer: `GetVerificationInfo` works with a specific CID, which determines a record name + version so we can't talk about versions here.
For the `GetVerificationInfoByName` we can talk about different versions, because one record with the same name can have different versions.

PR #985, opened 2026-02-25
feat(dir/reconciler): add signature verification task

paralta:
Adds server-side caching of signature verification result. When `from_server` is set, clients get trusted status and signer info from the server (reconciler cache). Otherwise, verification stays client-side.

- **Database:** New `signature_verifications` table (per-signature status, signer info).
- **Reconciler:** New `signature` task that fetches records needing verification, uses the client’s verification logic, and writes results to the DB.
- **Indexer:** Extend indexing with signed data so that the signature task is able to pick up signed records for verification.
- **Server:** Controller and types extended to read/write signature verifications and build verify responses from cache; record list/search can filter by `trusted`.
- **Client**: `VerifyRequest` extended with `from_server`. When true, verify uses the server’s cached result; when false, verify is client-side.

### DIFF
--- a/sdk/dir-js/package-lock.json
+++ b/sdk/dir-js/package-lock.json
@@ -1274,9 +1274,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1291,9 +1288,6 @@
         "arm"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1308,9 +1302,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1325,9 +1316,6 @@
         "arm64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1342,9 +1330,6 @@
         "loong64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1359,9 +1344,6 @@
         "loong64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1376,9 +1358,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1393,9 +1372,6 @@
         "ppc64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1410,9 +1386,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1427,9 +1400,6 @@
         "riscv64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1444,9 +1414,6 @@
         "s390x"
       ],
       "dev": true,
-      "libc": [
-        "glibc"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
@@ -1459,9 +1426,6 @@
       "integrity": "sha512-3AHmtQq/ppNuUspKAlvA8HtLybkDflkMuLK4DPo77DfthRb71V84/c4MlWJXixZz4uruIH4uaa07IqoAkG64fg==",
       "cpu": [
         "x64"
-      ],
-      "libc": [
-        "glibc"
       ],
       "license": "MIT",
       "optional": true,
@@ -1477,9 +1441,6 @@
         "x64"
       ],
       "dev": true,
-      "libc": [
-        "musl"
-      ],
       "license": "MIT",
       "optional": true,
       "os": [
--- a/sdk/dir-js/test/client.test.ts
+++ b/sdk/dir-js/test/client.test.ts
@@ -437,6 +437,32 @@ describe('Client', () => {
           }
         }
 
+        // Response with from-server (cached) must match local verification
+        const fromServerResponse = await client.verify(
+          create(models.sign_v1.VerifyRequestSchema, {
+            recordRef: ref,
+            fromServer: true,
+          }),
+        );
+        expect(fromServerResponse.success).toBe(response.success);
+        expect(fromServerResponse.signers).toBeDefined();
+        expect(fromServerResponse.signers.length).toBe(response.signers.length);
+        for (let i = 0; i < response.signers.length; i++) {
+          const rSigner = response.signers[i];
+          const sSigner = fromServerResponse.signers[i];
+          expect(rSigner).toBeDefined();
+          expect(sSigner).toBeDefined();
+          expect(sSigner!.type.case).toBe(rSigner!.type.case);
+          if (rSigner!.type.case === 'key' && sSigner!.type.case === 'key') {
+            expect(sSigner!.type.value.publicKey).toBe(rSigner!.type.value.publicKey);
+            expect(sSigner!.type.value.algorithm).toBe(rSigner!.type.value.algorithm);
+          }
+          if (rSigner!.type.case === 'oidc' && sSigner!.type.case === 'oidc') {
+            expect(sSigner!.type.value.issuer).toBe(rSigner!.type.value.issuer);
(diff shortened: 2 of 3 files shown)


proto/agntcy/dir/sign/v1/sign_service.proto
```diff
@@ -144,6 +144,10 @@ message VerifyRequest {
 
   // Verification provider to use
   VerifyRequestProvider provider = 2;
+
+  // When true, use the server's cached verification result (from the reconciler).
+  // When false or unset, verification is performed locally.
+  bool use_server_verification = 3;
```
ramizpolic: nit: rename to remote, similarly to CLI flags with `--from-remote`

```suggestion
  // When true, use cached verification result from server.
  // When false, verification is performed locally.
  bool from_remote = 3;
```

cli/cmd/verify/verify.go
```diff
@@ -22,8 +22,9 @@ var Command = &cobra.Command{
 	Long: `This command verifies the record signature against
 identity-based OIDC or key-based signing process.
 
-Verification is performed client-side by fetching signatures from the
-directory and verifying them locally using Sigstore libraries.
+By default, verification is performed locally (fetch signatures from the
+directory and verify with Sigstore). Use --use-server-verification to
+use the server's cached result from the reconciler instead.
```
ramizpolic: lets not use the `reconciler` namings in the high-level docs. all results are coming from remote server, it is irrelevant for the user how this is computed, only that it is available

client/verify.go
```diff
@@ -102,15 +121,7 @@ func (c *Client) Verify(ctx context.Context, req *signv1.VerifyRequest) (*signv1
...
-			// VerifyWithAny accepts any valid signature.
-			// If a signature has no bundle, it must be verified with a key.
-			if len(sig.GetContentBundle()) == 0 {
-				signerInfo, verifyErr = cosign.VerifyWithKeys(ctx, payload, publicKeys, sig)
-			} else {
-				signerInfo, verifyErr = cosign.VerifyWithOIDC(payload, &signv1.VerifyWithOIDC{
-					Options: provider.Any.GetOidcOptions().GetDefaultOptions(),
-				}, sig)
-			}
+			signerInfo, verifyErr = VerifySignature(ctx, payload, publicKeys, sig)
```
ramizpolic: not sure if we get any benefits here by moving to a separate method, seems clear enough to keep as is

client/verify.go
```diff
@@ -226,3 +237,109 @@ func (c *Client) pullPublicKeys(ctx context.Context, recordRef *corev1.RecordRef
...
+	if reqValue == "" {
+		return true
+	}
+
+	if re, err := regexp.Compile(reqValue); err == nil {
+		return re.MatchString(signerValue)
+	}
+
+	return reqValue == signerValue
+}
```
ramizpolic: nit: consider moving this logic to `client/utils/matcher.go`

reconciler/tasks/indexer/task.go
```diff
@@ -223,6 +223,25 @@ func (t *Task) indexRecord(ctx context.Context, tag string) error {
...
+			return nil
+		})
+		if hasSignatures {
+			if err := t.db.SetRecordSigned(tag); err != nil {
+				logger.Warn("Failed to mark record as signed", "cid", tag, "error", err)
+			} else {
+				logger.Debug("Record marked as signed (has signature referrers in store)", "cid", tag)
+			}
+		}
+	}
```
ramizpolic, line 243: this part may not be super efficient because we would need to always walk all referrers for the sake of finding the right match. i think a bit more optimal flow would be the following:
- when a signature is pushed, create a new row for signatures in the db with `record_cid, referrer_cid, trusted=nil`
- in the reconciler verification, select all signatures from db that have `trusted=nil`
- verify the signatures individually and set proper state `trusted=true` or `trusted=false`

later, when you need to query all the records that are trusted, simply check if there is at least one trusted signature for that record in db rather than setting the state on the record itself. this also ensures that if a signature is deleted, a matching row will also be deleted, and a get trusted record query will always respond properly without having to recompute

reconciler/tasks/signature/task.go
```diff
@@ -0,0 +1,234 @@
...
+		}
+	}
+
+	logger.Info("Signature verification complete", "succeeded", succeeded, "failed", failed)
+
+	return nil
+}
+
+// verifyRecordSignatures verifies all signatures for one record and persists results plus record.trusted.
+func (t *Task) verifySignatures(ctx context.Context, recordCID string) error {
```
ramizpolic: same notes as above, we dont need to manage the `record.trusted` at all. this is inferred from the signature table in the DB

reconciler/README.md
```diff
@@ -40,3 +40,12 @@ The name task re-verifies DNS/name ownership of named records and caches results
...
 4. Stores the verification result (verified or failed) in the database for efficient API filtering
+
+### Signature Task
+
+The signature task verifies record signatures and caches results. It:
+
+1. Queries the database for signed records with no or expired verification (per TTL)
+2. For each record, collects signatures and public keys from the store
+3. Verifies each signature using the client's verification logic (key-based or OIDC)
+4. Upserts verification results to the database and sets the record’s trusted flag
```
ramizpolic:
```suggestion
1. Queries the database for signatures with missing verification state (per TTL)
2. For each signature (key-based or OIDC), verify it against the record
3. Upserts verification state to the database for the given signature
```

ramizpolic: note: to simplify things, we may need to directly add a public key inside the signature rather than having two separate referrers for signature and public key. this would allow directly verifying a signature against a record rather than having to collect all signatures and all public keys before doing verification. 

this can be done later as a separate task

sdk/dir-py/agntcy/dir_sdk/client/client.py
ramizpolic, line 1: why do we need to do all this extra work of doing filtration on the client side when we are issuing a request on the server?

since we already pass the full `VerifyRequest` to the server, the server itself should apply filtration rather than having to do this locally for each client.

server/controller/sign.go
```diff
@@ -1,41 +1,99 @@
...
 	signLogger.Debug("Sign request received - redirecting to client-side")
 
 	// Sign functionality is handled client-side
 	return nil, status.Error(codes.Unimplemented, "server-side signing not implemented - use client SDK")
 }
 
-//nolint:wrapcheck
-func (s *signCtrl) Verify(_ context.Context, _ *signv1.VerifyRequest) (*signv1.VerifyResponse, error) {
-	signLogger.Debug("Verify request received - redirecting to client-side")
+func (s *signCtrl) Verify(ctx context.Context, req *signv1.VerifyRequest) (*signv1.VerifyResponse, error) {
```
ramizpolic, line 38: the signature signer info filtration should happen here rather than on the client-side

proto/agntcy/dir/sign/v1/sign_service.proto
paralta: I think `from_server` would be clearer

client/verify.go
paralta: the goal was to avoid duplication since the reconciler will also use `VerifySignature`

client/verify.go
paralta: moved to server/controller/sign.go

sdk/dir-py/agntcy/dir_sdk/client/client.py
paralta: neat 👍

reconciler/tasks/signature/task.go
paralta: The problem with creating rows only on push is that synced records don’t go through push, so no row would be created for their signatures.

We are currently marking synced records as signed during indexing https://github.com/agntcy/dir/blob/feat/signature-reconciler/reconciler/tasks/indexer/task.go#L226. Afterwards, these records will be picked up by the signature task for verification (they are signed but have no recent verification) and that is when the entries are added to the signature db.

client/verify.go
ramizpolic: but we can already do this from the client directly?

```go
client.Verify(ctx, sign.VerifyRequest{})
```

when we do not use the `from_server` flag, it will perform it locally which is something that we need. i think this may be duplicating logic instead of reusing since we are doing the same thing -- collecting signatures and keys from remote and then doing the verification

paralta: just pushed a commit to move the verification flow to utils so we do not duplicate it in the reconciler and client. With this change I tried to avoid having ti instantiate the client on the rceonciler

server/controller/sign.go
```diff
@@ -16,26 +18,93 @@ var signLogger = logging.Logger("controller/sign")
...
+}
+
+func signerInfoFromVerification(obj types.SignatureVerificationObject) *signv1.SignerInfo {
+	switch obj.GetSignerType() {
+	case "oidc":
+		return &signv1.SignerInfo{
+			Type: &signv1.SignerInfo_Oidc{
+				Oidc: &signv1.SignerInfoOIDC{
+					Issuer:  obj.GetSignerIssuer(),
+					Subject: obj.GetSignerSubject(),
```
ramizpolic: we are missing certificate_issuer in this response and in the DB

paralta: nice catch! thanks for reviewing

PR #993, opened 2026-03-02
feat(dir): add DeleteReferrer

csirmazbendeguz:
add DeleteReferrer endpoint for deleting referrers


proto/agntcy/dir/core/v1/record.proto
```diff
@@ -88,4 +89,7 @@ message RecordReferrer {
 
     // The actual data of the referrer.
     google.protobuf.Struct data = 5;
+
+    // The CID of the referrer.
+    core.v1.ReferrerRef referrer_ref = 6;
```
paralta, line 89: Adding the `referrer_ref` field here creates a circular dependency: the CID is computed from the serialized `RecordReferrer`, but including the CID inside the object changes the serialization, which will generate a different CID.

For example, the `Record` message does not include `RecordRef` inside itself.

proto/agntcy/dir/store/v1/store_service.proto
```diff
@@ -103,3 +106,23 @@ message PullReferrerResponse {
...
+
+message DeleteReferrerRequest {
+  // The record the referrer(s) are referring to.
+  core.v1.RecordRef record = 1 [
+    (buf.validate.field).required = true
+  ];
+
+  // The CID of the referrer to delete.
+  // If set, delete the referrer by CID.
+  core.v1.ReferrerRef referrer_ref = 2;
```
ramizpolic:
```suggestion
  optional core.v1.ReferrerRef referrer_ref = 2;
```

proto/agntcy/dir/core/v1/record.proto
ramizpolic: referrer_ref is generated from the data field here, so it works. the RecordReferrer object is a wrapper around it, it is not the object that is serialized, only its data field is

paralta: please confirm but I checked this before commenting. we use the entire `RecordReferrer` not just the `data` field

paralta: https://github.com/agntcy/dir/blob/main/api/core/v1/referrer.go#L33

ramizpolic: lets update it then, it shouldnt require full object. this is the same as RecordMeta

PR #1000, opened 2026-03-03
fix(dir): bump postgres version in docker compose

csirmazbendeguz:
https://github.com/bitnami/containers/blob/main/bitnami/postgresql/18/debian-12/Dockerfile

I investigated a little bit, but I am not sure what causes this error. I'm pretty sure something is wrong with the release on Bitnami's side

It seems to work with the `latest` tag. It also works with the previous release

We can either bump it to `latest`, or roll it back to the previous release. I would say rolling back is the more prudent option

### DIFF
--- a/install/docker/docker-compose.yml
+++ b/install/docker/docker-compose.yml
@@ -54,7 +54,7 @@ services:
       start_period: 15s
 
   postgres:
-    image: docker.io/bitnami/postgresql@sha256:1ce1f97f3fd21366731ef29460bc210fc64355cbfc8ded68875892f75f026945
+    image: docker.io/bitnami/postgresql@sha256:e8b68936d05ee665974430bb4765fedcdc5c9ba399e133e120e2deed77f54dbf
     environment:
       POSTGRES_USER: postgres
       POSTGRES_PASSWORD: postgres
--- a/.github/workflows/container-security-scan.yml
+++ b/.github/workflows/container-security-scan.yml
@@ -72,7 +72,7 @@ jobs:
           - ghcr.io/project-zot/zot:v2.1.14
           - ghcr.io/spiffe/spire-server:1.14.1
           - ghcr.io/spiffe/spire-agent:1.14.1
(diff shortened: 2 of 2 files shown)


install/docker/docker-compose.yml
```diff
@@ -54,7 +54,7 @@ services:
       start_period: 15s
 
   postgres:
-    image: docker.io/bitnami/postgresql@sha256:1ce1f97f3fd21366731ef29460bc210fc64355cbfc8ded68875892f75f026945
+    image: docker.io/bitnami/postgresql@sha256:c47e54a69b39aa97d4405d68ea02bf2ffe06b646748e2ddf9c761416ead6acd5
```
paralta: have you tried using the latest image digest `sha256:e8b68936d05ee665974430bb4765fedcdc5c9ba399e133e120e2deed77f54dbf`? the postgres bump was intended to solve a security issue

csirmazbendeguz: yeah, it works with both the latest tag & latest digest too

my thought process is, since this is for local development only, it's better to go back. so developers can't use a feature locally that's not present in production...

but I'm just splitting hairs here - we can bump it if you want

paralta: I think what is important here is that we have matching postgres versions on both installation modes (helm and docker compose)

csirmazbendeguz: the postgres versions are matching 👍

paralta: https://github.com/agntcy/dir/blob/main/.github/workflows/container-security-scan.yml#L75 update the image on the security scan as well

arpad-csepi:
> I investigated a little bit, but I am not sure what causes this error. I'm pretty sure something is wrong with the release on Bitnami's side

Sorry for the bad bump because of the security stuff, the merged tag has a `0cd6fa81a` blob in it which could not be found somehow. Ran `trivy` on that tag, but that throw nothing about the bad image. Not sure either is this a bad release or what but the image clearly wrong. Thank you the fix! 🙏

PR #1003, opened 2026-03-04
feat(dir): add e2e tests for the referrer API

csirmazbendeguz:
We don't have any e2e tests for the referrer API, so we should add some.

I found a couple things that are not clear to me:

**Push Referrer endpoint:**
- Why allow setting `RecordRef` in both `PushReferrerRequest` and `RecordReferrer`? I believe we should either remove `record_ref` from `RecordReferrer` or add some validation at least so it's always the same as the `record_ref` in the `PushReferrerRequest`
- Why allow setting `created_at` from the client? If the purpose of this field is to track when the referrer was pushed, then it shouldn't be writeable by the client - ANSWER: this based on the [OASF Record](https://schema.oasf.outshift.com/1.0.0/objects/record) object, where there is a `created_at` field
- A lot of validation is missing, see the test cases

### DIFF
--- a/e2e/local/10_referrers_test.go
+++ b/e2e/local/10_referrers_test.go
@@ -1,0 +1,189 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package local
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/rsa"
+	"crypto/x509"
+	"encoding/pem"
+
+	corev1 "github.com/agntcy/dir/api/core/v1"
+	signv1 "github.com/agntcy/dir/api/sign/v1"
+	storev1 "github.com/agntcy/dir/api/store/v1"
+	"github.com/agntcy/dir/client"
+	"github.com/agntcy/dir/e2e/shared/testdata"
+	"github.com/google/uuid"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+func generatePublicKey() string {
+	key, err := rsa.GenerateKey(rand.Reader, 2048)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	pub, ok := key.Public().(*rsa.PublicKey)
+	gomega.Expect(ok).To(gomega.BeTrue())
+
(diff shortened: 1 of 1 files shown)


e2e/local/10_referrers_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+var _ = ginkgo.Describe("Running e2e tests for referrers", func() {
+	var (
+		c         *client.Client
+		ctx       context.Context
+		err       error
+		recordRef *corev1.RecordRef
+	)
+
+	publicKey := signv1.PublicKey{Key: testdata.PublicKey}
```
ramizpolic: lets generate a simple public key here using a go library

e2e/local/10_referrers_test.go
```diff
@@ -0,0 +1,189 @@
...
+		err = c.PushReferrer(ctx, &storev1.PushReferrerRequest{
+			RecordRef: record1,
+			Referrer:  referrer,
+		})
+
+		gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+		referrers := pullReferrers(c, ctx, record1, corev1.PublicKeyReferrerType)
+		gomega.Expect(referrers).To(gomega.HaveLen(1))
+		gomega.Expect(referrers[0].GetRecordRef()).To(gomega.BeNil())
```
paralta, line 115: I see the issue you mentioned before here. I would expect this to be filled with the `RecordRef` from `PushReferrerRequest`

e2e/local/10_referrers_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+		referrer := generateReferrer()
+		referrer.RecordRef = record2
+		err = c.PushReferrer(ctx, &storev1.PushReferrerRequest{
+			RecordRef: record1,
+			Referrer:  referrer,
+		})
+
+		// Should there be an error?
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
```
paralta, line 156: nice catch here and on the other checks below 👍  maybe we can fix these cases in the scope of this PR, or else do open an issue for them so we can address later

e2e/local/10_referrers_test.go
csirmazbendeguz: yep

e2e/local/10_referrers_test.go
csirmazbendeguz: I would like to merge this and then address it in follow-up PRs if that's ok?

PR #1014, opened 2026-03-05
feat(importer): add security scanner boilerplate

adamtagscherer:
## Scanner Architecture

The scanner pipeline stage follows a **plugin-based architecture** where each security scanner is an independent, self-registering component. The design separates two concerns:

1. **Orchestration** — channel wiring, result merging, fail-on-error/warning drop logic (`scanner/orchestrator.go`)
2. **Scanning** — per-record security analysis, isolated in its own subpackage (`scanner/behavioral/`, `scanner/static/`, `scanner/remote/`)

Adding a new scanner requires **two files in a new subpackage** and **one blank import** in the consumer — no changes to the orchestrator, factory, or pipeline.

The architecture also supports adding entirely different scanner tools (beyond `mcp-scanner`) as additional pipeline stages, independent of this registry.

```mermaid
graph TD
    subgraph "Pipeline"
        A[Fetcher] --> B[DuplicateChecker]
        B --> C[Transformer]
        C --> D[Orchestrator]
        D --> E[Pusher]
    end

    subgraph "scanner/"
        D -->|"per record"| F{Run all registered<br/>scanners}
        F --> G[Merge results]
        G --> H{FailOnError /<br/>FailOnWarning?}
        H -->|pass| E
        H -->|drop| I[IncrementFailedCount]
    end

    subgraph "scanner/behavioral/"
        J["init() → Register('behavioral', ...)"]
    end
    subgraph "scanner/static/"
        K["init() → Register('static', ...)"]
    end
    subgraph "scanner/remote/"
        L["init() → Register('remote', ...)"]
    end

    J -.->|self-register| F
    K -.->|self-register| F
    L -.->|self-register| F

### DIFF
--- a/importer/scanner/scanner.go
+++ b/importer/scanner/scanner.go
@@ -10,12 +10,11 @@ import (
 
 	corev1 "github.com/agntcy/dir/api/core/v1"
 	"github.com/agntcy/dir/importer/pipeline"
+	_ "github.com/agntcy/dir/importer/scanner/behavioral" // register behavioral scanner
 	scannerconfig "github.com/agntcy/dir/importer/scanner/config"
 	"github.com/agntcy/dir/importer/scanner/factory"
 	"github.com/agntcy/dir/importer/scanner/types"
 	"github.com/agntcy/dir/utils/logging"
-
-	_ "github.com/agntcy/dir/importer/scanner/behavioral" // register behavioral scanner
 )
 
 var logger = logging.Logger("importer/scanner")
@@ -57,9 +56,6 @@ func (s *Scanner) Scan(ctx context.Context, inputCh <-chan *corev1.Record, resul
 				}
 
 				recordName, _ := pipeline.ExtractNameVersion(record)
-				if recordName == "" {
-					recordName = "unknown@unknown"
-				}
 
 				scanResult, err := s.runAll(ctx, record, recordName)
 				if err != nil {

### CODE REVIEW

File: importer/mcpscanner/registry.go [hunk 1]
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// RunnerFactory creates a Runner from scanner config.
+type RunnerFactory func(cfg mcpscannerconfig.Config) Runner
+
+// registry maps scan mode names to their factory functions.
+// To add a new scan mode, implement the Runner interface and add one entry here.
+var registry = map[string]RunnerFactory{
+	"behavioral": func(cfg mcpscannerconfig.Config) Runner { return NewBehavioralRunner(cfg) },
+}
```
ramizpolic: this structure makes it a bit unclear how to register when we have multiple scanners available. consider enabling dynamic registration from implementations

```suggestion
// registry maps scan mode names to their factory functions.
// To add a new scan mode, implement the Runner interface and add one entry here.
var registry := map[string]RunnerFactory{}

func Register(name string, scannerFactory func(mcpscannerconfig.Config) Runner) {
	// if same name present, panic
	// if not, register it to registry
	// call this from behavioral/etc
}
```

importer/config/config.go
ramizpolic, line 1: im not sure i understand the filestructure here, but i would assume there are things like

```
scanner/
	factory.go - register factories, create scanners in one go
	runner.go - runs scanning logic based on whats registered
	behavioural/
		scanner.go
		config.go
	trivy/
		scanner.go
		config.go
types/
	scanner.go
```

where the scanners interface can be very simple

```
type Scanner interface {
	// Scan runs a scan for a single record and returns the result of scanning
	Scan(ctx context.Context, *corev1.Record) (*ScannerResult, error)
}
```

the wiring logic for running scanners and processing happens outside of the scanners as part of a scan runner

importer/mcp/importer.go
```diff
@@ -10,6 +10,10 @@ import (
 
 	"github.com/agntcy/dir/importer/config"
 	"github.com/agntcy/dir/importer/pipeline"
+	"github.com/agntcy/dir/importer/scanner"
+	_ "github.com/agntcy/dir/importer/scanner/behavioral" // register behavioral scanner
+	_ "github.com/agntcy/dir/importer/scanner/remote"     // register remote scanner
+	_ "github.com/agntcy/dir/importer/scanner/static"     // register static scanner
```
paralta: should this registration be done in the scanner package instead? this way the scanner pipeline/orchestrator would own its implementations the same way the importer pipeline does

importer/pipeline/pipeline.go
```diff
@@ -43,6 +43,13 @@ type DuplicateChecker interface {
 	FilterDuplicates(ctx context.Context, inputCh <-chan any, result *Result) <-chan any
 }
 
+// Scanner is a pipeline stage that runs security scans between transform and push.
```
paralta, line 46: transform can be a slow and expensive pipeline stage (using LLMs), should we run the scanner first and avoid transforming/enriching records that fail the security pipeline stage?

importer/mcp/importer.go
adamtagscherer: I refactored the code a bit because of cyclic dependency and put the registration in the scanner package.

importer/scanner/scanner.go
```diff
@@ -0,0 +1,214 @@
...
+			case <-ctx.Done():
+				return
+			case record, ok := <-inputCh:
+				if !ok {
+					return
+				}
+
+				recordName, _ := pipeline.ExtractNameVersion(record)
+				if recordName == "" {
+					recordName = "unknown@unknown"
```
paralta: should we set this here?

importer/pipeline/pipeline.go
adamtagscherer: The problem is that the scanner uses LLM too, so it's expensive as much as the transformer. Currently it's also a bit hard to put the scanner before the transformer, because before the transformer the data that we have has `any` type so it's hard to work with it. What we can do is I think to separate the transformer stage into 2 stages 1. transform MCP -> OASF record and 2. enrich OASF record so it's more composable. Also we could use the official Anthropic MCP Go struct instead of any, because every record is in this format when we fetch them. I don't want to make this PR bigger than it is, because it will be harder to review, but I've created an issue to do it in a follow up PR.

https://github.com/agntcy/dir/issues/1033

paralta: I understand now! thanks for explaining 👍

importer/scanner/scanner.go
adamtagscherer: I removed it.

PR #1015, opened 2026-03-05
refactor(dir): refactor the Dockerfile's build arguments

csirmazbendeguz:
for debugging purposes, we need to override the `go build`'s `-ldflags` (and we also need to provide `-gcflags`)

to support this, we need to refactor the build arguments

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -22,14 +22,15 @@ vars:
   REGSYNC_VERSION: "v0.11.1"
 
   ## Image config
+  BUILD_LDFLAGS: "-s -w -extldflags -static {{.VERSION_LDFLAGS}}"
   IMAGE_REPO: '{{ .IMAGE_REPO | default "ghcr.io/agntcy" }}'
   IMAGE_TAG: "{{ .IMAGE_TAG | default .COMMIT_SHA }}"
   IMAGE_BAKE_ENV: "IMAGE_REPO={{.IMAGE_REPO}} IMAGE_TAG={{.IMAGE_TAG}} REGSYNC_VERSION={{.REGSYNC_VERSION}}"
   IMAGE_BAKE_OPTS: '{{ .IMAGE_BAKE_OPTS | default "" }}'
-  BAKE_ENV: '{{ .IMAGE_BAKE_ENV }} EXTRA_LDFLAGS="{{.VERSION_LDFLAGS}}"'
+  BAKE_ENV: '{{ .IMAGE_BAKE_ENV }} BUILD_LDFLAGS="{{.BUILD_LDFLAGS}}"'
   COVERAGE_IMAGE_TAG: "{{ .IMAGE_TAG | default .COMMIT_SHA }}-coverage"
   COVERAGE_IMAGE_BAKE_ENV: "IMAGE_REPO={{.IMAGE_REPO}} IMAGE_TAG={{.COVERAGE_IMAGE_TAG}}"
-  COVERAGE_BAKE_ENV: '{{ .COVERAGE_IMAGE_BAKE_ENV }} EXTRA_LDFLAGS="{{.VERSION_LDFLAGS}}"'
+  COVERAGE_BAKE_ENV: '{{ .COVERAGE_IMAGE_BAKE_ENV }} BUILD_LDFLAGS="{{.BUILD_LDFLAGS}}"'
   COVERAGE_PKGS: '{{ .COVERAGE_PKGS | default "github.com/agntcy/dir/api/...,github.com/agntcy/dir/cli/...,github.com/agntcy/dir/client/...,github.com/agntcy/dir/importer/...,github.com/agntcy/dir/utils/..." }}'
 
   ## Dependency config
--- a/auth/cmd/envoy-authz/Dockerfile
+++ b/auth/cmd/envoy-authz/Dockerfile
@@ -23,14 +23,15 @@ RUN --mount=type=cache,target=/go/pkg/mod \
     xx-go mod download -x
 
 ARG BUILD_OPTS
-ARG EXTRA_LDFLAGS
+ARG BUILD_GCFLAGS
+ARG BUILD_LDFLAGS
 
 ENV CGO_ENABLED=0
 
 RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=bind,source=.,target=/build,ro \
-    xx-go build ${BUILD_OPTS} -ldflags="-s -w -extldflags -static ${EXTRA_LDFLAGS}" \
+    xx-go build ${BUILD_OPTS} -gcflags="${BUILD_GCFLAGS}" -ldflags="${BUILD_LDFLAGS}" \
     -o /bin/envoy-authz ./main.go
 
 RUN xx-verify /bin/envoy-authz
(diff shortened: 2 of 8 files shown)


auth/cmd/envoy-authz/Dockerfile
```diff
@@ -22,15 +22,25 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
+if [ -n "$GCFLAGS" ]; then BUILD_OPTS+=("-gcflags" "$GCFLAGS"); fi
+if [ "$COVER" == "true" ]; then BUILD_OPTS+=("-cover"); fi
+if [ -n "$COVERMODE" ]; then BUILD_OPTS+=("-covermode" "$COVERMODE"); fi
+EOF
+
 RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
     --mount=type=bind,source=.,target=/build,ro \
-    xx-go build ${BUILD_OPTS} -ldflags="-s -w -extldflags -static ${EXTRA_LDFLAGS}" \
+    xx-go build $BUILD_OPTS \
```
ramizpolic: it is already possible to pass build_opts and ldflags through env vars and taskfile. IMO we should avoid doing build option configuration as part of dockerfile and rather move this into taskfile. this does two things:
- keeps the things simple and modular/reusable by using taskfile to configure build options
- keeps the dockerfile simple (no additional changes needed as part of dockerfile, used as-is)


for example, the build option can be constructed inside the taskfile

```
# avoid this
RUN bash <<EOF
BUILD_OPTS=()
if [ -n "$LDFLAGS" ]; then BUILD_OPTS+=("-ldflags" "$LDFLAGS"); fi
if [ -n "$GCFLAGS" ]; then BUILD_OPTS+=("-gcflags" "$GCFLAGS"); fi
if [ "$COVER" == "true" ]; then BUILD_OPTS+=("-cover"); fi
if [ -n "$COVERMODE" ]; then BUILD_OPTS+=("-covermode" "$COVERMODE"); fi
EOF

# instead, configure through taskfile and pass as part of BUILD_OPTS and LDFLAGS
```

I think the only thing to do as part of dockerfile is to remove the current flags and only set like this

```
xx-go build ${BUILD_OPTS} -ldflags="${EXTRA_LDFLAGS}"
```

with a default of `-s -w -extldflags -static` being set inside taskfile

csirmazbendeguz: I understand where you're coming from @ramizpolic and this was my first thought as well!

I actually wanted to remove the `EXTRA_LDFLAGS` too and just have `BUILD_OPTS` for simplicity's sake ...

I'm not an expert on Bash programming, but as far as I understand this is not possible. The issue is, the shell can't parse arguments like this 😢 

You can try it in your terminal:

```sh
> BUILD_OPTS='-ldflags="-s -w -extldflags -static"'
> go build $BUILD_OPTS -o ./apiserver ./server/cmd/main.go

invalid value "\"-s -w -extldflags -static\"" for flag -ldflags: missing =<value> in <pattern>=<value>
```

The shell splits the arguments by space, so when `$BUILD_OPTS` receives something with a space in it, e.g. `foo="bar baz"`, the shell won't pass it properly to `go build`

---

Edit: another option would be to parse the `$BUILD_OPTS` into an array - I don't know if there's an easy way to do that

---

Edit 2: another option is to add feature flags instead of build options. For example:
- `COVERAGE=true` argument would set the `-cover covermode=atomic` build options in `-ldflags`
- `DEBUG=true` argument would set the `-gcflags` and override the default `-ldflags`

note: we would need to pass the `VERSION_LDFLAGS` for this too

tkircsi: @csirmazbendeguz  What about this?

```sh
BUILD_OPTS=(-ldflags "-s -w -extldflags -static")

go build "${BUILD_OPTS[@]}" -o ./apiserver ./server/cmd/main.go
```

csirmazbendeguz: Yes @tkircsi that's the goal - the question is how to pass the `BUILD_OPTS` to the Dockerfile. Because arguments are always passed as `string`, not `array`

csirmazbendeguz: We had a pair programming session with @ramizpolic , we decided on the following:

- parameterize `BUILD_GCFLAGS` & `BUILD_LDFLAGS`
- keep `BUILD_OPTS` for simple flags like coverage
- remove Shell array logic, if `BUILD_GCFLAGS` is empty just let it pass an empty string i.e. `-gcflags=""`

After this PR is merged, we need to:

- add a debug image that will include the debugger
- expose the dlv port on random port via docker compose
- add a debug flag to the Task file command(s) (but check if Task has native functionality for this first)

docker-bake.hcl
```diff
@@ -41,9 +40,7 @@ target "_common" {
     "linux/arm64",
     "linux/amd64",
   ]
-  args = {
-    EXTRA_LDFLAGS = "${EXTRA_LDFLAGS}"
-  }
+  args = {}
```
ramizpolic: nit: maybe better to move the values for build flags here rather than redefine them in dockerfile, e.g. for `BUILD_LDFLAGS=-s -w -extldflags -static`

docker-bake.hcl
```diff
@@ -6,7 +6,6 @@
 # Docker build args
 variable "IMAGE_REPO" { default = "ghcr.io/agntcy" }
 variable "IMAGE_TAG" { default = "v0.1.0-rc" }
-variable "EXTRA_LDFLAGS" { default = "" }
 variable "IMAGE_NAME_SUFFIX" { default = "" }
 variable "REGSYNC_VERSION" { default = "v0.11.1" }
```
ramizpolic, line 12: you can set defaults here for the buildflags and options

Taskfile.yml
```diff
@@ -22,14 +22,15 @@ vars:
...
   IMAGE_REPO: '{{ .IMAGE_REPO | default "ghcr.io/agntcy" }}'
   IMAGE_TAG: "{{ .IMAGE_TAG | default .COMMIT_SHA }}"
   IMAGE_BAKE_ENV: "IMAGE_REPO={{.IMAGE_REPO}} IMAGE_TAG={{.IMAGE_TAG}} REGSYNC_VERSION={{.REGSYNC_VERSION}}"
   IMAGE_BAKE_OPTS: '{{ .IMAGE_BAKE_OPTS | default "" }}'
-  BAKE_ENV: '{{ .IMAGE_BAKE_ENV }} EXTRA_LDFLAGS="{{.VERSION_LDFLAGS}}"'
+  BAKE_ENV: '{{ .IMAGE_BAKE_ENV }} BUILD_LDFLAGS="{{.BUILD_LDFLAGS}}"'
   COVERAGE_IMAGE_TAG: "{{ .IMAGE_TAG | default .COMMIT_SHA }}-coverage"
   COVERAGE_IMAGE_BAKE_ENV: "IMAGE_REPO={{.IMAGE_REPO}} IMAGE_TAG={{.COVERAGE_IMAGE_TAG}}"
-  COVERAGE_BAKE_ENV: '{{ .COVERAGE_IMAGE_BAKE_ENV }} EXTRA_LDFLAGS="{{.VERSION_LDFLAGS}}"'
+  COVERAGE_BAKE_ENV: '{{ .COVERAGE_IMAGE_BAKE_ENV }} BUILD_LDFLAGS="{{.BUILD_LDFLAGS}}"'
```
ramizpolic, line 33: i think the BUILD_LDFLAGS will be automatically active with bakenv without having to set them for BAKE_ENV and COVERAGE_BAKE_ENV, so they can be removed

csirmazbendeguz: the `{{.VERSION_LDFLAGS}}` is not in the bake file

PR #1016, opened 2026-03-05
test(e2e): move to tests and add mcp tests to ci

paralta:
This PR moves e2e tests under `tests/e2e`, adds MCP protocol e2e tests to CI, and cleans up unused testdata.

- Move e2e from `e2e/` to `tests/e2e/` and update all references (Taskfile, CI workflows, codecov, README, versions).
- Add MCP protocol end-to-end test task and run it in CI alongside local, network, and spire e2e jobs.
- Remove unused testdata file `record_warnings_080.json`.

**Note:** MCP e2e tests use a different deployment (docker compose via `install/docker`) than the other e2e suites (Kubernetes/Helm). This is desirable because it gates against regressions in the docker compose installation path, which was previously untested in CI.

### DIFF
--- a/Taskfile.vars.yml
+++ b/Taskfile.vars.yml
@@ -66,4 +66,4 @@ vars:
   GO_MOD_DIR:
     sh: find . -name go.mod -not -path "./tmp*" -exec dirname {} \;
   GO_MOD_DIR_UNIT_TEST:
-    sh: find . -name go.mod -not -path "./tests/e2e*" -not -path "./tmp*" -exec dirname {} \;
+    sh: find . -name go.mod -not -path "./tests*" -not -path "./tmp*" -exec dirname {} \;


tests/go.mod
ramizpolic, line 1: consider keeping this inside `tests/` as a single package with submodules for `e2e`, `integration`, etc. this ensures one package for all tests that are modular and can be executed like `dir/tests/main.go run --case e2e --case integration --case unit` rather than `dir/tests/e2e/main.go run` etc

ramizpolic: this is good if we want to run integration tests (or other kinds of tests) elsewhere so we can only expose one single package to run specific tests

paralta: agreed 👍 will make this a single package

Taskfile.yml
```diff
@@ -1408,16 +1409,28 @@ tasks:
...
+  test:e2e:mcp:
+    desc: Run MCP server protocol end-to-end tests (starts local Directory stack for push test)
+    aliases: [e2e:mcp]
+    env:
+      DIRECTORY_CLIENT_SERVER_ADDRESS: localhost:8888
+      OASF_API_VALIDATION_SCHEMA_URL: "{{ .OASF_API_VALIDATION_SCHEMA_URL | default \"https://schema.oasf.outshift.com\" }}"
+    cmds:
+      - defer: { task: server:stop }
+      - task: server:start
+      - bash -c 'echo "Waiting for Directory apiserver on localhost:8888..."; for ((i=0;i<60;i++)); do (nc -z localhost 8888 2>/dev/null || timeout 1 bash -c "echo >/dev/tcp/localhost/8888" 2>/dev/null) && break; sleep 2; done; (nc -z localhost 8888 2>/dev/null || timeout 1 bash -c "echo >/dev/tcp/localhost/8888" 2>/dev/null) || (echo "Apiserver did not become ready on port 8888"; exit 1); sleep 3'
```
ramizpolic: nit: this should be handled as part of setup/prepare/init/ready logic inside mcp golang e2e natively rather than doing it here.

PR #1017, opened 2026-03-06
fix(install/docker): compose health checks

paralta:
This PR fixes health checks for the Docker Compose stack so services are reliably ready before use and the zot image (distroless, no `curl`) is handled correctly.

- **Remove zot healthcheck from docker-compose** — The zot service uses a distroless image without `curl`, so the previous `curl`-based healthcheck is removed.
- **Add grpc-health-probe to server image** — Copy `grpc-health-probe` into both server Dockerfile stages so gRPC health checks can run in the final image.
- **Use `docker compose up --wait` in Taskfile** — Start the stack with `--wait` so Compose waits for defined healthchecks before the task continues.
- **Wait for zot readiness in Taskfile** — After `docker compose up`, add a loop that polls zot’s `/readyz` until it succeeds or times out.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -518,7 +518,7 @@ tasks:
     env:
       DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL: 1s
     cmds:
-      - docker compose up -d --build
+      - docker compose up -d --build --wait
 
   server:stop:
     desc: Stop local Directory server stack


install/docker/docker-compose.yml
```diff
@@ -47,11 +47,6 @@ services:
       replicas: 1
       restart_policy:
         condition: on-failure
-    healthcheck:
-      test: curl -f http://localhost:5000/v2/ || exit 1
-      interval: 5s
-      retries: 60
-      start_period: 15s
```
ramizpolic, line 54: why do we remove the healthcheck here? cant we just curl the readyz until its ready from the docker-compose natively rather than in taskfile?

this ensures that the stack is created only when services are ready. we dont need to then do curl from taskfile at all since `--wait` will wait for ZOT to become ready and then for DIR to become ready

Taskfile.yml
```diff
@@ -518,7 +518,15 @@ tasks:
...
-      - docker compose up -d --build
+      - docker compose up -d --build --wait
+      - |
+        echo "Waiting for zot to be healthy..."
+        for i in $(seq 1 60); do
+          curl -sf http://localhost:5555/readyz >/dev/null && break
+          [[ $i -eq 60 ]] && { echo "zot /readyz did not become ready"; exit 1; }
+          sleep 2
+        done
+        echo "zot is healthy."
```
ramizpolic: not needed, this is covered by healthcheck inside docker-compose and dependency check before starting apiserver

install/docker/docker-compose.yml
paralta: @ramizpolic thanks for reviewing! this health check was removed because zot is using a distroless image without `curl`, hence this removed health check was not working

ramizpolic: in that case, we can use apiserver healtchecks which will report false as long as zot is not ready

PR #1020, opened 2026-03-07
feat(dir): add Delve debugger

csirmazbendeguz:
add the Delve debugger to the `apiserver`, `reconciler`, `runtime` images

modify `task server:start`, so if the `DEBUG=1` environment variable is set, it builds the `apiserver` and `reconciler` with the debugger included. e.g.:

```sh
DEBUG=1 task server:start
```

the debugger ports are exposed at:
- apiserver: 2345
- reconciler: 2346

to use the debugger, start the server in debug mode as described above and use your favorite IDE to create a remote debugging session

for example, to debug the apiserver, connect to 0.0.0.0:2345

instructions for Goland:
https://www.jetbrains.com/help/go/attach-to-running-go-processes-with-debugger.html#step-3-start-the-debugging-process-on-your-local-computer

it would be good to document this somewhere other than this PR. I would also include screenshots. should I create a README.md for this? where should I put this info?

ALTERNATIVE SOLUTION:

could add a new command instead of the `DEBUG` environment variable:
```
task server:start:debug
```
or pass it as a flag (needs additional parsing logic so I would avoid this):
```
task server:start -- --debug
```

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -521,6 +521,14 @@ tasks:
     cmds:
       - docker compose up -d --build --wait
 
+  server:start:debug:
+    desc: Start local Directory server stack in debug mode
+    dir: ./install/docker
+    env:
+      DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL: 1s
+    cmds:
+      - docker compose -f docker-compose.yml -f docker-compose.debug.yml up -d --build --wait
+
   server:stop:
     desc: Stop local Directory server stack
     dir: ./install/docker
--- a/install/docker/docker-compose.debug.yml
+++ b/install/docker/docker-compose.debug.yml
@@ -1,0 +1,30 @@
+# A set of Compose overrides to run the services in debug mode.
+# It's intended for local development only.
+# Developers may remote debug using Delve by connecting to:
+# - port 2345 for apiserver
+# - port 2346 for reconciler
+services:
+  apiserver:
+    build:
+      target: debug
+      args:
+        BUILD_GCFLAGS: "all=-N -l"
+        BUILD_LDFLAGS: ""
+    ports:
+      - 2345:2345
+    security_opt:
+      - apparmor=unconfined
+    cap_add:
+      - SYS_PTRACE
+  reconciler:
+    build:
+      target: debug
+      args:
+        BUILD_GCFLAGS: "all=-N -l"
+        BUILD_LDFLAGS: ""
+    ports:
+      - 2346:2345
+    security_opt:
+      - apparmor=unconfined
+    cap_add:
+      - SYS_PTRACE
--- a/reconciler/Dockerfile
+++ b/reconciler/Dockerfile
@@ -54,6 +54,18 @@ RUN --mount=type=cache,target=/go/pkg/mod \
 
 RUN xx-verify /bin/regsync
 
+# Debug image - includes Delve for remote debugging. For local development only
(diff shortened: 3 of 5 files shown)


reconciler/Dockerfile
```diff
@@ -54,6 +54,17 @@ RUN --mount=type=cache,target=/go/pkg/mod \
...
+FROM builder AS debug
+
+WORKDIR /
+
+COPY --from=builder /bin/reconciler ./reconciler
+
+RUN xx-go install github.com/go-delve/delve/cmd/dlv@v1.26.0
+
+ENTRYPOINT ["/go/bin/dlv", "--listen=0.0.0.0:2345", "--headless=true", "--accept-multiclient=true", "--api-version=2", "exec", "--continue", "./reconciler"]
+
```
ramizpolic, line 68: this will not work for sync since regsync binary is missing. can you please check the runtime/coverage stage and update accordingly

install/docker/docker-compose.debug.yml
```diff
@@ -0,0 +1,25 @@
+services:
```
ramizpolic, line 6: consider adding a small comment on the top of this file about the details like "this is a debug compose that exposes delve debugger for apiserver at port `2345` and reconciler at `2346`

install/docker/docker-compose.debug.yml
```diff
@@ -0,0 +1,25 @@
...
+  apiserver:
+    build:
+      target: debug
+      args:
+        BUILD_GCFLAGS: "all=-N -l"
+        BUILD_LDFLAGS: ""
+    ports:
+      - 2345:2345
```
ramizpolic, line 14: can you confirm if the existing ports will also be exposed that are part of regular compose? ie. both 8000 and 2345 are exposed in debug mode? im not familiar with compose patching

Taskfile.yml
```diff
@@ -517,9 +517,15 @@ tasks:
     desc: Start local Directory server stack
     dir: ./install/docker
     env:
+      DEBUG: "{{ .DEBUG }}"
       DIRECTORY_SERVER_PUBLICATION_SCHEDULER_INTERVAL: 1s
     cmds:
-      - docker compose up -d --build --wait
+      - |
+        if [ "$DEBUG" -eq "1" ]; then
```
ramizpolic: nit: consider also `debug=true` or `debug=1` casess

install/docker/docker-compose.debug.yml
csirmazbendeguz: yes, see [Merging rules](https://docs.docker.com/compose/how-tos/multiple-compose-files/merge/#merging-rules)

> For the multi-value options ports, expose, external_links, dns, dns_search, and tmpfs, Compose concatenates both sets of values:
> 
> original service:
>
> ```
> services:
> myservice:
>   # ...
>   expose:
>     - "3000"
> ```
> local service:
>
> ```
> services:
> myservice:
>   # ...
>   expose:
>     - "4000"
>     - "5000"
> ```
>
> result:
> ```
> services:
> myservice:
>   # ...
>   expose:
>     - "3000"
>     - "4000"
>     - "5000"
> ```

Taskfile.yml
arpad-csepi: nit: might be better to have `server:start:debug` or similar task instead so it is more visible to have this option without lookup the docs, will not cause the environment variable problems to other application (potentially) and in this case we do not need to handle different cases.

csirmazbendeguz: wdyt about adding a new command instead i.e. `task server:start:debug`?

Issue #926, opened 2026-02-10
Revisit local deployment configuration


ramizpolic:
Currently, local deployment uses certain ports that may be used by other common processes and system services (e.g. port 5000). 

Revisit the local deployment configuration to avoid making assumptions around port availability. For example:
- docker deployment should use random ports and use docker networking for communication
- dev setup (`task server:start`) should use random ports

PR #1025, opened 2026-03-09
refactor(dir): split up taskfile tasks

arpad-csepi:
This PR split up the single main Taskfile.yml into multiple files according to the tasks scope and make some fixes.
The taskfile refactor will be continued with #926 to implement dynamic port allocation, remove configuration duplications.


bin/Taskfile.yml
```diff
@@ -0,0 +1,299 @@
...
+      - which controller-gen
+
+  deps:bin-dir:
+    desc: Create bin directory
+    run: once
+    cmd: mkdir -p {{.BIN_DIR}}
+    status:
+      - test -d {{.BIN_DIR}}
+
+  deps:dirctl-bin:
```
paralta:
```suggestion
  deps:dirctl:
```

paralta: to be inline with other deps

bin/Taskfile.yml
```diff
@@ -0,0 +1,299 @@
...
+        cmd: |
+          echo "Updating dependencies in {{.ITEM}}" && \
+          go -C {{.ITEM}} get -u ./... && \
+          go -C {{.ITEM}} mod tidy -go={{.GO_VERSION}}
+
+  deps:tidy:
+    desc: Ensure dependencies are up-to-date
+    cmds:
+      - for: { var: GO_MOD_DIR }
+        cmd: go -C {{.ITEM}} mod tidy -go={{.GO_VERSION}}
```
paralta: not sure is this should be in the `bin` folder scope. maybe it belongs in the root taskfile, WDYT?

e2e/Taskfile.test-env.yml
```diff
@@ -0,0 +1,556 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
```
paralta, line 1: should this be in the `e2e` scope? most of these are test environment deployments (e.g. `deploy:kubernetes:local`, etc)

bin/Taskfile.yml
arpad-csepi: I prefer to keep separate the `deps` because a lot of code and can be well defined category to be stored a dedicated `Taskfile.yml`. The `bin` directory was chosen as all of the tasks in there interacting with that directory. If `bin` is not an option then maybe a `Taskfile.deps.yml` in the repository root sufficient?

e2e/Taskfile.test-env.yml
arpad-csepi: Can be later the `deploy` stuff used for anything else later outside of `e2e`, so would keep it. If want to move it then I suggest again a separate `Taskfile.deploy.yml` for it to avoid big yaml files.

bin/Taskfile.yml
paralta: I am fine with keeping as is for now then 👍

PR #1027, opened 2026-03-09
feat(dir): add referrer CIDs

csirmazbendeguz:
similar to records, referrer should also have a CID

this PR adds referrer CIDs - when a referrer is pushed, a CID is generated similar to how it's generated for records. the generated CID is returned in the push response


proto/agntcy/dir/store/v1/store_service.proto
```diff
@@ -59,6 +59,9 @@ message PushReferrerResponse {
   
   // Optional error message if push failed
   optional string error_message = 2;
+
+  // The CID of the referrer
+  string cid = 3;
```
paralta:
```
message ReferrerRef {
  string cid = 1;
}
```

Although we don’t plan to use anything other than CIDs for referrers right now, using a `ReferrerRef` message would keep the referrer methods aligned with the record methods (e.g. `  rpc Push(stream core.v1.Record) returns (stream core.v1.RecordRef);`). WDYT?

csirmazbendeguz: yes ok let's be consistent

personally, I don't like `ReferrerRef`. it feels weird that we need to nest a single field. but I understand [protobuf doesn't support type aliases](https://github.com/protocolbuffers/protobuf/issues/3521) so we're stuck with this

csirmazbendeguz: I added a new `CID` message for this 👍  I would like to migrate `RecordRef` to this as well in the future: https://github.com/agntcy/dir/pull/1048

csirmazbendeguz: or let's add a `ReferrerRef` for consisentcy's sake and then we can refactor later? I'll add that then

server/types/store.go
```diff
@@ -38,7 +38,7 @@ type StoreAPI interface {
 // Used by: store.Controller, sync.Monitor, signing.Service.
 type ReferrerStoreAPI interface {
 	// PushReferrer pushes a referrer to content store
-	PushReferrer(context.Context, string, *corev1.RecordReferrer) error
+	PushReferrer(context.Context, string, *corev1.RecordReferrer) (string, error)
```
paralta:
```suggestion
	PushReferrer(context.Context, string, *corev1.RecordReferrer) (*corev1.ReferrerRef, error)
```
we should return the new type here

paralta (commented):
Small comment, looking good overall 👍

server/types/store.go
csirmazbendeguz: Actually, I don't like this pattern of passing around `protobuf` objects like this

I know that this is already a pattern in the project so I will change it for consistency for now...

But it's an anti-pattern IMO. The only `protobuf`-aware code should be the `controllers`:
- The controller layer parses, validates & assembles the `protobuf` objects
- The storage layer doesn't know how data is serialized (separation of concerns)

I didn't mention it as there are bigger issues, but now that you brought it up, maybe I should create an issue for this?

PR #1034, opened 2026-03-10
feat(dir): multiprovider OIDC-based external authorization module

tkircsi:
## Summary

Adds an OIDC-based ext-authz service with Casbin RBAC and an Envoy integration test setup.

## Features

- **OIDC ext-authz**: gRPC service that validates JWT payloads, applies Casbin policies, and enforces deny lists
- **Multiple issuers**: Zitadel and GitHub OIDC via Envoy jwt_authn
- **Header handling**: Strip `x-jwt-payload` on JWT routes, strip `Authorization` before backend
- **Dev path**: `/api/test` accepts `x-jwt-payload` for testing without real JWTs

## Changes

- New `auth/authzserver/` module (config, rbac, payload, oidc_server)
- Integration test setup: docker-compose, mock-directory, Envoy config
- Envoy config fixes for v1.31 compatibility


auth/authzserver/oidc_server.go
```diff
@@ -0,0 +1,171 @@
...
+	return s.allowResponse(principal, principalType), nil
+}
+
+func (s *OIDCAuthorizationServer) allowResponse(principal, principalType string) *authv3.CheckResponse {
+	headers := []*corev3.HeaderValueOption{}
+
+	if principal != "" {
+		headers = append(headers,
+			&corev3.HeaderValueOption{
+				Header: &corev3.HeaderValue{Key: "x-authorized-principal", Value: principal},
```
ramizpolic: nit: consider taking the header values as consts to make clear around used headers

```
const principalHeader = "x-authorized-principal"
...
```

auth/authzserver/oidc_server.go
```diff
@@ -0,0 +1,171 @@
...
+		roleResolver: roleResolver,
+		logger:       logger,
+	}, nil
+}
+
+// Check implements the ext_authz Check RPC.
+func (s *OIDCAuthorizationServer) Check(ctx context.Context, req *authv3.CheckRequest) (*authv3.CheckResponse, error) {
+	httpReq := req.GetAttributes().GetRequest().GetHttp()
+	path := httpReq.GetPath()
+
```
ramizpolic, line 69: we should probably drop any auth-related headers that we inject into the request somewhere to avoid users trying to bypass checks

auth/cmd/envoy-authz/test/config.test.yaml
```diff
@@ -0,0 +1,27 @@
...
+# Allows mock user for /api/* paths.
+
+claims:
+  userID: "sub"
+  emailPath: "email"
+
+issuers:
+  "https://tenant.zitadel.cloud":
+    principalType: "auto"
+    machineIdentityClaim: "client_id"
```
ramizpolic, line 11: similar note here on standardized config regarding issuers/providers config

auth/cmd/envoy-authz/config.yaml
```diff
@@ -0,0 +1,48 @@
...
+  userID: "sub"
+  emailPath: "email"
+
+# Issuer-specific extraction. Allowed principalType: auto | user | client | github
+issuers:
+  "https://tenant.zitadel.cloud":
+    principalType: "auto"
+    machineIdentityClaim: "client_id"
+  "https://token.actions.githubusercontent.com":
+    principalType: "github"
```
ramizpolic: provider config standardization

auth/authzserver/oidc_server.go
tkircsi: Nice catch! The headers now overwrite any client-supplied values instead of appending, so clients cannot override the trusted principal.

auth/authzserver/oidc_server.go
tkircsi: fix it

auth/cmd/envoy-authz/test/config.test.yaml
tkircsi: Thx! Refactored.

auth/cmd/envoy-authz/config.yaml
tkircsi: Fixed.

PR #1038, opened 2026-03-11
ci(dir): collect coverage and upload at once

arpad-csepi:
This PR refactor the codecov coverage action by:
- Use of OIDC instead of token
- Remove default options as not needed
- Remove fix coverage file name references
- Collect all coverage results after tests and upload it when all tests are passed.
- Simplify the coverage tasks in `Taskfile.yml`

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -191,13 +191,13 @@ jobs:
         include:
           - artifact_name: coverage-unit
             codecov_flag: unit
-            coverage_path: .coverage/unit
+            artifact_path: .coverage/unit
           - artifact_name: coverage-e2e-Local
             codecov_flag: e2e-Local
-            coverage_path: .coverage/e2e
+            artifact_path: .
           - artifact_name: coverage-e2e-Network
             codecov_flag: e2e-Network
-            coverage_path: .coverage/e2e
+            artifact_path: .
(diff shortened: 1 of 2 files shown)


Taskfile.yml
```diff
@@ -225,87 +225,72 @@ tasks:
...
+          {{end}}
+
+          path={{.ITEM}}
+          coverage_dir_path="${path%/*}"
+          mkdir -p {{.COVERAGE_DIR}}/unit/${coverage_dir_path}
+
+          echo "Running coverage tests in {{.ITEM}}"
+          go test -C {{.ITEM}} -covermode=atomic -coverprofile="{{.COVERAGE_DIR}}/unit/{{.ITEM}}.out" ./... > /dev/null
+
+  test:unit:coverage:local:
```
paralta, line 258: for e2e tests, we will not be able to see an equivalent report, right? we could potential remove this task but I am not against keeping it

arpad-csepi: The `test:unit:coverage:local` was part of the `test:unit:coverage` task before, but I moved out the filtering part which is needed only for local usage (as `codecov.yml` has a separate config for that). So `test:unit:coverage:local` is usable locally for unit tests coverage. The E2E tests needs a bit more work later to have inline with this kind of taskfile structure.

PR #1039, opened 2026-03-11
feat(dir/helm): add OIDC support for Envoy and ext-authz helm charts

tkircsi:
## Summary

Adds OIDC-based Envoy auth gateway support to the Helm charts, replacing the previous GitHub-specific auth path with a more flexible `jwt_authn + ext_authz + Casbin` setup.

### What changed

- Added `envoy-authz` chart support for OIDC/JWT authentication with Envoy `jwt_authn`
- Added multi-issuer support for:
  - Zitadel user/client tokens
  - GitHub Actions OIDC workload tokens
- Updated Envoy config to:
  - validate JWTs
  - forward verified payload via `x-jwt-payload`
  - strip `Authorization` before forwarding to the backend
- Updated authz config rendering to use config-defined principals and roles only
- Added support for issuer-specific principal extraction:
  - `user:{iss}:{sub}`
  - `client:{iss}:{client_id}`
  - GitHub workflow principals for Actions
- Wired the `envoy-authz` chart into the `apiserver` chart as an optional dependency
- Extended chart values to configure OIDC issuers, JWKS endpoints, public paths, and Casbin role mappings

### Why

This moves the auth gateway toward an OIDC-first design, improves separation between authentication and authorization, supports both human and workload identities, and removes dependence on GitHub-specific token authorization logic.


install/charts/dir/apiserver/values.yaml
```diff
@@ -585,63 +585,87 @@ envoyAuthz:
...
+          allowedMethods:
+            - "/agntcy.dir.store.v1.StoreService/Push"
+            - "/agntcy.dir.store.v1.StoreService/Pull"
+          users: []
+          clients:
+            - "client:https://tenant.zitadel.cloud:69234237810729234"
+        prod-deployer:
+          allowedMethods: ["*"]
+          githubWorkflows:
+            - "ghwf:repo:agntcy/dir:workflow:deploy.yml:ref:refs/heads/main:env:prod"
```
paralta: since the `deploy.yml` workflow does not exist yet, we could comment out this role for now

PR #1043, opened 2026-03-12
feat(importer): add behavioral scanner

adamtagscherer:
# Implement behavioral security scanner
Implements the behavioral scanner. For each imported record, the scanner:

- Extracts the source_code locator URL and optional mcp_data.repository.subfolder
- Clones the repo (git clone --depth=1) to a temp directory
- Runs mcp-scanner behavioral --raw on the source
- Parses the JSON output and maps findings to ScanResult (CRITICAL/HIGH → error, MEDIUM → warning, LOW → info)

Records are gracefully skipped when:

- No source_code locator exists
- URL is a placeholder (example.com)
- Repository is not cloneable (private/deleted)
- Also wires the scanner into the CI workflow (import-records.yaml):

This PR also fixes a go version error because the importer currently is failing:
https://github.com/agntcy/dir/actions/runs/22980792321

Successful CI run logs:
https://github.com/agntcy/dir/actions/runs/22997376095

### DIFF
--- a/cli/presenter/output_test.go
+++ b/cli/presenter/output_test.go
@@ -510,6 +510,8 @@ func TestAddOutputFlags(t *testing.T) {
 	flag := cmd.Flags().Lookup(outputFlagName)
 	if flag == nil {
 		t.Fatal(flagNotFoundMsg)
+
+		return
 	}
 
 	// Check short flag exists


.github/workflows/import-records.yaml
```diff
@@ -139,6 +149,10 @@ jobs:
           DIRECTORY_CLIENT_AUTH_MODE: "github"
           DIRECTORY_CLIENT_SERVER_ADDRESS: ${{ env.SERVER_ADDRESS }}
           DIRECTORY_CLIENT_GITHUB_TOKEN: ${{ secrets.DIRECTORY_MCP_BOT_PAT }}
+          MCP_SCANNER_LLM_API_KEY: ${{ secrets.AZURE_OPENAI_API_KEY }}
+          MCP_SCANNER_LLM_BASE_URL: ${{ secrets.AZURE_OPENAI_ENDPOINT }}
+          MCP_SCANNER_LLM_MODEL: "azure/gpt-4o"
+          MCP_SCANNER_LLM_API_VERSION: "2024-10-21"
```
paralta: would be neat to reuse the more generic vars above `AZURE_*`. the vars are already used for the importer enricher and we are using the same here

importer/scanner/behavioral/behavioral.go
```diff
@@ -25,10 +37,195 @@ func New(cfg scannerconfig.Config) *Scanner {
...
+	rawOutput, err := runMCPScanner(ctx, s.cfg.CLIPath, absDir)
+	if err != nil {
+		return nil, fmt.Errorf("mcp-scanner: %w", err)
+	}
+
+	return parseOutput(rawOutput)
+}
+
+// isPlaceholderURL returns true for URLs that are not real repositories
+// (e.g. example.com placeholders injected by the transformer).
```
paralta: where is this injection made? maybe we need to refrain from injecting these placeholders in the transformer stage

.github/workflows/import-records.yaml
adamtagscherer: These env vars are defined in the mcp scanner repository - https://github.com/cisco-ai-defense/mcp-scanner - I don't think we can rewrite it based on our needs.

importer/scanner/behavioral/behavioral.go
```diff
@@ -25,10 +37,195 @@ func New(cfg scannerconfig.Config) *Scanner {
...
+}
+
+func extractSourceCodeURL(fields map[string]*structpb.Value) string {
+	locatorsVal, ok := fields["locators"]
+	if !ok || locatorsVal == nil {
+		return ""
+	}
+
+	listVal := locatorsVal.GetListValue()
+	if listVal == nil {
```
ramizpolic: why are we still doing data extraction like this? we have our corev1 package that can take in a protobuf and return a record, lets use that to access data

importer/scanner/behavioral/behavioral.go
adamtagscherer: It's defined here: https://github.com/agntcy/oasf-sdk/blob/main/pkg/translator/mcp.go#L691

@akijakya should know more likely why we need that.

.github/workflows/import-records.yaml
paralta: you can pass the existing vars here https://github.com/agntcy/dir/blob/feat/add-behavioral-scanner/importer/scanner/behavioral/behavioral.go#L220C14-L220C28

importer/scanner/behavioral/behavioral.go
akijakya: Ah yes, I think this is a leftover from when locator was a required field in the record (up until OASF 0.7.0), so we had to put something there, but it should be removed now. I'll create an issue about it in the SDK's repo, but I think this placeholder check can be removed either way, because it is also gracefully handled if the repo doesn't exist/private, right?

adamtagscherer: Yes we can remove, we are just unnecessarily trying to clone a repo at an URL that doesn't exist, and most of the records have this placeholder URL. I'll remove it and when we removed it from the OASF-SDK the problem will cease to exist.

Issue #1036, opened 2026-03-11
Replace deprecated actions/upload-release-asset


arpad-csepi:
### Problem Statement

The `reusable-release.yaml` uses `[actions/upload-release-asset](https://github.com/actions/upload-release-asset)` action to release the binaries but that action is no longer maintained.

### Proposed Solution

Find an alternative action which is maintained like `[softprops/action-gh-release](https://github.com/softprops/action-gh-release)`

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

PR #1052, opened 2026-03-13
ci(workflow): migrate deprecated release actions to softprops/action-gh-release

kondekarshubham123:
This PR replaces the deprecated `actions/create-release` and `actions/upload-release-asset` with the actively maintained `softprops/action-gh-release` in `reusable-release.yaml`. 

### Key Changes:
- **Unified Action**: Migrated to `softprops/action-gh-release` which is currently maintained and widely used.
- **Improved Security**: Added `permissions: contents: write` to the workflow jobs.
- **Workflow Simplification**: 
  - The new action auto-discovers releases by `tag_name`, so explicit `upload_url` passing is no longer required.
  - Asset naming and MIME types are handled automatically by the new action.

### Validation:
- Verified workflow structure and syntax.
- Checked parameter mapping against official documentation.
- Verified in fork (core CI passed; `Buf CI` failure in fork is expected as it requires upstream secrets like `BUF_TOKEN`).

Ref: https://github.com/agntcy/dir/issues/1036

### DIFF
--- a/.github/workflows/import-records.yaml
+++ b/.github/workflows/import-records.yaml
@@ -23,6 +23,10 @@ on:
         description: "Force push even if record already exists"
         type: boolean
         default: false
+      scanner_enabled:
+        description: 'Run security scanner on each record'
+        type: boolean
+        default: true
       server_address:
         description: "Directory server address"
         type: string
@@ -41,6 +45,7 @@ jobs:
       server_address: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.server_address || 'prod.gateway.ads.outshift.io:443' }}
       sign: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.sign || 'true' }}
       force: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.force || 'false' }}
+      scanner_enabled: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.scanner_enabled || 'true' }}
     env:
       IMPORT_CONFIG: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.import_config || '.github/workflows/scripts/import-records.json' }}
     steps:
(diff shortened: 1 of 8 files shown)


.github/workflows/reusable-release.yaml
```diff
@@ -69,17 +69,19 @@ jobs:
...
+    permissions:
+      contents: write
     outputs:
-      upload_url: ${{ steps.create_release.outputs.upload_url }}
+      release_id: ${{ steps.create_release.outputs.id }}
     steps:
       - name: Create Release
         id: create_release
-        uses: actions/create-release@0cb9c9b65d5d1901c1f53e5e66eaf4afd303e70e # v1.1.4
+        uses: softprops/action-gh-release@c7b3105e195da124803737b836d5069a53f19379 # v2.4.0
```
ramizpolic: Can you bump to latest version `v2.5.0`?

kondekarshubham123: I have bumped the action to `v2.5.0`

.github/workflows/reusable-release.yaml
```diff
@@ -115,11 +119,10 @@ jobs:
       - name: Upload Release Asset
         if: ${{ !(contains(matrix.os, 'windows') && contains(matrix.arch, 'arm64')) }}
         id: upload-release-asset
-        uses: actions/upload-release-asset@e8f9f06c4b078e705bd2ea027f0926603fc9b4d5 # v1.0.2
+        uses: softprops/action-gh-release@a06a81a03ee405af7f2048a818ed3f03bbf83c7b # v2.5.0
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN  }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         with:
```
ramizpolic, line 123: i think we are missing here `draft: true`, otherwise after pushing the resources it will publish the release

.github/workflows/reusable-release.yaml
```diff
@@ -69,17 +69,19 @@ jobs:
       - chart
       - cli
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
     outputs:
-      upload_url: ${{ steps.create_release.outputs.upload_url }}
+      release_id: ${{ steps.create_release.outputs.id }}
```
ramizpolic: the outputs for this package is not needed, it uses named releases

.github/workflows/reusable-release.yaml
kondekarshubham123: I have added `draft: true` to the upload step to ensure the release strictly maintains its draft status during the asset upload phase.

.github/workflows/reusable-release.yaml
kondekarshubham123: I've removed the obsolete outputs from the `release` job since the new action doesn't require passing the URL.

PR #1055, opened 2026-03-13
ci(dir): extend security scanner to support more tag

arpad-csepi:
This changes are extending the security scanner daily runs to check security issues from the latest main commit as well as the latest released version.

### DIFF
--- a/.github/workflows/cron.yaml
+++ b/.github/workflows/cron.yaml
@@ -17,10 +17,16 @@ jobs:
       image_tag: nightly
       push: true
 
-  security:
-    name: Container Security Scan
+  nightly-scan:
+    name: Container Security Scan (nightly)
     uses: ./.github/workflows/container-security-scan.yml
     needs:
       - build-images
     with:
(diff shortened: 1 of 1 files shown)


.github/workflows/container-security-scan.yml
ramizpolic, line 1: We can already perform a custom build of images (e.g. `ghcr.io/agntcy/dir-apiserver-dev:feat-feat-add-zitadel-chart-253a20c`). Lets not reinvent here custom build process. I would suggest the following:
- the custom image build builds every night one tag `ghcr.io/agntcy/dir-apiserver-dev:nightly`
- the security scan can be executed against a nightly build by allowing to select `repo-suffix` and `tag`

Options for a security scan:
- 1. Scan latest (select latest , sets the repo-suffix and tag appropriately)
- 2. Scan nightly (select nighly, sets the repo-suffix and tag appropriately)
- 3. Scan custom (input custom repo-suffix and tag)

PR #1063, opened 2026-03-17
feat(cli/client):  implement OIDC support cli and client

tkircsi:


client/oidc.go
ramizpolic, line 1: nit: code structure/usage

In the client package, all methods are related to functionalities. To make it a bit explicit on method usage, I would consider exposing auth-flows in the following way (related to auth-methods). For example, OIDC flow can be exposed like:

```go
// client/oidc.go
var OIDC = &OIDCProvider{}

type OIDCProvider struct {}

func (OIDCProvider) RunPKCEFlow() {

}

// cli/login
func Run() {
	client.OIDC.RunPKCEFlow()
}
```

tkircsi: Thx! This is indeed much clearer.

PR #1067, opened 2026-03-18
refactor(dir): refactor PullReferrer

csirmazbendeguz:
add CID field to PullReferrer, so referrers can be pulled by CID


client/store.go
```diff
@@ -154,7 +154,7 @@ func (c *Client) PushReferrer(ctx context.Context, req *storev1.PushReferrerRequ
 }
 
 // PullReferrer retrieves all referrers using the PullReferrer RPC.
-func (c *Client) PullReferrer(ctx context.Context, req *storev1.PullReferrerRequest) (<-chan *storev1.PullReferrerResponse, error) {
+func (c *Client) PullReferrer(ctx context.Context, req *storev1.PullReferrerRequest) ([]*storev1.PullReferrerResponse, error) {
```
tkircsi, line 157: Could we follow the `client` methods pattern here? (x, xStream, xBatch)

tkircsi: Ok. I see, we request referrers for one record.

csirmazbendeguz: not sure what you mean by (x, xStream, xBatch) ?

PR #1093, opened 2026-03-23
feat(dir): add integration tests for server

### LINKED ISSUE #1031: [Epic]: Standardize integration tests
I wanted to touch base regarding the type of tests we have in the system and I wanted to go over some definitions. Please, feel free to correct me if any of this is incorrect

**The testing pyramid**

I'm sure you have heard about the testing pyramid before. If not, here is an example by [Martin Fowler](https://martinfowler.com/articles/practical-test-pyramid.html):

pyramid from testgrid.io

We can see 3 types of tests on this pyramid: **unit tests**, **service tests** and **UI tests**

**Unit tests**

A unit is a function, class or method. A unit test is a test that tests one unit. Unit tests are isolated. A unit test doesn't interact with units other than the one it's supposed to test

- 👍 Unit tests are suitable for testing complex functions and algorithms
- 👎 Unit tests are not suitable for testing integration between services, such as API - OCI - DB interactions
- ✅ I haven't seen any issues with the unit tests in this project

**Service tests**

A service test (integration test) tests the interaction between services. For example, API - OCI - DB interactions

- 👍 Integration tests are suitable for testing the API works with the OCI and the DB
- 👍 Integration tests are easier to isolate. The test calls the code **directly** (not via a K8S pod or a Docker Compose service). This makes it possible to isolate tests better
- NOTE: I have always done it this way, but it's ok to do integration testing via K8S or Docker Compose too, but isolation will be much harder at the OCI / DB level I think
- 👎 Integration tests do not replace unit tests and E2E tests. It's better to test algorithms with unit tests and it's better to test user workflows with E2E tests
- ❌  We don't have any service tests. We can't test the state of the OCI, database or related services

Image

The anatomy of an integration test:

```
1. make sure the dependencies are running (Zot, Postgres), config is correct
2. call the function (e.g. rpc PushReferrer)
3. make assertions about state changes (ORAS, gorm)
```

If "integration test" sounds confusing, we could call it "service test" or "component test"

**UI tests**

UI, in this context, is any interface where the user can interact. We may not have a UI, but we have `dirctl`. The name we use for this layer is **end-to-end** tests. Any test which tests user workflows at a high level can be considered an E2E test

- 👍 E2E tests are suitable for testing user interactions and workflows
- 👎 E2E tests are not suitable for testing details, such as side effects of API calls (e.g. what annotations are created in the OCI registry?)
- 👎 E2E tests cannot be easily [isolated](https://github.com/agntcy/dir/issues/1018)
- ✅ We have E2E tests

**Proposal**

While we have unit tests & E2E tests, we don't have any integration tests. I think we should consider adding them

I know @paralta is working on adding a new, more production-like test setup. We could call that setup E2E and repurpose the existing E2E tests for integration testing

---

Note: there's an example of an integration test here, which is great, but this should be done in a re-usable way in my opinion:

https://github.com/agntcy/dir/blob/9958c8dde563eafba48dd2c1f385e783a2af5c5a/server/store/oci/integration_test.go#L155

Image

csirmazbendeguz:
Add a suite of integration tests for testing the API server. The main benefits are:

- white-box testing
- better isolation than E2E tests:
  - separate OCI repositroy per test to avoid conflicts in Zot
  - tests are executed in a database transaction to avoid conflicts in Postgres
- parallel execution

For more info see `README.md`:

https://github.com/agntcy/dir/pull/1093/changes#diff-26b30a17d0427f4dac432205fb815e5303f32f5d61851792cb9d6d9adaef0d52R1

### DIFF
--- a/server/config/config.go
+++ b/server/config/config.go
@@ -288,17 +288,37 @@ func (c ConnectionConfig) WithDefaults() ConnectionConfig {
 	return c
 }
 
-//nolint:maintidx
-func LoadConfig() (*Config, error) {
+type ConfigOptions struct {
+	file string
+}
+
+type ConfigOption func(*ConfigOptions)
+
+func WithFile(file string) ConfigOption {
+	return func(opts *ConfigOptions) {
+		opts.file = file
+	}
+}
+
+func LoadConfig(opts ...ConfigOption) (*Config, error) {
+	var options ConfigOptions
+	for _, opt := range opts {
+		opt(&options)
+	}
+
 	v := viper.NewWithOptions(
 		viper.KeyDelimiter("."),
 		viper.EnvKeyReplacer(strings.NewReplacer(".", "_", "-", "_")),
 	)
 
-	v.SetConfigName(DefaultConfigName)
+	if options.file != "" {
+		v.SetConfigFile(options.file)
+	} else {
+		v.SetConfigName(DefaultConfigName)
+	}
+
 	v.SetConfigType(DefaultConfigType)
 	v.AddConfigPath(DefaultConfigPath)
-
 	v.SetEnvPrefix(DefaultEnvPrefix)
 	v.AllowEmptyEnv(true)
 	v.AutomaticEnv()
--- a/server/server.go
+++ b/server/server.go
@@ -154,9 +154,26 @@ func newOASFValidator(cfg *config.Config) (corev1.Validator, error) {
 	return v, nil
 }
 
+type ServerOptions struct {
+	database types.DatabaseAPI
+}
+
+type ServerOption func(*ServerOptions)
+
+func WithDatabase(database types.DatabaseAPI) ServerOption {
+	return func(opts *ServerOptions) {
+		opts.database = database
+	}
+}
+
 //nolint:cyclop // This function has been at the limit; refactoring is out of scope.
-func New(ctx context.Context, cfg *config.Config) (*Server, error) {
+func New(ctx context.Context, cfg *config.Config, opts ...ServerOption) (*Server, error) {
 	logger.Debug("Creating server with config", "config", cfg, "version", version.String())
+
+	var o ServerOptions
+	for _, opt := range opts {
+		opt(&o)
+	}
 
 	oasfValidator, err := newOASFValidator(cfg)
 	if err != nil {
@@ -230,9 +247,12 @@ func New(ctx context.Context, cfg *config.Config) (*Server, error) {
 		return nil, fmt.Errorf("failed to create routing: %w", err)
 	}
 
-	databaseAPI, err := database.New(cfg.Database)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create database API: %w", err)
+	databaseAPI := o.database
+	if databaseAPI == nil {
+		databaseAPI, err = database.New(cfg.Database)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create database API: %w", err)
+		}
 	}
 
 	// Create JWT authentication service if enabled
--- a/tests/integration/server/integration_server_suite_test.go
+++ b/tests/integration/server/integration_server_suite_test.go
@@ -1,0 +1,18 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"testing"
+
+	"github.com/agntcy/dir/tests/test_utils"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+)
+
+func TestIntegrationServer(t *testing.T) {
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	test_utils.InitGofakeit()
+	ginkgo.RunSpecs(t, "Integration Test Suite")
+}
--- a/tests/integration/server/setup.go
+++ b/tests/integration/server/setup.go
@@ -1,0 +1,213 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package server
+
+import (
+	"context"
+	"crypto/sha1" //nolint:gosec
+	"encoding/hex"
+	"flag"
+	"fmt"
+	"net"
+	"path/filepath"
+	"runtime"
+
+	"github.com/agntcy/dir/server"
+	"github.com/agntcy/dir/server/config"
+	"github.com/agntcy/dir/server/database" //nolint:typecheck
+	gormdb "github.com/agntcy/dir/server/database/gorm"
+	"github.com/agntcy/dir/server/events"
+	"github.com/agntcy/dir/server/store/oci"
+	"github.com/agntcy/dir/server/types"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/test/bufconn"
+	"gorm.io/gorm"
+	"oras.land/oras-go/v2/registry/remote"
+)
+
+const DefaultConfig = "./testenv/test-config.yaml"
+
+const (
+	bufSize = 1024 * 1024
+)
+
+type TestVars struct {
+	options    types.APIOptions
+	repository *remote.Repository
+	db         *gorm.DB
+	server     *server.Server
+	conn       *grpc.ClientConn
+}
+
+var (
+	conf *config.Config
+	t    TestVars
+)
+
+var _ = ginkgo.BeforeSuite(func() {
+	conf = loadConfig()
+})
+
+func init() {
+	flag.String("test-config", "", "Absolute path to test configuration file (optional)")
+}
+
+func loadConfig() *config.Config {
+	configFile := getConfigFile()
+	c, err := config.LoadConfig(config.WithFile(configFile))
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	return c
+}
+
+func getConfigFile() string {
+	f := flag.Lookup("test-config")
+	if f != nil && f.Value.String() != "" {
+		return f.Value.String()
+	} else {
+		_, filename, _, _ := runtime.Caller(0)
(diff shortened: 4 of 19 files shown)


server/server.go
```diff
@@ -153,7 +153,7 @@ func configureOASFValidation(cfg *config.Config) error {
 }
 
 //nolint:cyclop // This function has been at the limit; refactoring is out of scope.
-func New(ctx context.Context, cfg *config.Config) (*Server, error) {
+func New(ctx context.Context, cfg *config.Config, databaseAPI types.DatabaseAPI) (*Server, error) {
```
tkircsi: This is a public API break caused by test needs, not by a product requirement. Possibly most normal callers passing `nil`. Keep the existing public constructor stable, and add a separate dependency-aware constructor.

```go
type Dependencies struct {
	Database types.DatabaseAPI
}
func New(ctx context.Context, cfg *config.Config) (*Server, error) {
	return NewWithDependencies(ctx, cfg, Dependencies{})
}
func NewWithDependencies(
	ctx context.Context,
	cfg *config.Config,
	deps Dependencies,
) (*Server, error) {
	databaseAPI := deps.Database
...
```

tests/integration-server/setup.go
```diff
@@ -0,0 +1,179 @@
...
+
+	return db
+}
+
+// startServer starts a gRPC server in the background for the current spec.
+func startServer(ctx context.Context, options types.APIOptions, db *gorm.DB) (*grpc.Server, *grpc.ClientConn) {
+	databaseAPI, err := gormdb.New(db)
+	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+
+	s, err := server.New(ctx, options.Config(), databaseAPI)
```
tkircsi: The integration harness bypasses the server.Server lifecycle by extracting and stopping only the raw gRPC server. That is unsafe because server.New(...) eagerly constructs routing, and routing starts background goroutines/resources during construction. Those are only cleaned up by Server.Close() -> routing.Stop(), not by grpc.Server.Stop(). Please keep the full *server.Server in the test and shut it down via Close(ctx). Longer term, consider adding a listener-based start method so tests can use bufconn without bypassing lifecycle management.

tkircsi (dismissed):
Another comment, the new integration suite still depends on https://schema.oasf.outshift.com via DIRECTORY_SERVER_OASF_API_VALIDATION_SCHEMA_URL, and server.New() initializes the validator from that remote URL during startup. That means the suite is not fully hermetic and can fail due to external network/service issues rather than product regressions. Can we use a local schema fixture or local mock/schema server for this test layer?

csirmazbendeguz:
> Another comment, the new integration suite still depends on https://schema.oasf.outshift.com via DIRECTORY_SERVER_OASF_API_VALIDATION_SCHEMA_URL, and server.New() initializes the validator from that remote URL during startup. That means the suite is not fully hermetic and can fail due to external network/service issues rather than product regressions. Can we use a local schema fixture or local mock/schema server for this test layer?

Yes you're right... we should use a local OASF API server. I'll adjust. This is an issue in all tests unfortunately

Personally, I would rather have the OASF SDK do the validation without any server at all: https://github.com/agntcy/oasf-sdk/issues/140

server/server.go
csirmazbendeguz: Yeah, this is good! Thanks

Not sure I agree it's a public API break though, I have a different idea of what a public API is

tkircsi: Yeah, API means a lot. In a Go package API is what is the consumer can call. Not private.

tests/integration/server/integration_server_suite_test.go
ramizpolic, line 1: can we move all this to `tests/integration/server` from `tests/integration-server`? i would also ask to attach a similar up/down taskfiles and wire it into the `tests/Taskfile` test matrix and import routes for setting up/testing

tests/integration/server/integration_server_suite_test.go
```diff
@@ -0,0 +1,26 @@
...
+func TestIntegrationServer(t *testing.T) {
+	err := godotenv.Load(envFile)
+	if err != nil {
+		t.Errorf("failed to load %s: %s", envFile, err)
+	}
+
+	gomega.RegisterFailHandler(ginkgo.Fail)
+	test_utils.InitGofakeit()
+	ginkgo.RunSpecs(t, "E2E Production (production-like) Test Suite")
+}
```
ramizpolic: lets avoid this. instead i would reuse the config/suite that we have for e2e tests (similar approach by passing config flag (-test-config) with path to the config

csirmazbendeguz: yes... that's better than bringing in the godotenv library

csirmazbendeguz: but the configs would be different. it's not as simple as configuring a client here

csirmazbendeguz: @ramizpolic , I refactored it a little bit so the config can be read from the yaml now, let me know if that's ok for you

server/server.go
csirmazbendeguz: It has been fixed, I'm resolving this

tests/integration-server/setup.go
csirmazbendeguz: Yes, nice catch, thanks for this. I adjusted. I marking this as resolved

tests/integration/server/integration_server_suite_test.go
csirmazbendeguz: Yes, done

tests/integration/server/setup.go
```diff
@@ -0,0 +1,215 @@
...
+func getConfigFile() string {
+	f := flag.Lookup("test-config")
+	if f != nil && f.Value.String() != "" {
+		return f.Value.String()
+	} else {
+		_, filename, _, _ := runtime.Caller(0)
+		base := filepath.Dir(filename)
+
+		return filepath.Join(base, DefaultConfig)
+	}
```
ramizpolic, line 76:
```suggestion
	f := flag.Lookup("test-config")
	if f == nil {
		// ginkgo fail here or return an error
	}
	
	return f.Value.String()
```

server/config/config.go
```diff
@@ -288,17 +288,29 @@ func (c ConnectionConfig) WithDefaults() ConnectionConfig {
...
+	File string
+}
+
 func LoadConfig() (*Config, error) {
+	return LoadConfigWithDependencies(Dependencies{})
+}
+
+//nolint:maintidx
+func LoadConfigWithDependencies(deps Dependencies) (*Config, error) {
 	v := viper.NewWithOptions(
```
ramizpolic: in golang, the "override" pattern for passing dependencies to constructor is usually implemented by defining a common object and passing object mutators to the ctor, for example:

```go

// dependencies to be passed to specific ctors
// private fields to force usage of mutators to callers
type ConfigOptions {
	filePath string // private fields to force usage of functions
	db *db.DB
}

// type declaration for mutator functions
type ConfigOption func(*ConfigOptions)

// common mutators (implemented as function generators)
func WithConfigFile(cfgPath string) ConfigOption {
	return func(opts *ConfigOptions) {
		// set filepath
		opts.filePath = cfgPath
	}
}

func WithSomeRandomDB(db *db.DB) ConfigOption {
	return func(opts *ConfigOptions) {
		// set filepath
		opts.db = db
	}
}

// CTOR receives an optional list of mutators
func LoadConfig(opts ...ConfigOption) (*Config, error) {
	// create common object which holds overrides (dependencies for the ctor)
	var options ConfigOptions

	// apply all received mutators on the 
	for _, opt := range opts {
		opt(&options)
	}

	// get a specific dependency
	options.filePath
	options.db
}

// Usage
func main() {
	// no mutators, default behavior, no breaking code changes
	config, err := LoadConfig()

	// load config from a custom file
	config, err := LoadConfig(WithConfigFile("/path/to/my-override-file.yaml"))

	// load config external dependencies (just an example, config doesnt need db)
	config, err := LoadConfig(
		WithConfigFile("my-file.com"),
		WithSomeRandomDB(db),
	)
}
```

ramizpolic: this can be used for both the server config and database configurability

server/server.go
```diff
@@ -154,8 +154,16 @@ func newOASFValidator(cfg *config.Config) (corev1.Validator, error) {
...
+type Dependencies struct {
+	Database types.DatabaseAPI
+}
+
 func New(ctx context.Context, cfg *config.Config) (*Server, error) {
+	return NewWithDependencies(ctx, cfg, Dependencies{})
+}
+
+//nolint:cyclop // This function has been at the limit; refactoring is out of scope.
+func NewWithDependencies(ctx context.Context, cfg *config.Config, deps Dependencies) (*Server, error) {
```
ramizpolic: same comment as above

tests/integration/server/setup.go
```diff
@@ -0,0 +1,215 @@
...
+	conf *config.Config
+	t    TestVars
+)
+
+var _ = ginkgo.BeforeSuite(func() {
+	conf = loadConfig()
+})
+
+func init() {
+	flag.String("test-config", "", "Absolute path to test configuration file (optional)")
```
ramizpolic, line 56: we should probably modularize the current `e2e/shared/config` methods/objects to be usable across all tests (both integrations and e2e) in a separate PR

tests/integration/server/setup.go
ramizpolic: i think it should fail here because test config is required

csirmazbendeguz: The reason I would like this to work without a `-test-config` flag is that we need to be able to execute tests locally from the IDE as well for debugging purposes

Yes, we could set the flag from the IDE as well, but I feel like this is simpler this way

server/config/config.go
csirmazbendeguz: I see, okay, we could do it this way!

csirmazbendeguz: done

server/server.go
csirmazbendeguz: done

tests/integration/server/setup.go
csirmazbendeguz: I think it's fine to have this single line of duplication, there's nothing else in `e2e/shared/config` that we can re-use here

ramizpolic (approved):
LGTM

PR #1094, opened 2026-03-23
test(dir): add e2e test with ghcr storage

paralta:
Add end-to-end production tests for the GHCR OCI backend and the CI workflow to run them.

* Add GHCR E2E tests covering push, pull, lookup, cosign sign/verify, search (by name, skill, schema version), and delete
* Add reusable CI workflow that runs the tests and cleans up GHCR packages, gated to main and release tags
* Update tests/Taskfile.yml to handle authentication before running production E2E tests

**Example run:** https://github.com/agntcy/dir/actions/runs/23440612216/job/68191817738

### DIFF
--- a/.github/workflows/reusable-test-e2e-production.yaml
+++ b/.github/workflows/reusable-test-e2e-production.yaml
@@ -16,9 +16,8 @@ jobs:
         with:
           fetch-depth: 0
 
-      - name: Setup Taskfile
-        shell: bash
-        run: sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b ~/.local/bin
+      - name: Install Task
+        uses: go-task/setup-task@3be4020d41929789a01026e0e427a4321ce0ad44 #v2.0.0
 
       - name: Setup Cosign
         uses: sigstore/cosign-installer@ba7bc0a3fef59531c69a25acd34668d6d3fe6f22 # v4.1.0


.github/workflows/reusable-test-e2e-production.yaml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Setup Taskfile
+        shell: bash
+        run: sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b ~/.local/bin
```
arpad-csepi: Taskfile finally has official reusable action :)
```suggestion
      - name: Install Task
        uses: go-task/setup-task@3be4020d41929789a01026e0e427a4321ce0ad44 #v2.0.0
```

tests/Taskfile.yml
```diff
@@ -18,7 +18,18 @@ tasks:
...
     cmds:
+      - |
+        if [ -n "${DIRECTORY_CLIENT_GITHUB_TOKEN:-}" ]; then
+          echo "DIRECTORY_CLIENT_GITHUB_TOKEN is set, skipping interactive login."
+        elif {{ .DIRCTL_BIN }} auth status 2>&1 | grep -q "Token: Valid"; then
+          echo "Already authenticated (cached token valid)."
+        else
+          echo "Not authenticated, starting login..."
+          {{ .DIRCTL_BIN }} auth login
+        fi
```
arpad-csepi, line 32: Just a nit, but taskfile added if directive which allows to template task steps out, so when the task is executed the output shows less bash script :)
```suggestion
      - cmd: echo "DIRECTORY_CLIENT_GITHUB_TOKEN is set, skipping interactive login."
        if: '[ -n "${DIRECTORY_CLIENT_GITHUB_TOKEN:-}" ]'

      - cmd: |
          if {{ .DIRCTL_BIN }} auth status 2>&1 | grep -q "Token: Valid"; then
            echo "Already authenticated (cached token valid)."
          else
            echo "Not authenticated, starting interactive login..."
            {{ .DIRCTL_BIN }} auth login
          fi
        if: '[ -z "${DIRECTORY_CLIENT_GITHUB_TOKEN:-}" ]'
```

arpad-csepi (approved):

tests/Taskfile.yml
paralta: this PR needs to be updated to leverage the new authentication, I will wait until the test environment is updated

PR #1168, opened 2026-03-24
refactor(dir/ci): add reusable build-dirctl and GitHub OIDC token actions

tkircsi:
- Add a new local composite action: `.github/actions/build-dirctl`
- Add a new local composite action: `.github/actions/fetch-oidc-token`
- Refactor `.github/workflows/oidc-test.yml` to use these reusable actions

### DIFF
--- a/.github/actions/build-dirctl/action.yaml
+++ b/.github/actions/build-dirctl/action.yaml
@@ -29,8 +29,8 @@ runs:
         go-version: ${{ inputs.go_version }}
         cache-dependency-path: "**/*.sum"
 
-     - name: Install Task
-        uses: go-task/setup-task@3be4020d41929789a01026e0e427a4321ce0ad44 #v2.0.0
+    - name: Install Task
+      uses: go-task/setup-task@3be4020d41929789a01026e0e427a4321ce0ad44 #v2.0.0
 
     - name: Build dirctl
       id: build


.github/actions/build-dirctl/action.yaml
```diff
@@ -0,0 +1,49 @@
...
+  steps:
+    - name: Setup Go
+      uses: actions/setup-go@4b73464bb391d4059bd26b0524d20df3927bd417 # v6.3.0
+      with:
+        go-version: ${{ inputs.go_version }}
+        cache-dependency-path: "**/*.sum"
+
+    - name: Setup Taskfile
+      shell: bash
+      run: sh -c "$(curl --location https://taskfile.dev/install.sh)" -- -d -b ~/.local/bin
```
arpad-csepi: Taskfile finally has an official reusable action :)
```suggestion
     - name: Install Task
        uses: go-task/setup-task@3be4020d41929789a01026e0e427a4321ce0ad44 #v2.0.0
```

.github/actions/fetch-oidc-token/action.yaml
```diff
@@ -0,0 +1,35 @@
...
+  token:
+    description: The fetched OIDC token
+    value: ${{ steps.fetch.outputs.token }}
+
+runs:
+  using: "composite"
+  steps:
+    - name: Fetch OIDC token
+      id: fetch
+      uses: actions/github-script@v7
```
arpad-csepi:
```suggestion
      uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8.0.0
```

.github/actions/build-dirctl/action.yaml
```diff
@@ -0,0 +1,49 @@
...
+
+    - name: Build dirctl
+      id: build
+      shell: bash
+      run: |
+        set -euo pipefail
```
arpad-csepi:
```suggestion
```
Might missing something but:
`-e` aka exit after non-zero return code is not needed as taskfile also will do that
`-u` aka exit if undefined variable is referenced also does not seems to be useful here
`-o pipefail` aka disable pipe error masking is also does nothing as the taskfile command has no pipe in it. (also not sure if this set applies to the task command or also applies to the sub shell where the tasks are executed)

Issue #1151, opened 2026-03-24
[Bug]: Large coverage artifacts generated by CICD pipeline


ramizpolic:
### Bug Description

Our current coverage artifacts are over 300MB in size per each coverage run.

### Expected Behavior

Analyse if the coverage artifacts are properly packaged and if we are packaging things we dont need

### Affected Version

v1.1.0

### Steps to Reproduce

https://github.com/agntcy/dir/actions/runs/23504662843?pr=1136

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing issue

PR #1182, opened 2026-03-25
fix(tests): flaky e2e network tests due to shared test data and insufficient sync timeout

muscariello:
## Summary

Three independent sources of flakiness were identified in the e2e test suite under CI load.

---

## Root Cause 1 — Test data collision between `02_sync_test.go` and `05_name_resolution_test.go`

`05_name_resolution_test.go` was reusing the same fixture and record name/version as `02_sync_test.go`:
- Both tests pushed records under `directory.agntcy.org/cisco/marketing-strategy-v4 @ v4.0.0`
- Both compared pull output against `testdata.ExpectedRecordV070SyncV4JSON`

This created an implicit ordering dependency between independent test files. When the sync test ran concurrently or left state in the directory, the name resolution test's pre-conditions (e.g., _"should fail to resolve by name from peer 2 before sync"_) could pass or fail non-deterministically depending on which test seeded the record first.

**Affected files:**
- `tests/e2e/network/05_name_resolution_test.go`
- `tests/e2e/shared/testdata/embed.go`

---

## Root Cause 2 — Sync wait timeout too short for CI environments

Three `ShouldEventuallyContain("COMPLETED", 120*time.Second)` calls across two test files would time out under slow CI runners or network jitter between containers, since sync can legitimately take longer than 2 minutes in those environments.

**Affected files:**
- `tests/e2e/network/02_sync_test.go` (×2: peer2 and peer3 sync waits)
- `tests/e2e/network/05_name_resolution_test.go` (×1: peer2 sync wait)

---

## Root Cause 3 — `kubectl port-forward` processes dying silently during test runs

Each test job sets up port-forwards from `kubectl port-forward` processes before invoking the test binaries. These processes are short-lived: they exit whenever the target pod is temporarily unavailable — for example, when a pod is restarted to flush Go coverage data via `SIGTERM`. The previous setup used a bare `sleep 10` to wait for ports to become available, which:

1. **Did not reconnect** — once a `kubectl port-forward` process died mid-run, the connection was gone for the rest of the test run, causing any subsequent gRPC call to fail with `connection refused`.
2. **Did not verify readiness** — `sleep 10` assumed the port was up after 10 seconds with no actual confirmation, so tests could race the forwarding process on slow nodes.
3. **Did not handle missing services** — the ingress controller service (`dir-ingress-controller`) is not deployed in all environments; a hard port-forward attempt would fail silently under `sleep 10` but was correctly detected as an error once real readiness checks were added.

**Affected file:**
- `tests/Taskfile.test-env.yml`

---

## Fix

- Added a dedicated fixture `tests/e2e/shared/testdata/record_070_name_resolution.json` with a unique record name (`directory.agntcy.org/cisco/name-resolution-test @ v1.0.0`) and wired it up in `embed.go` as `ExpectedRecordV070NameResolutionJSON`
- Updated `05_name_resolution_test.go` to use the new fixture and record name exclusively, fully isolating it from the sync test data
- Doubled all three sync wait timeouts from `120s` to `240s`
- Replaced all `kubectl port-forward` invocations with `while true; do kubectl port-forward ...; sleep 2; done` reconnecting loops running in the background, so a dying port-forward is automatically restarted within 2 seconds
- Added TCP readiness polling via `nc -z 127.0.0.1 $port` (up to 30 retries) to block until each required port is actually accepting connections before tests start
- Made the ingress controller port-forward conditional on the service existing in the target namespace, and excluded port 8080 from the required-ports readiness check
- Extracted the shared readiness polling logic into an internal `port-forward:wait` Taskfile task to avoid duplication across all three test environments (`local`, `local-ghcr`, `network`)

Fix branch: `fix/flakytest`

### DIFF
--- a/tests/Taskfile.test-env.yml
+++ b/tests/Taskfile.test-env.yml
@@ -219,25 +219,12 @@ tasks:
     vars:
       HELM_NAMESPACE: '{{ .HELM_NAMESPACE | default "dir-server-ghcr" }}'
     cmds:
-      # Port-forward with automatic reconnection on failure
       - |
         (while true; do {{ .KUBECTL_BIN }} port-forward service/dir-apiserver 8888:8888 -n {{ .HELM_NAMESPACE }}; sleep 2; done) &
         (while true; do {{ .KUBECTL_BIN }} port-forward service/dir-apiserver 9090:9090 -n {{ .HELM_NAMESPACE }}; sleep 2; done) &
-
-      # Wait for ports to accept connections instead of a fixed sleep
-      - |
-        for port in 8888 9090; do
-          retries=30
-          while ! nc -z 127.0.0.1 "$port" 2>/dev/null; do
-            retries=$((retries-1))
-            if [ $retries -eq 0 ]; then
-              echo "ERROR: Port $port not ready after 30 seconds"
-              exit 1
-            fi
-            sleep 1
-          done
-          echo "Port $port is ready"
-        done
+      - task: port-forward:wait
+        vars:
+          PORTS: "8888 9090"
 
   test-env:kubernetes:local:ghcr:cleanup:
     aliases: [deploy:local:ghcr:cleanup]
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -191,10 +191,10 @@ jobs:
             artifact_path: .coverage/unit
           - artifact_name: coverage-e2e-Local
             codecov_flag: e2e-Local
-            artifact_path: .
+            artifact_path: .coverage/e2e
           - artifact_name: coverage-e2e-Network
             codecov_flag: e2e-Network
-            artifact_path: .
+            artifact_path: .coverage/e2e
```
ramizpolic: seems like the coverage is not properly collected, the codecov is reporting 10% less than on the main branch

https://app.codecov.io/gh/agntcy/dir/tree/fix%2Fflakytest

@paralta please help here, this is also related to #1151

arpad-csepi: @paralta @ramizpolic I already made the fix for it: https://github.com/agntcy/dir/pull/1091

arpad-csepi: The `server.out` is stored in the `.coverage/` path while the `network.out` is under `.coverage/e2e/`, that's the case of missing report.

.github/workflows/ci.yaml
```diff
@@ -233,7 +233,7 @@ jobs:
       - name: Upload to Codecov
         uses: codecov/codecov-action@1af58845a975a7985b0beb0cbe6fbbb71a41dbad # v5.5.3
         with:
-          codecov_yml_path: .codecov.yml
+          codecov_yml_path: codecov.yml
```
ramizpolic, line 236: @arpad-csepi is this okay?

arpad-csepi: Yes, it is: https://github.com/agntcy/dir/blob/main/codecov.yml

PR #1190, opened 2026-03-26
refactor(install/chart): remove chart tarballs

paralta:
Remove vendored Helm chart tarballs from the repo. Dependencies are resolved dynamically via `helm dependency build` using the existing `Chart.yaml/Chart.lock` declarations.

### DIFF
--- a/tests/Taskfile.test-env.yml
+++ b/tests/Taskfile.test-env.yml
@@ -7,18 +7,6 @@ tasks:
   default:
     cmd: echo "Run the main Taskfile instead of this one."
 
-  helm:dep:build:
-    desc: Build Helm dependencies for the DIR chart and its subcharts
-    internal: true
-    deps:
-      - deps:helm
-    vars:
-      HELM_CHART_PATH: "{{ .ROOT_DIR }}/install/charts/dir"
-    cmds:
-      - task: helm:gen:add-repo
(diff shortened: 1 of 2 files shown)


tests/Taskfile.test-env.yml
```diff
@@ -7,6 +7,18 @@ tasks:
...
+    desc: Build Helm dependencies for the DIR chart and its subcharts
+    internal: true
+    deps:
+      - deps:helm
+    vars:
+      HELM_CHART_PATH: "{{ .ROOT_DIR }}/install/charts/dir"
+    cmds:
+      - task: helm:gen:add-repo
+      - "{{ .HELM_BIN }} dependency build {{ .HELM_CHART_PATH }}/apiserver"
+      - "{{ .HELM_BIN }} dependency build {{ .HELM_CHART_PATH }}"
```
ramizpolic: there are some other charts like runtime, im not sure if thats all, please check if we missed any

ramizpolic: nvm, i see that this is just for test-env, but it would be good to regenerate all somewhere like in `install/` taskfile

paralta: the runtime chart seems to have no dependencies so it doesn't need a build step. I will move this to the install taskfile, good point 👍

PR #1192, opened 2026-03-26
fix(ci): update to v1.2.0 and unify dirctl path across actions and wfs

tkircsi:
## Summary
- update CI workflows to use `dirctl` v1.2.0 and migrate Directory access from legacy GitHub token flags to the documented OIDC `--auth-mode=oidc` plus `--auth-token` flow
- unify `dirctl` binary handling across reusable actions and workflows by exposing `setup-dirctl` and `build-dirctl` output paths and passing `dirctl_path` through validation, push, publish, and sign actions
- update `demo-dir` and `import-records` workflows to fetch GitHub Actions OIDC tokens for Directory access while keeping Sigstore OIDC signing flow separate

## Test plan
- [ ] run `Demo - DIR` with `image_version=v1.2.0`
- [ ] run `Import MCP Records` in dry-run mode against the public gateway
- [ ] verify push, publish, search, verify, and import commands succeed with OIDC auth and no `--github-token` usage

### DIFF
--- a/.github/workflows/demo-dir.yaml
+++ b/.github/workflows/demo-dir.yaml
@@ -9,7 +9,7 @@ on:
       image_version:
         required: true
         type: string
-        default: v1.1.0
+        default: v1.2.0
         description: "dirctl version to use"
       server_addr:
         required: true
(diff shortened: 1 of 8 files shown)


.github/workflows/demo-dir.yaml
paralta: please check these methods as well

tkircsi: nice catch! Thx!

Issue #1072, opened 2026-03-19
[Epic] Software distribution for multiple platform


muscariello:
We want to support local usage of agent directory in desktop devices.

- Windows
- macOS
- Linux 

zot already supports these targets and we want to have the same for the apiserver.

paralta:
The goal of this epic is to enable local usage of the Agent Directory on desktop devices (macOS, Linux, Windows) by introducing a **directory daemon** -- we plan to achieve this with a `dirctl daemon` command that packages all required components (gRPC server, reconciler, database, OCI store) into one process with zero external dependencies.

Users should be able to run a fully functional local directory with:

```sh
dirctl daemon start
```

## Sugested Approach

### Phase 1: Basic local directory (this epic)

The initial implementation focuses on a **self-contained local deployment** that requires no external services:

- **Database**: Embedded SQLite via GORM (pure-Go driver, no CGO)
- **OCI store**: Filesystem-based OCI layout via ORAS (no Zot or external registry)
- **Server**: gRPC apiserver bound to `localhost`
- **Reconciler**: All background tasks (indexer, signature, name, regsync) running in-process

### Phase 2: External backends and dedicated binary (future)

Once the daemon foundation is in place, we can extend it to connect to external services and optionally ship it as a standalone binary:

- **External OCI registries**: Zot, GHCR, ECR, ACR, or any OCI-compliant registry as the backing store instead of the local filesystem
- **External databases**: PostgreSQL or other GORM-supported databases for production-grade deployments that need concurrent access or larger-scale indexing
- **Dedicated binary**: Extract the daemon into a separate binary for cleaner separation of concerns between the CLI client and the server process, while keeping `dirctl daemon` as an alias for convenience

## Architecture

### Current: Multi-process deployment

Today, running the directory requires 4 separate processes coordinated via Docker Compose or Helm:

```
dirctl ──gRPC──> apiserver ──OCI──> Zot (registry)
                    │
                    └──SQL──> PostgreSQL
                                  ▲
         reconciler ──SQL─────────┘
              │
              └──────OCI──> Zot (registry)
```

### Proposed: Single-process daemon

The daemon bundles everything into a single `dirctl daemon` process:

```
dirctl ──gRPC (localhost)──> ┌─────────────────────────────┐
                             │     dirctl daemon            │
                             │                              │
                             │  ┌─────────────┐             │
                             │  │ gRPC Server │             │
                             │  └──────┬──────┘             │
                             │         │                    │
[...]

PR #1195, opened 2026-03-27
feat(server/database): add SQLite support and configuration

paralta:
Re-adds SQLite support removed in [#951](https://github.com/agntcy/dir/pull/951). Needed for the upcoming `dirctl daemon` single-process mode ([#1072](https://github.com/agntcy/dir/issues/1072)) which runs locally without external PostgreSQL.

* Add `SqliteConfig` and sqlite case to the GORM database factory using github.com/glebarez/sqlite (pure Go, no CGO)
* Enable `PRAGMA foreign_keys = ON` for CASCADE support
* Wire `database.sqlite.path` into server and reconciler config loaders
* Restore unit tests against in-memory SQLite

### DIFF
--- a/server/database/database.go
+++ b/server/database/database.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/agntcy/dir/server/database/config"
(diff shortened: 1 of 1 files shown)


server/database/config/config.go
```diff
@@ -7,6 +7,9 @@ const (
 	// DefaultType is the default database type if not specified in the config.
 	DefaultType = "postgres"
 
+	// SQLite defaults.
+	DefaultSQLitePath = "/tmp/dir.db"
```
ramizpolic: this will flush the database between restarts, we need to use some other dir that wont be flushed such as homedir instead. we probably would need to handle creation as well since we will use it for other common config later on. this will preserve the data between restarts.

```go
var DefaultDataDir = EnsureFilePath(filepath.Join(GetDataDir(), ".dir"))
var DefaultSQLitePath = EnsureFilePath(filepath.Join(DefaultDataDir, "dir.db"))

// Shared utils
func GetDataDir() string {
	homeDir, err := os.UserHomeDir()
	if err != nil {
		return os.TempDir()
	}

	return homeDir
}

func EnsureFilePath(path string) string {
	// Create dir-path, panic on error
	if err := os.MkdirAll(filepath.Dir(path), 0o666); err != nil {
		panic(err)
	}

	// Return abs path, fallback to relative
	absPath, err := filepath.Abs(path)
	if err != nil {
		return path
	}
	return absPath
}
```

paralta: thanks for reviewing! updated 👍

PR #1200, opened 2026-03-27
feat(cli): add daemon command

paralta:
Adds `dirctl daemon start|stop|status` — a self-contained local directory server that bundles the gRPC apiserver and reconciler in one process with embedded SQLite and filesystem OCI store.

- Add `dirctl daemon` subcommands (`start`, `stop`, `status`) with PID file management, `--data-dir` override, and SIGTERM graceful shutdown
- Export `server.Start()` and refactor `reconciler/service.New()` to accept DB/store/config so both can be embedded by the daemon
- Change indexer task from `*remote.Repository` to `registry.TagLister` interface to support local OCI stores
- Add `walkReferrersViaPredecessors` fallback in OCI store so sign/verify work on local layout stores without the Referrers API
- Add `test:e2e:daemon` Taskfile task and Ginkgo test suite covering push, pull, sign, and verify against a running daemon
- Register daemon e2e job in CI workflow matrix

### DIFF
--- a/tests/e2e/daemon/daemon_suite_test.go
+++ b/tests/e2e/daemon/daemon_suite_test.go
@@ -4,12 +4,11 @@
 package daemon
 
 import (
-	"fmt"
-	"net"
 	"os"
 	"testing"
 	"time"
 
+	"github.com/agntcy/dir/tests/e2e/shared/utils"
 	ginkgo "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 )
@@ -31,23 +30,7 @@ var _ = ginkgo.BeforeSuite(func() {
 		addr = defaultServerAddress
 	}
 
-	ginkgo.GinkgoWriter.Printf("Waiting for daemon at %s...\n", addr)
-	gomega.Expect(waitForServerReady(addr)).To(gomega.Succeed())
-	ginkgo.GinkgoWriter.Printf("Daemon is ready at %s\n", addr)
+	ginkgo.GinkgoWriter.Printf("Waiting for Directory daemon at %s...\n", addr)
+	gomega.Eventually(utils.IsGrpcServerReady).WithArguments(addr).WithPolling(readyPollInterval).WithTimeout(readyTimeout).Should(gomega.Succeed())
+	ginkgo.GinkgoWriter.Printf("Directory daemon is ready at %s\n", addr)
 })
-
-func waitForServerReady(addr string) error {
-	deadline := time.Now().Add(readyTimeout)
-	for time.Now().Before(deadline) {
-		conn, err := net.DialTimeout("tcp", addr, 2*time.Second) //nolint:gosec,noctx,mnd
-		if err == nil {
-			conn.Close()
-
-			return nil
-		}
-
-		time.Sleep(readyPollInterval)
-	}
-
-	return fmt.Errorf("daemon did not become ready at %s within %v", addr, readyTimeout)
-}
--- a/tests/e2e/mcp/mcp_suite_test.go
+++ b/tests/e2e/mcp/mcp_suite_test.go
@@ -31,10 +31,6 @@ var _ = ginkgo.BeforeSuite(func() {
 	}
 
 	ginkgo.GinkgoWriter.Printf("Waiting for Directory apiserver at %s...\n", addr)
-	gomega.Eventually(utils.IsGrpcServerReady).
-		WithArguments(addr).
-		WithPolling(readyPollInterval).
-		WithTimeout(readyTimeout).
-		Should(gomega.Succeed())
+	gomega.Eventually(utils.IsGrpcServerReady).WithArguments(addr).WithPolling(readyPollInterval).WithTimeout(readyTimeout).Should(gomega.Succeed())
 	ginkgo.GinkgoWriter.Printf("Directory apiserver is ready at %s\n", addr)
 })


.github/workflows/reusable-test-e2e.yaml
```diff
@@ -101,20 +103,20 @@ jobs:
           IMAGE_REPO: ${{ inputs.image_repo }}
           IMAGE_TAG: ${{ inputs.image_tag }}
           E2E_COMPILE_OUTPUT_DIR: tmp
-          E2E_COVERAGE_ENABLED: ${{ matrix.task != 'test:e2e:spire' && matrix.task != 'test:e2e:mcp' && inputs.enable_coverage && 'true' || 'false' }}
+          E2E_COVERAGE_ENABLED: ${{ matrix.task != 'test:e2e:spire' && matrix.task != 'test:e2e:mcp' && matrix.task != 'test:e2e:daemon' && inputs.enable_coverage && 'true' || 'false' }}
```
ramizpolic: unrelated nit here because it keeps growing: we should simply extend the matrix config to avoid doing if checks, simply by doing this:


```yaml
	# extending the matrix with new keys for granular step execution
      matrix:
        include:
          - task: test:e2e:local
            label: "Local"
			coverage: true
          - task: test:e2e:network
            label: "Network"
			coverage: true
          - task: test:e2e:mcp
            label: "MCP"
			coverage: false
          - task: test:e2e:spire
            label: "Federation"
			coverage: false

	  # simply pass the argument here
      - name: Run end-to-end tests
        env:
          IMAGE_REPO: ${{ inputs.image_repo }}
          IMAGE_TAG: ${{ inputs.image_tag }}
          E2E_COMPILE_OUTPUT_DIR: tmp
          E2E_COVERAGE_ENABLED: ${{ matrix.coverage && inputs.enable_coverage }}
        run: |
          # Use a random cluster name to avoid conflicts in concurrent runs.
          # Kind cluster names must be lowercase and can contain only letters and numbers.
          export KIND_CLUSTER_NAME=test-$(openssl rand -base64 12 | tr -dc 'a-z0-9')
          task ${{ matrix.task }}
.... rest of configs

```

cli/cmd/daemon/daemon.go
```diff
@@ -0,0 +1,59 @@
...
+func DefaultDataDir() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return filepath.Join(os.TempDir(), ".agntcy", "dir")
+	}
+
+	return filepath.Join(home, ".agntcy", "dir")
+}
+
+func pidFilePath() string { return filepath.Join(dataDir, "daemon.pid") }
```
ramizpolic: nit: unify all the variables into one place, there are several places where these are read/accessed/written

```suggestion
var (
	dataDir = getDataDir()
	dataDbFile = filepath.Join(dataDirPath, "dir.db")
	dataStoreDir = filepath.Join(dataDirPath, "store")
	dataPidFile = filepath.Join(dataDirPath, "daemon.pid")
)

// getDataDir() is the base directory for all daemon state.
func getDataDir() string {
	home, err := os.UserHomeDir()
	if err != nil {
		return filepath.Join(os.TempDir(), ".agntcy", "dir")
	}

	return filepath.Join(home, ".agntcy", "dir")
}
```

ramizpolic: if you want to enable dynamic configuration, then variables should flow from a config

```
type Options struct {
	DataDir string
}

func (o *Options) GetDBFile() string {
	return filepath.Join(o.DataDir, "dir.db")
}

etc
```

cli/cmd/daemon/start.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create server: %w", err)
+	}
+
+	if err := srv.Start(ctx); err != nil {
+		return fmt.Errorf("failed to start server: %w", err)
+	}
+	defer srv.Close(ctx)
+
+	presenter.Println(cmd, "Server started", "address", serverCfg.ListenAddress)
```
ramizpolic: lets not use presenter in the daemon and use logger instead. if this is pinned to something like systemd, then we want to be able to collect logs properly

cli/cmd/daemon/start.go
```diff
@@ -0,0 +1,230 @@
...
+
+	// Create a local OCI tag lister for the indexer to discover records.
+	storePath := filepath.Join(dataDir, "store")
```
ramizpolic: should be received from `options` method

cli/cmd/daemon/start.go
```diff
@@ -0,0 +1,230 @@
...
+		return 0, fmt.Errorf("failed to read PID file: %w", err)
+	}
+
+	pid, err := strconv.Atoi(string(data))
+	if err != nil {
+		return 0, fmt.Errorf("invalid PID file: %w", err)
+	}
+
+	return pid, nil
+}
```
ramizpolic: consider moving this to status command/file and simplifying a bit. you can reuse it in both daemon start, stop, and in status command. the handling logic for read/write/remove is a bit verbose (e.g. checkNotRunning could be moved to readPID to simply return `readPID() (bool, int, error)` where true is found and running, false found and not running, and error any reading-related error

reconciler/tasks/indexer/task.go
```diff
@@ -52,7 +52,7 @@ var emptySnapshot = &registrySnapshot{
 }
 
 // NewTask creates a new indexer reconciliation task.
-func NewTask(config Config, localRegistry ociconfig.Config, store types.StoreAPI, repo *remote.Repository, db types.SearchDatabaseAPI) (*Task, error) {
```
ramizpolic, line 55: nit: in the future, we should avoid passing the internal dependency (`*remote.Repository/registry.TagLister`) across methods and use only StoreAPI instead. if there are methods we need, we should define them on the StoreAPI and implement them using the dependency. 

this makes the internal dependencies cross API boundaries and makes the code harder to change in the future

tests/Taskfile.yml
```diff
@@ -251,6 +252,31 @@ tasks:
...
+          rm -rf {{ .DAEMON_DATA_DIR }}
+      - |
+        {{ .DIRCTL_BIN }} daemon start --data-dir {{ .DAEMON_DATA_DIR }} &
+        # Wait for the daemon to be ready (PID file written)
+        for i in $(seq 1 30); do
+          if {{ .DIRCTL_BIN }} daemon status --data-dir {{ .DAEMON_DATA_DIR }} 2>&1 | grep -q "running"; then
+            break
+          fi
+          sleep 1
+        done
```
ramizpolic: this logic gate check should be implemented inside `Before` logic inside the e2e daemon tests similar to how we check for API being up for other e2e tests

cli/cmd/daemon/daemon.go
paralta: thanks for reviewing! just pushed a commit to address these comments 👍

tests/e2e/daemon/daemon_suite_test.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+var _ = ginkgo.BeforeSuite(func() {
+	addr := os.Getenv("DIRECTORY_CLIENT_SERVER_ADDRESS")
+	if addr == "" {
+		addr = defaultServerAddress
+	}
+
+	ginkgo.GinkgoWriter.Printf("Waiting for daemon at %s...\n", addr)
+	gomega.Expect(waitForServerReady(addr)).To(gomega.Succeed())
```
ramizpolic: nit: we dont need to redeclare a new method for checking running state, simply reuse from utils package `github.com/agntcy/dir/tests/e2e/shared/utils`

```suggestion
	ginkgo.GinkgoWriter.Printf("Waiting for Directory daemon at %s...\n", addr)
	gomega.Eventually(utils.IsGrpcServerReady).
		WithArguments(addr).
		WithPolling(readyPollInterval).
		WithTimeout(readyTimeout).
		Should(gomega.Succeed())
	ginkgo.GinkgoWriter.Printf("Directory apiserver is daemon at %s\n", addr)
```

PR #1209, opened 2026-03-30
feat(dir): add pre-commit hooks

csirmazbendeguz:
add [pre-commit](https://pre-commit.com/) hooks to the project, so the linter is run automatically when committing code (note: this only works if pre-commit is installed with `task deps`, this will not work in GitHub editor for example)

### DIFF
--- a/Taskfile.vars.yml
+++ b/Taskfile.vars.yml
@@ -91,7 +91,7 @@ vars:
   CONTROLLER_GEN_BIN: "sigs.k8s.io/controller-tools/cmd/controller-gen@{{ .CONTROLLER_GEN_VERSION }}"
 
   ## Go module discovery
-  GO_MOD_DIR:
+  GO_MODULES:
     sh: find . -name go.mod -not -path "./tmp*" -exec dirname {} \;
-  GO_MOD_DIR_UNIT_TEST:
+  GO_MODULES_UNIT_TEST:
     sh: find . -name go.mod -not -path "./tests*" -not -path "./tmp*" -exec dirname {} \;
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -223,7 +223,7 @@ tasks:
     vars:
       EXTRA_ARGS: '{{ .EXTRA_ARGS | default "" }}'
     cmds:
-      - for: { var: GO_MOD_DIR_UNIT_TEST }
+      - for: { var: GO_MODULES_UNIT_TEST }
         cmd: |
           echo "Running tests in {{.ITEM}}"
           go -C {{.ITEM}} test ./... {{.EXTRA_ARGS}}
@@ -241,7 +241,7 @@ tasks:
           mkdir -p {{.COVERAGE_DIR}}/unit
         if: "[ ! -d {{.COVERAGE_DIR}}/unit ]"
 
-      - for: { var: GO_MOD_DIR_UNIT_TEST }
+      - for: { var: GO_MODULES_UNIT_TEST }
         # NOTE: CI does not need go test stdout, local uses separate command to show the report
         cmd: |
           {{if ne .OVERRIDEN_COVERAGE_MODULES ""}}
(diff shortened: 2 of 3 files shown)


Taskfile.vars.yml
```diff
@@ -90,6 +92,13 @@ vars:
...
   GO_MOD_DIR:
-    sh: find . -name go.mod -not -path "./tmp*" -exec dirname {} \;
+    sh: |
+      if [ -n "${GO_MOD_DIR:-}" ]; then
+        for d in $GO_MOD_DIR; do
+          printf '%s\n' "$d"
+        done
+      else
+        find . -name go.mod -not -path "./tmp*" -exec dirname {} \;
+      fi
```
ramizpolic: why change this? you can pass a function to `find` as well to print the found paths, ie `find -exec [COMMAND]`

Taskfile.yml
```diff
@@ -311,13 +311,18 @@ tasks:
...
       FIX: '{{ .FIX | default "false" }}'
-      FIX_FLAG: '{{if eq .FIX "true"}}--fix{{end}}'
+      NEW_FROM_REV: '{{ .NEW_FROM_REV | default "" }}'
     cmds:
       - for: { var: GO_MOD_DIR }
         cmd: |
-          echo "Running golangci-lint in {{.ITEM}}"
+          GOLANGCI_LINT_ARGS=()
+          if [ "{{.FIX}}" == "true" ]; then GOLANGCI_LINT_ARGS+=("--fix"); fi
+          if [ -n "{{.NEW_FROM_REV}}" ]; then GOLANGCI_LINT_ARGS+=("--new-from-rev", "{{.NEW_FROM_REV}}"); fi
```
ramizpolic: avoid doing this logic checks using bash and try to reuse taskfile as much as possible where applicable, e.g. (not tested and there is probably a cleaner solution)

```
vars:
	FIX_FLAG: '{{if eq .FIX "true"}}--fix{{end}}'
	FIX_FLAG_REV: '{{if ne .NEW_FROM_REV ""}}--new-from-rev {{.NEW_FROM_REV}}{{end}}'
	LINT_FLAGS: '{{.FIX_FLAG}} {{.FIX_FLAG_REV}}'
```

Taskfile.vars.yml
csirmazbendeguz: I added a comment to explain

Taskfile.yml
csirmazbendeguz: ok

golangci-lint.sh
ramizpolic, line 1: this is not needed and goes against the structure of the project. we should only use taskfile

.pre-commit-config.yaml
```diff
@@ -0,0 +1,12 @@
+repos:
+  - repo: local
+    hooks:
+      - id: golangci-lint
+        name: golangci-lint
+        description: |
+          Same as golangci-lint's official hook, but works with multiple Go modules.
+          https://github.com/golangci/golangci-lint/blob/main/.pre-commit-hooks.yaml
+        entry: ./golangci-lint.sh
```
ramizpolic:
```suggestion
        entry: task lint:go NEW_FROM_REV=HEAD FIX=true
```

no need to pass extra modules, the previous logic applied things for all modules

Taskfile.yml
```diff
@@ -312,12 +312,13 @@ tasks:
...
       FIX: '{{ .FIX | default "false" }}'
       FIX_FLAG: '{{if eq .FIX "true"}}--fix{{end}}'
+      NEW_FROM_REV: '{{ .NEW_FROM_REV | default "" }}'
     cmds:
       - for: { var: GO_MOD_DIR }
         cmd: |
           echo "Running golangci-lint in {{.ITEM}}"
           cd {{.ITEM}}
-          {{.GOLANGCI_LINT_BIN}} run --config {{.ROOT_DIR}}/.golangci.yml {{.FIX_FLAG}}
+          {{.GOLANGCI_LINT_BIN}} run --config {{.ROOT_DIR}}/.golangci.yml {{.FIX_FLAG}} --new-from-rev="{{.NEW_FROM_REV}}"
```
ramizpolic: what happens here if {{.NEW_FROM_REV}} is empty?

.pre-commit-config.yaml
csirmazbendeguz: what I was going for is to only run it for the modules that changed, so it's not too slow

Taskfile.yml
csirmazbendeguz: based on my testing, it works the same as if it wasn't set

golangci-lint.sh
csirmazbendeguz: I can move it to a task but I would argue that it's needed

PR #1219, opened 2026-03-31
feat(sdk/js): add OIDC auth support for TypeScript SDK

arpad-csepi:
This PR adds OAuth 2.0 / OIDC authentication to the Directory Python SDK so gRPC calls can send a Bearer access token when auth_mode is oauth_pkce.

Supports: 
- interactive browser login (Authorization Code + PKCE) with a loopback callback
- non-interactive access via a pre-provisioned token
- machine / CI flows via the client credentials grant

Configuration is extended with OIDC/OAuth fields: issuer, client id/secret, redirect URI, callback port, auth timeout, scopes, optional static access token, and machine client settings (id, secret, secret file, scopes, optional token endpoint) with the corresponding `DIRECTORY_CLIENT_*` environment variable wiring for these options.

### DIFF
--- a/sdk/dir-js/src/client/tokenCache.ts
+++ b/sdk/dir-js/src/client/tokenCache.ts
@@ -31,7 +31,7 @@ function formatTimestamp(value: Date | undefined): string | undefined {
   if (value === undefined) {
     return undefined;
   }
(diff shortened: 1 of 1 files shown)


tkircsi (commented):
LGTM

PR #1277, opened 2026-04-07
feat(daemon): wire in runtime discovery component

ramizpolic:
This PR wires runtime discovery into the daemon and integrates it with the runtime API surface, introducing an embedded discovery flow.

### What changed

- Added runtime configuration to the daemon, including:
  - Feature toggle for runtime discovery
  - Runtime adapter settings for Docker and Kubernetes
  - Resolver settings for A2A and OASF
  - Store type selection with SQLite as default for runtime flow
- Updated daemon startup so that when runtime discovery is enabled it:
  - Initializes a runtime store
  - Starts the discovery service in-process
  - Registers runtime gRPC services on the existing API server
- Refactored runtime components into reusable packages
- Added SQLite-backed runtime store implementation using GORM and pure-Go SQLite, including:
  - Workload registration, update, delete, list, get, and watch behavior
  - Schema migration and workload payload serialization
- Enhanced Docker runtime adapter behavior:
  - Host mode support
  - Improved address and port extraction with deduplication
  - Deterministic ordering for extracted collections
- Improved resolver behavior:
  - A2A resolver now probes both HTTP and HTTPS discovery URLs
  - OASF resolver supports both name:version and CID-based record resolution
  - OASF resolver can use optional custom client config
  - OASF result payload now includes verification success and verification data

### Why this matters

This change makes runtime discovery a first-class daemon capability, reduces standalone process coupling, and provides an ephemeral runtime state path via SQLite that can be rebuilt on startup. It also improves resolver flexibility and runtime metadata discovery fidelity in containerized environments.

### DIFF
--- a/runtime/store/sql/store.go
+++ b/runtime/store/sql/store.go
@@ -7,6 +7,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"time"
 
 	runtimev1 "github.com/agntcy/dir/runtime/api/runtime/v1"
@@ -178,11 +179,15 @@ func (s *store) ListWorkloads(ctx context.Context) ([]*runtimev1.Workload, error
 
 // WatchWorkloads watches for workload changes.
 func (s *store) WatchWorkloads(ctx context.Context, handler func(workload *runtimev1.Workload, deleted bool)) error {
+	// Take an initial snapshot of the workloads in the store to compare against for changes.
+	// We will use this to detect created, updated, and deleted workloads.
 	prev, err := s.snapshot(ctx)
 	if err != nil {
 		return err
 	}
 
+	// Poll the store at regular intervals to detect changes.
+	// We cannot rely on database triggers or notifications since they are not supported by SQL.
 	ticker := time.NewTicker(defaultWatchPollInterval)
 	defer ticker.Stop()
 
@@ -192,6 +197,7 @@ func (s *store) WatchWorkloads(ctx context.Context, handler func(workload *runti
 			//nolint:wrapcheck
 			return ctx.Err()
 		case <-ticker.C:
+			// Take a snapshot of the current workloads in the store. We will compare this with the previous snapshot to detect changes.
 			current, snapErr := s.snapshot(ctx)
 			if snapErr != nil {
 				logger.Error("watch snapshot failed", "error", snapErr)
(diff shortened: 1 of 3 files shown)


runtime/store/sql/store.go
```diff
@@ -0,0 +1,245 @@
...
+
+				handler(workload, false)
+			}
+
+			for id := range prev {
+				if _, ok := current[id]; ok {
+					continue
+				}
+
+				handler(&runtimev1.Workload{Id: id}, true)
```
ramizpolic: nit: needs to return the full data

runtime/store/sql/store.go
```diff
@@ -0,0 +1,245 @@
...
+
+			prev = current
+		}
+	}
+}
+
+func (s *store) snapshot(ctx context.Context) (map[string]string, error) {
+	var records []workloadRecord
+
+	err := s.db.WithContext(ctx).Find(&records).Error
```
ramizpolic, line 260: nit: needs to return only IDs since we only care about IDs

runtime/discovery/resolver/oasf/resolver.go
```diff
@@ -135,28 +149,35 @@ func (r *resolver) Apply(ctx context.Context, workload *runtimev1.Workload, resu
...
 	} else {
-		return "", ""
+		return "", "", "", fmt.Errorf("workload %s does not have label or annotation %s", workload.GetId(), r.labelKey)
 	}
 
 	// Split the record FQDN into name and version
-	// Expected formats: name, name:version
 	nameParts := strings.SplitN(recordFQDN, ":", 2)
+
+	// If both name and version are present, return them combined as "name:version"
```
paralta, line 172: misleading comment, name and version are returned separately

cli/cmd/daemon/start.go
```diff
@@ -64,11 +68,50 @@ func runStart(cmd *cobra.Command, _ []string) error {
...
 	defer cancel()
 
+	// Create API server
 	srv, err := server.New(ctx, &cfg.Server)
 	if err != nil {
 		return fmt.Errorf("failed to create server: %w", err)
 	}
 
+	// Create ephemeral runtime store
+	discoveryErrCh := make(chan error, 1)
```
paralta, line 78: only allocate when runtime enabled?

ramizpolic: if we do that, we wont be able to listen to it below. this way, if runtime is disabled, nothing is ever written to this channel so it does nothing, otherwise it will cancel on error

runtime/discovery/resolver/oasf/resolver.go
ramizpolic: will patch in another PR, thanks!

PR #1325, opened 2026-04-15
fix(cli): expired OIDC token related error message

### LINKED ISSUE #1317: [Bug]: Expired cached OIDC token falls back to insecure transport and surfaces a low-level connection reset
## Summary
When `dirctl` is using auth auto-detection and the cached OIDC token has expired, the client stops selecting OIDC auth and falls back to an insecure gRPC connection. Against the hosted Directory endpoint this produces a low-level transport error like `connection reset by peer` instead of a clear authentication error.

## Current behavior
With a cached but expired OIDC token, commands such as:

```text
dirctl routing list -o json
```

can fail with an error similar to:

```text
rpc error: code = Unavailable desc = connection error: desc = \"error reading server preface: read tcp ...: read: connection reset by peer\"
```

## Expected behavior
If an expired OIDC token is present, the client should fail with a clear authentication-focused error message, for example prompting the user to run `dirctl auth login` again.

It should not silently fall back to insecure transport for a remote endpoint that is expected to use OIDC/TLS.

## Suspected cause
The client auto-detection logic uses `GetValidToken()` to decide whether to attempt OIDC auth. If the cached token is expired, `GetValidToken()` returns `nil`, and the client then falls back to insecure transport instead of surfacing an auth error.

## Impact
This is confusing for users because:
- the real problem is expired auth
- the returned error looks like a network or server problem
- the configured OIDC issuer/client ID environment variables do not help once the command has chosen the insecure fallback path

## Suggested fix
When OIDC config signals are present (for example issuer/client ID are configured, or an OIDC cache entry exists even if expired), prefer returning an auth-specific error instead of falling back to insecure transport.

tkircsi:
Fix expired cached OIDC token handling in auth auto-detect mode, avoid insecure fallback when OIDC signals are present, and return a clear prompt to run dirctl auth login.

Closes #1317

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -290,6 +290,8 @@ func TestClientClose_AggregatesErrors(t *testing.T) {
 
 // TestNew_WithInsecureConfig tests creating a client with insecure configuration.
 func TestNew_WithInsecureConfig(t *testing.T) {
+	t.Setenv("XDG_CONFIG_HOME", t.TempDir())
+
 	ctx, cancel := context.WithTimeout(context.Background(), testContextTimeout)
 	defer cancel()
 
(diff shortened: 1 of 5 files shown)


paralta (approved):
LGTM 👍

PR #1327, opened 2026-04-16
feat(cli/export): add a2a and agent skill export formats 

### LINKED ISSUE #1250: Task: Add an AgentSkills download and export flow that materializes SKILL.md locally
## Problem Statement

Discovery alone is not enough. Users must be able to take an AgentSkills record from the local
Directory and materialize a usable local artifact that their agentic CLI can consume.

## Proposed Solution

Add a pull or export workflow that resolves an AgentSkills record into a concrete local artifact
layout, with `SKILL.md` as the primary deliverable.

The workflow should define:

- how the user selects the desired AgentSkills package
- where `SKILL.md` is written locally
- whether companion metadata or supporting assets are also materialized
- how the flow integrates with tools such as GitHub Copilot, Claude Code, or similar CLIs

## Alternatives Considered

Leave download as raw record retrieval only and require users to manually extract the skill file.
That would reuse existing pull behavior, but it would not satisfy the requirement of letting the
AI tool use the downloaded skill directly.

## Additional Context

This task should stay aligned with the record contract defined in task 02-01 so the exported
artifact structure remains predictable.

## Acceptance Criteria

- users can select a stored AgentSkills package and materialize `SKILL.md` locally
- the resulting artifact is usable by at least one supported local agentic CLI workflow
- the export layout and overwrite behavior are documented
- example usage is included in documentation or tests

### LINKED ISSUE #1270: Task: Add an A2A search and export workflow for the local Directory
## Problem Statement

After the A2A contract is defined, users still need a supported way to find A2A-capable agents
in the local Directory and export the corresponding A2A card or deployment material.

## Proposed Solution

Implement a workflow that lets users search for A2A-capable records in the local Directory and
export them into the A2A representation needed for local deployment.

The workflow should include:

- a clear way to identify A2A-capable records in search results
- search filters aligned with skills, domains, labels, or runtime expectations
- export of the A2A card and any companion deployment metadata required downstream
- documentation that shows how the exported output is used by the next deployment step

## Alternatives Considered

Require users to pull generic records and run raw export tooling by hand. That would preserve the
current primitives, but it would not provide an end-to-end A2A user workflow.

## Additional Context

This task should build on the existing pull, search, and export capabilities rather than invent a
second discovery path for A2A data.

## Acceptance Criteria

- users can search the local Directory for A2A-capable agents
- users can export a selected record into the A2A card and deployment material defined by task 04-01
- the search-to-export workflow is documented end to end
- at least one happy-path example is repeatable

paralta:
Adds `a2a` and `agent-skill` export formats to dirctl export, enabling records to be exported back to their original A2A AgentCard JSON or SKILL.md representation. 

Includes end-to-end tests that verify the full import→export round-trip, using a test-only static enricher to bypass the LLM dependency.


tkircsi (approved):
LGTM

PR #1334, opened 2026-04-17
feat(dir/helm): add dedicated Envoy timeout handling for EventService/Listen

### LINKED ISSUE #1333: [Feature]: Add dedicated Envoy timeout handling for EventService/Listen
## Summary
`dirctl events listen --output jsonl` intermittently exits with `rpc error: code = Internal desc = stream terminated by RST_STREAM with error code: INTERNAL_ERROR` when used through the public gateway.

Investigation points to the Envoy authz gateway using a single backend route timeout for all `/agntcy.dir` gRPC methods, which is not suitable for the long-lived `EventService/Listen` server stream.

## Current behavior
- Envoy routes all Directory gRPC methods through the generic `prefix: \"/agntcy.dir\"` route.
- That route uses `envoy.backend.timeout` (default `30s`).
- `EventService/Listen` is a long-lived stream and can be reset by Envoy once the route timeout is hit.
- From the client this appears as `RST_STREAM ... INTERNAL_ERROR`.

## Proposed fix
- Add a dedicated exact-match route for `/agntcy.dir.events.v1.EventService/Listen` ahead of the generic `/agntcy.dir` route.
- Give that route its own timeout value, e.g. `envoy.backend.listenRouteTimeout`, while keeping the existing default timeout for normal unary RPCs.
- Keep the same auth and header-mutation behavior as the generic Directory route.
- Update dev/prod overlays so authorized roles include `/agntcy.dir.events.v1.EventService/Listen`.
- Review ingress gRPC timeout annotations so nginx does not terminate the stream before Envoy.

## Acceptance criteria
- `EventService/Listen` gets a dedicated Envoy route with its own timeout knob.
- Generic `/agntcy.dir` traffic keeps the existing timeout behavior.
- Dev/prod gateway values can set the stream-specific timeout without changing global gRPC timeout behavior.
- `dirctl events listen --output jsonl` remains connected beyond the previous failure window after rollout.

tkircsi:
## Summary
- add a dedicated Envoy route for `/agntcy.dir.events.v1.EventService/Listen` with its own timeout setting
- keep the default Envoy backend timeout for other Directory gRPC methods unchanged
- document how Envoy route timeouts relate to ingress gRPC timeout settings in chart defaults and env overlays

## Details
This change fixes the event-listener stream being terminated by the generic Envoy route timeout.

`EventService/Listen` is a long-lived server stream, but it previously matched the generic `/agntcy.dir` route and inherited the default `backend.timeout` value. That made the stream vulnerable to premature termination even while normal unary RPCs should continue using the shorter default timeout.

This PR:
- adds an exact-match Envoy route for `/agntcy.dir.events.v1.EventService/Listen`
- introduces `envoy.backend.listenRouteTimeout` for that route
- keeps `envoy.backend.timeout` as the default for normal RPCs
- adds explanatory comments around Envoy and ingress timeout behavior
- adds ingress timeout annotations in dev/prod overlays to reduce the chance that ingress closes long-lived streams first

Closes #1333

### DIFF
--- a/install/charts/dir/apiserver/values.yaml
+++ b/install/charts/dir/apiserver/values.yaml
@@ -598,6 +598,15 @@ envoy-authz:
     backend:
       address: "" # e.g., dir-dir-dev-argoapp-apiserver.dir-dev-dir.svc.cluster.local
       port: 8888
+      # Default Envoy route timeout for Directory gRPC methods.
+      # This is an overall Envoy -> apiserver request/stream timeout, unlike the
+      # ingress read/send timeouts below which are mostly idle-gap timeouts.
+      timeout: 30s
+      # Dedicated timeout for the long-lived EventService/Listen stream route.
+      # Set to 0s to disable the Envoy route timeout for that RPC only.
+      # The effective stream lifetime is still capped by any tighter ingress or
+      # downstream timeout because the first layer to time out breaks the stream.
+      listenRouteTimeout: 0s
 
     oidc:
       dex:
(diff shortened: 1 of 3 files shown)


paralta (approved):
LGTM 👍

PR #1338, opened 2026-04-17
feat(cli/export): add batch export from search results

### LINKED ISSUE #1296: [Feature]: Add batch export from search results
### Problem Statement

Issue https://github.com/agntcy/dir/issues/1295 covers single-record export. Users also need to export multiple records at once, for example all MCP servers matching a skill or all A2A agents in a domain. Today this requires scripting `dirctl search` + `dirctl pull` in a loop with manual transformation.

### Proposed Solution

Extend `dirctl export` with a search-based batch mode that exports multiple records to a target
directory.

**Command structure (batch mode):**

```
dirctl export --search [query flags] --output-dir=./exports/ [flags]
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

paralta:
Extends `dirctl export` with a search-based batch mode (`--output-dir`) that searches the Directory, pulls matching records, and writes them to a local directory.

Each format handles batch export differently:

- **a2a** — one JSON file per record: `./exports/my-agent.json`
- **agent-skill** — one subdirectory per skill with its SKILL.md: `./exports/code-review/SKILL.md`
- **mcp-ghcopilot** — all matching MCP servers merged into a single config: `./exports/mcp.json`

```bash
dirctl export --output-dir=./exports/ --format=a2a --module "integration/a2a"
dirctl export --output-dir=./exports/ --format=agent-skill --module "core/language_model/agentskills"
dirctl export --output-dir=./exports/ --format=mcp-ghcopilot --module "integration/mcp"
```

By default, when multiple versions of the same record exist, only the latest semver version is exported. Use `--all-versions` to export every version with the version included in the filename.

### DIFF
--- a/cli/cmd/export/format/batch.go
+++ b/cli/cmd/export/format/batch.go
@@ -13,34 +13,6 @@ import (
 	"golang.org/x/mod/semver"
 )
 
-// RecordName extracts the name field from a record's data.
-func RecordName(record *corev1.Record) string {
-	data := record.GetData()
-	if data == nil {
-		return ""
-	}
-
-	if nameVal, ok := data.GetFields()["name"]; ok {
-		return nameVal.GetStringValue()
-	}
-
-	return ""
-}
-
-// recordVersion extracts the version field from a record's data.
-func recordVersion(record *corev1.Record) string {
-	data := record.GetData()
-	if data == nil {
-		return ""
-	}
-
-	if v, ok := data.GetFields()["version"]; ok {
-		return v.GetStringValue()
-	}
-
-	return ""
-}
-
 // SanitizeName replaces characters unsafe for filenames with hyphens.
 func SanitizeName(name string) string {
 	r := strings.NewReplacer("/", "-", "\\", "-", ":", "-", " ", "-")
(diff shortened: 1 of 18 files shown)


cli/cmd/export/format/batch.go
```diff
@@ -0,0 +1,170 @@
...
+	if data == nil {
+		return ""
+	}
+
+	if v, ok := data.GetFields()["version"]; ok {
+		return v.GetStringValue()
+	}
+
+	return ""
+}
```
adamtagscherer: Maybe we could move these functions to `api/core/v1` because I think it belongs there as a receiver on the Record structure. There might be a lot of duplications of these kind of features in the codebase at a lot of places which we could replace with it. It's ok to replace the other in another PR or just capture in an issue, but I think it's a good start to put these kind of functions there.

adamtagscherer (approved):

cli/cmd/export/format/batch.go
paralta: @adamtagscherer thanks for reviewing! agreed, working on these changes now

PR #1377, opened 2026-04-22
fix(security): 2 improvements across 2 files

tomaioo:
## Summary

fix(security): 2 improvements across 2 files

## Problem

**Severity**: `Medium` | **File**: `cli/cmd/daemon/stop.go:L24`

The `runStop` function reads a PID from a PID file and unconditionally sends `SIGTERM` to that process ID after a basic liveness check. If an attacker can tamper with the PID file (or if it becomes stale and PID is reused), this can terminate an unintended process. This is especially risky when the CLI is run with elevated privileges.

## Solution

Harden PID handling by validating process identity before signaling (e.g., verify executable path/command line matches daemon binary, optionally verify start time), ensure PID file is created with strict permissions/ownership, and reject PID files not owned by expected user.

## Changes

- `cli/cmd/daemon/stop.go` (modified)
- `cli/cmd/export/format/skill.go` (modified)

### DIFF
--- a/cli/cmd/daemon/stop.go
+++ b/cli/cmd/daemon/stop.go
@@ -100,15 +100,7 @@ func validateDaemonProcess(pid int) error {
 	}
 
 	args := strings.Split(strings.TrimRight(string(cmdline), "\x00"), "\x00")
-	isDaemonCommand := false
-	for _, arg := range args[1:] {
-		if arg == "daemon" {
-			isDaemonCommand = true
-			break
-		}
-	}
-
-	if !isDaemonCommand {
+	if len(args) < 2 || args[1] != "daemon" {
 		return fmt.Errorf("pid %d is not a daemon process", pid)
 	}
 
--- a/cli/cmd/daemon/stop.go
+++ b/cli/cmd/daemon/stop.go
@@ -101,26 +101,36 @@ func validateDaemonProcess(pid int) error {
 	}
 
 	args := strings.Split(strings.TrimRight(string(cmdline), "\x00"), "\x00")
-	if len(args) < 2 {
+	if len(args) < 3 {
 		return fmt.Errorf("pid %d has an invalid command line", pid)
 	}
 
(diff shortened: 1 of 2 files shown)


cli/cmd/daemon/stop.go
```diff
@@ -63,3 +69,30 @@ func runStop(_ *cobra.Command, _ []string) error {
...
+	}
+
+	if resolvedSelfExe, resolveErr := filepath.EvalSymlinks(selfExe); resolveErr == nil {
+		selfExe = resolvedSelfExe
+	}
+
+	targetExe, err := os.Readlink(fmt.Sprintf("/proc/%d/exe", pid))
+	if err != nil {
+		return fmt.Errorf("failed to verify process identity for pid %d: %w", pid, err)
+	}
```
ramizpolic, line 87: will this only work on certain os/architectures? my concern is that the daemon will no longer work for windows

cli/cmd/daemon/stop.go
```diff
@@ -63,3 +69,30 @@ func runStop(_ *cobra.Command, _ []string) error {
 
 	return nil
 }
+
+func validateDaemonProcess(pid int) error {
```
tkircsi, line 74: Thanks @tomaioo  for the PR!  My concern is it only checks “same executable path”, not “same daemon process”. Since dirctl is a single multi-command binary, any other live dirctl process can satisfy that predicate. Besides the validation of the binary identity, we should validate the subcommand identity, start time, PID-file ownership, or another daemon-specific marker.

tomaioo:
Great callout — the strict PID identity checks (e.g., `/proc` executable/start-time validation) are Unix-specific, so they should not be applied unconditionally. I’ll split this behind OS-specific helpers/build tags (`*_unix.go` / `*_windows.go`) so Windows keeps a compatible stop path and doesn’t regress. I’ll also add a Windows CI check to verify `daemon stop` still works.

PR #1395, opened 2026-04-22
fix(e2e/export): move export batch test

paralta:
Move the "Batch export" test context before the a2a/mcp-ghcopilot/skill contexts to avoid stale search index entries. 

The import-based contexts import records with integration/a2a and integration/mcp modules then delete them, but the delete's search index cleanup is best-effort and can leave behind CIDs that the batch export's search finds but can no longer pull.

**Failure example:** https://github.com/agntcy/dir/actions/runs/24770266620/job/72475318131?pr=1383#step:12:1105


arpad-csepi (approved):
Thank you 🔥

PR #1400, opened 2026-04-22
feat(client/cli): cache OIDC tokens per issuer

### LINKED ISSUE #1379: [Feature]: Cache dirctl OIDC tokens per issuer
## Summary
`dirctl` currently stores OIDC tokens in a single global cache file (`~/.config/dirctl/auth-token.json`) and only keeps the issuer as metadata inside that JSON. That allows valid tokens from one issuer to be reused when the user switches to a different issuer via `--oidc-issuer` or `DIRECTORY_CLIENT_OIDC_ISSUER`.

This feature should make the token cache issuer-aware so `dirctl` resolves, stores, refreshes, and clears OIDC tokens per configured issuer.

## Current behavior
- `dirctl auth login` short-circuits on any valid cached token.
- OIDC runtime auth reuses any valid cached token from the global cache.
- Auto-detect treats any cached OIDC token as a signal to use OIDC auth.
- `status` and `logout` operate on the single global cache file.

## Proposed behavior
- Cache OIDC tokens under the existing config root, but in issuer-scoped files.
- Prefer a layout like:
  - `~/.config/dirctl/tokens/<sha256(normalizedIssuer)>.json`
- Keep the `issuer` field inside each JSON file and validate that loaded cache entries match the requested issuer.
- Continue to give explicit `--auth-token` / `DIRECTORY_CLIENT_AUTH_TOKEN` precedence over cache lookup.
- Do not include backward-compatibility handling for the legacy root-level `auth-token.json`; users can re-authenticate once after release.

## CLI behavior
- `dirctl auth login`: only reuse a cached token when it matches the requested issuer.
- `dirctl auth status`: use the configured issuer when present.
- `dirctl auth logout`: use the configured issuer when present.
- OIDC auto-detect: use the configured issuer when present.
- If no issuer is configured and exactly one issuer cache exists, `status`, `logout`, and OIDC auto-detect may use it implicitly.
- If multiple issuer caches exist and no issuer is provided, `status`, `logout`, and OIDC auto-detect should return a helpful error and list cached issuers instead of guessing.

## Acceptance criteria
- Valid token for issuer A is never reused for issuer B.
- Expired token refresh reads and writes the issuer-scoped cache for the selected issuer.
- If exactly one issuer cache exists and no issuer is configured, `status`, `logout`, and OIDC auto-detect may use it.
- If multiple issuer caches exist and no issuer is configured, explicit issuer selection is required.
- Users re-authenticate once after the release; no legacy single-file cache migration is required.
- Tests cover multi-issuer cache lookup and refresh behavior.

tkircsi:
## Summary
`dirctl` now caches OIDC tokens per issuer instead of using a single global cache file. This prevents cross-issuer token reuse and makes issuer selection explicit when multiple cached issuers exist.

## What changed
- added issuer-scoped token cache resolution and hashed cache filenames
- updated login, OIDC runtime auth, status, logout, and auto-detect to use issuer-aware caches
- introduced explicit not-found and ambiguous-cache error handling
- kept CLI messaging in `cli/cmd/auth` and avoided CLI-specific behavior in `client`
- added tests covering single-issuer, multi-issuer, and refresh flows

## Behavior
- if exactly one issuer cache exists and no issuer is configured, `status`, `logout`, and auto-detect can use it
- if multiple issuer caches exist and no issuer is configured, explicit issuer selection is required
- users re-authenticate once after release; no legacy cache migration is included

Closes https://github.com/agntcy/dir/issues/1379

### DIFF
--- a/cli/cmd/auth/login.go
+++ b/cli/cmd/auth/login.go
@@ -29,7 +29,7 @@ var loginCmd = &cobra.Command{
 Default: Authorization Code + PKCE flow (opens browser).
 Device flow: --device flag (no browser needed; authorize on any device).
 
-Configuration (flags, env, or config file):
+Configuration (flags or environment):
   --oidc-issuer, DIRECTORY_CLIENT_OIDC_ISSUER        OIDC issuer URL (e.g. https://dex.example.com)
   --oidc-client-id, DIRECTORY_CLIENT_OIDC_CLIENT_ID  OIDC client ID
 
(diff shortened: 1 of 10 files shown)


paralta (approved):
LGTM 👍

Issue #1353, opened 2026-04-20
[Feature]: Enable regsync in daemon default configuration


paralta:
### Problem Statement

Even after Zot is integrated into the daemon (Issue 2), regsync is disabled by default in the daemon configuration (`reconciler.regsync.enabled: false`). 

### Proposed Solution

With the sync limitation removed, regsync should be enabled so that daemon-mode deployments can participate in federated sync out of the box.

### Checklist

- [x] I have read the [contributing guidelines](/agntcy/repo-template/blob/main/CONTRIBUTING.md)
- [x] I have verified this does not duplicate an existing feature request

PR #1422, opened 2026-04-27
feat(dirctl): embed zot into daemon subcommand

### LINKED ISSUE #1352: [Feature]: Implement embedded Zot integration in daemon startup
### Problem Statement

Once the technical approach is decided #1351, the daemon startup flow needs to be updated
to initialize and manage the Zot registry alongside the existing apiserver and reconciler.

### Proposed Solution

Depends on #1351 

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

arpad-csepi:
This PR switch daemon local OCI storage solution to Zot by adding an optional embedded Zot registry to the dir daemon flow, configured with `server.store.oci.local_dir`, `server.store.oci.registry_address`, `server.store.oci.repository_name`.

When defined, the daemon starts Zot as an daemon process and based on the apiserver OCI store config values.

The `k8s.io` go packages required to be downgraded because zot go package has older version than latest.

These changes enables to provide sync operation and federation in local development environments which detailed here: #1353 

Resolves: #1352


cli/cmd/daemon/config.go
```diff
@@ -47,6 +55,30 @@ type RuntimeConfig struct {
...
+}
+
+func registerDaemonDefaults(v *viper.Viper) {
+	v.SetDefault("daemon.zot.enabled", false)
+	v.SetDefault("daemon.zot.address", DefaultZotAddress)
+	v.SetDefault("daemon.zot.port", DefaultZotPort)
+	v.SetDefault("daemon.zot.repository_name", DefaultZotRepositoryName)
+	v.SetDefault("daemon.zot.root_directory", DefaultZotRootDirectory)
+	v.SetDefault("daemon.zot.log_level", DefaultZotLogLevel)
+}
```
ramizpolic: i would reconsider the way we configure the zot/oci registry here. my suggestion is to do the following:

- Server storage config can either be a) local dir or b) remote registry
- For local-dir storage, do the following:
	- the folder is used as a backing store
	- ZOT in-proccess is started at a random port
	- there is no mention of zot


Example configs

**Remote registry**

```
store:
    provider: "oci"
    oci:
		# uses remote-OCI registry
```

**Local registry**
```
store:
	provider: "local"
	data_dir: "./dir/to/data" # where the data will be backed
	registry_address: "localhost:5000" # what the OCI storage endpoint will be (in-process ZOT)
```

arpad-csepi: As the in-process Zot is integrated to `dirctl` not the `apiserver` I think we should not pollute the server store config with options what are not used by the server itself (like data_dir). So how I see it is that the server config describe where to connect, daemon config is to where to serve.

With this daemon section approach (which only applies to the daemon config), the user can set up any daemon related option without using server config later as well if we want to expand.

Also might be better to restructure the daemon config to include the apiserver as a subsection, so we can include the server config as is.

**Daemon config**
```yaml
apiserver:
  store:
    provider: "oci"
    ...

daemon:
  zot:
    ...
```

WDYT? Should we create a unique config for the daemon or still want to include all options to the server config?

arpad-csepi: If we add the option to the user to set where to serve the `local zot` then they can know all info how use it with `apiserver` and/or integrate with other stuff.
If we add the option to set where to listen and random pick a port then the user has no idea where to find the local zot and has the possibility for that port is already in-use.

ramizpolic: That's okay as well, we don't need to do this in server, it is definitely better to only do it on the daemon side. However, I still have the same notes on the config, for example:

```
daemon:
	store:
		# Uses OCI store with external endpoint
		provider: "oci"
		oci:
			address: "ghcr.io"

		# Uses local fs as an OCI store + starts an in-process ZOT instance
		# Works only in "daemon", and errors in "server"
		provider: "local"
		local:
			data_dir: "path/to/data"	# stores data here
			address: "localhost:5000" # default address, where the in-process ZOT is exposed
```

The approach would be the following:
- The server configuration supports both OCI or local OCI config
- The server only works for OCI-backed store
- The daemon works for OCI-backed store (no changes)
- The daemon starts in process ZOT instance on configured `address` backed by `data_dir` when "local" store provider is used

Finally, the workflow is like this:
1. User configures the **server** to use external store (`provider: oci`) - works currently
2. User configures the **server** to use local store (`provider: local`) - throw an error
3. User configures the **daemon** to use external store (`provider: oci`) - works currently
4. User configures the **daemon** to use local store (`provider: local`)
	- The daemon starts a ZOT in-process on provided `address` from dir `data_dir`
	- The daemon "patches" the `store` config to use `provider: oci` with details of the in-process OCI zot server
	- The daemon starts the `server` -> same as part 3

arpad-csepi:
```yaml
daemon:
    store: # <<< it is a server.Config struct (does not want to create a new struct for this in daemon as apiserver config changes are needed to be sync with daemon.
      provider: "local" # <<< in-process zot is the same as connecting to a remote/self-hosted oci instance
        local: # <<< needs inside server.Config struct which is in server package where no use of it
            ...
```

> The server configuration supports both OCI or local OCI config

This is what I does not like, the server why support local OCI configuration when the implementation is in daemon?

What I see as a workflow is:
- User configure `provider: oci` `registry_address` `...` for external store (remote or self-hosted)
- User configure `daemon.zot` (and apiserver part in the daemon config can be set/overridden based on these values) for "internal store" which is the same as self-hosted just does not need user action.

paralta: +1 on Ramiz's approach here. Zot is an implementation detail (users just want "local" vs "remote" storage, they shouldn't need to configure Zot-specific fields). `provider: "local"` eliminates the config duplication and keeps the interface storage-backend agnostic.

server/store/local/local.go
```diff
@@ -0,0 +1,102 @@
...
+	"strconv"
+
+	"github.com/agntcy/dir/utils/logging"
+	zotapi "zotregistry.dev/zot/v2/pkg/api"
+	zotconfig "zotregistry.dev/zot/v2/pkg/api/config"
+)
+
+var logger = logging.Logger("local_oci")
+
+func RunEmbeddedZot(parentCtx context.Context, address string, port int, rootDirectory string, logLevel string) context.Context {
```
paralta: as we discussed, I believe the server should be "storage-backend agnostic". If this component cannot be mocked in the unit tests that are failing after the local store was removed, then I suggest we move this to the utils module

server/store/local/local.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+
+	defer listener.Close()
+
+	if addr, ok := listener.Addr().(*net.TCPAddr); ok {
+		return addr.Port, nil
+	}
+
+	return 0, fmt.Errorf("failed to get address")
+}
```
paralta: If a user explicitly configures `store.oci.registry_address` like localhost:6000, the configured port value will be overwritten, right?

server/store/oci/config/config.go
```diff
@@ -17,9 +17,9 @@ const (
 )
 
 type Config struct {
-	// Path to a local directory that will be to hold data instead of remote.
-	// If this is set to non-empty value, only local store will be used.
-	LocalDir string `json:"local_dir,omitempty" mapstructure:"local_dir"`
+	// Path to a local directory that will be used to store data.
+	// If empty, data will not be stored locally.
+	DataDir string `json:"data_dir,omitempty" mapstructure:"data_dir"`
```
paralta: Initially I was confused with this naming because in the daemon we are also using `DataDir` for the directory where we save the db, keys, etc

server/store/store.go
```diff
@@ -14,13 +14,14 @@ import (
...
+	OCI   = Provider("oci")
+	Local = Provider("local")
 )
 
 // TODO: add options for adding cache.
 func New(opts types.APIOptions) (types.StoreAPI, error) {
 	switch provider := Provider(opts.Config().Store.Provider); provider {
-	case OCI:
+	case OCI, Local:
 		store, err := oci.New(opts.Config().Store.OCI)
```
paralta: If the OCI config will be shared between OCI and local maybe we can keep the old logic and only have `OCI` as provider. If `DataDir` is not empty, then we know it is local on the daemon side

cli/cmd/daemon/README.md
```diff
@@ -0,0 +1,159 @@
+# dirctl daemon configuration
```
paralta: [nit] I would refrain from adding documentation in the code as it will be harder to maintain as we already have daemon documentation in the official docs

server/store/local/local.go
ramizpolic: This is not going to work since the listener will unassign the available port after closing. It is not guaranteed that the same random port is available after checks (calling this method).

cli/cmd/daemon/README.md
arpad-csepi: Thanks to point out that we already have daemon documentation, I will update that with the changes.

server/store/oci/config/config.go
arpad-csepi: Routing, PID file and database also stored in under `data-dir`. Those are also needs a value when not local registry is in use, am I correct?

LocalDir is good for this then?

server/store/local/local.go
arpad-csepi: Yes, the implementation has flaws in it, as the listener closed too soon and because of that the guarantee is lost. It can be re-implemented, but after discussion the decision is to use fixed `registry_address` only for now.

server/store/oci/config/config.go
paralta: the original `LocalDir ` is fine, my comment was on the outdated naming that was conflicting with the daemon's `DataDir`

cli/cmd/daemon/start.go
```diff
@@ -68,6 +68,22 @@ func runStart(cmd *cobra.Command, _ []string) error {
...
+	zotCtx := context.Background()
+
+	if cfg.Server.Store.OCI.LocalDir != "" {
+		if cfg.Server.Store.OCI.RegistryAddress == "" {
+			return fmt.Errorf("registry address is required for in-process OCI server when using local_dir in oci config")
+		}
+
+		if cfg.Server.Store.OCI.RepositoryName == "" {
+			return fmt.Errorf("repository name is required for in-process OCI server when using local_dir in oci config")
+		}
```
paralta: https://github.com/agntcy/dir/blob/main/server/store/oci/config/config.go#L15 we can use these default values to simplify user configuration

cli/cmd/daemon/start.go
```diff
@@ -68,6 +68,22 @@ func runStart(cmd *cobra.Command, _ []string) error {
...
+
+		zotCtx = runEmbeddedZot(ctx, cfg.Server.Store.OCI.RegistryAddress, cfg.Server.Store.OCI.LocalDir)
+
+		cfg.Server.Store.OCI.Insecure = true
```
paralta, line 76: you need `cfg.Server.Store.OCI.LocalDir = ""` here as discussed

arpad-csepi: Added

cli/cmd/daemon/start.go
arpad-csepi: Removed it and made default values applied in the config parsing process.

paralta (approved):
LGTM

ramizpolic (approved):

PR #1428, opened 2026-04-29
refactor(reconciler): use fork to execute regsync

csirmazbendeguz:
adding my fork of regclient: https://github.com/csirmazbendeguz/regclient , so we can remove the regsync binary and use regsync like a Go library instead

we should find a way to contribute this to regclient so we can remove my fork in the future. adding this so it's done for the release

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -8,7 +8,6 @@ variable "IMAGE_REPO" { default = "ghcr.io/agntcy" }
 variable "IMAGE_TAG" { default = "v0.1.0-rc" }
 variable "BUILD_LDFLAGS" { default = "-s -w -extldflags -static" }
 variable "IMAGE_NAME_SUFFIX" { default = "" }
-variable "REGSYNC_VERSION" { default = "v0.11.1" }
 
 function "get_tag" {
   params = [tags, name]
@@ -87,9 +86,7 @@ target "dir-reconciler" {
     "_common",
     "docker-metadata-action",
   ]
-  args = {
-    REGSYNC_VERSION = "${REGSYNC_VERSION}"
-  }
+  args = {}
   tags = get_tag(target.docker-metadata-action.tags, "${target.dir-reconciler.name}")
 }
 
--- a/reconciler/tasks/regsync/config.go
+++ b/reconciler/tasks/regsync/config.go
@@ -16,9 +16,6 @@ const (
 	// DefaultConfigPath is the default path for the regsync configuration file.
 	DefaultConfigPath = "/etc/regsync/regsync.yaml"
 
-	// DefaultBinaryPath is the default path to the regsync binary.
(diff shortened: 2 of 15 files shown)


reconciler/tasks/regsync/worker.go
```diff
@@ -134,43 +133,22 @@ func (w *Worker) runRegsync(ctx context.Context, configPath string) error {
...
-
 	//nolint:gosec // Binary path is from trusted configuration
-	cmd := exec.CommandContext(execCtx, binaryPath, "once", "-c", configPath)
-
-	// Capture stdout and stderr
-	var stdout, stderr bytes.Buffer
-
-	cmd.Stdout = &stdout
-	cmd.Stderr = &stderr
+	cmd, _ := regsync.NewRootCmd()
```
ramizpolic, line 139: lets check here against an error (if it can return an error), regardless if its always nil
```suggestion
	cmd, err := regsync.NewRootCmd()
	if err != nil {
	////
```

csirmazbendeguz: that's not an error, it returns `rootOpts`

ramizpolic (approved):

ramizpolic:
@csirmazbendeguz there are places in the code that are still downloading (taskfile, bake, etc) and using regsync. can we remove those as well?

csirmazbendeguz:
> @csirmazbendeguz there are places in the code that are still downloading and using regsync. can we remove those as well?

yes of course, gotta remove it from the Dockerfile

reconciler/tasks/regsync/config.go
```diff
@@ -16,9 +16,6 @@ const (
 	// DefaultConfigPath is the default path for the regsync configuration file.
 	DefaultConfigPath = "/etc/regsync/regsync.yaml"
```
ramizpolic, line 17: do we use this anywhere?

docker-bake.hcl
```diff
@@ -8,7 +8,6 @@ variable "IMAGE_REPO" { default = "ghcr.io/agntcy" }
 variable "IMAGE_TAG" { default = "v0.1.0-rc" }
 variable "BUILD_LDFLAGS" { default = "-s -w -extldflags -static" }
 variable "IMAGE_NAME_SUFFIX" { default = "" }
-variable "REGSYNC_VERSION" { default = "v0.11.1" }
```
ramizpolic, line 11: i think there are other places in the code (Taskfile) defining this as well

reconciler/tasks/regsync/config.go
csirmazbendeguz: I think this may be leftover from development... it's not used anywhere (not even in main). If you agree, we can address this in a different PR, I feel like it's out of scope for this one

docker-bake.hcl
csirmazbendeguz: removed it

PR #1447, opened 2026-05-04
chore(ci): separate api and server release

tkircsi:
- Split the Directory release flow into API-only and server/artifact release tasks.
- Restricted artifact publishing to root release tags only.
- Updated post-release tagging so publishing a root release creates only server-side module tags.
- Documented the new release process in `RELEASE.md`.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -187,22 +187,10 @@ tasks:
       # Switch to new branch
       - 'if [ "$(git rev-parse --abbrev-ref HEAD)" != "{{.RELEASE_BRANCH}}" ]; then git checkout -b {{.RELEASE_BRANCH}}; fi'
       # Update only the api module set in versions.yaml with the new version
-      - |
-        awk -v module_set="api" -v version="{{.RELEASE_VERSION}}" '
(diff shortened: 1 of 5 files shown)


paralta (approved):
LGTM 👍 heads up that the added tasks refer the new module sets but the `versions.yaml` file was not update

Taskfile.yml
```diff
@@ -175,6 +176,100 @@ tasks:
...
+      # Add release changes
+      - |
+        git add .
+        git commit -S --signoff -m "release(dir): prepare api release {{.RELEASE_VERSION}}"
+      # Verify Go release for the api module set only
+      - |
+        {{ .MULTIMOD_BIN }} verify
+        {{ .MULTIMOD_BIN }} prerelease --module-set-names api --skip-go-mod-tidy=true --commit-to-different-branch=false
+      - |
+        git commit --amend -S --signoff -m "release(dir): prepare api release {{.RELEASE_VERSION}}"
```
arpad-csepi: Nit: Can we do this with `yq` or else which is more readable or at least make it to a separate task in taskfile (for reusing in the server)?

ramizpolic (approved):

PR #1450, opened 2026-05-05
fix(ci): remove multimod verify

tkircsi:


paralta (approved):
[nit] maybe just add a small description on the PR on why we are removing this

Issue #1468, opened 2026-05-06
[Epic] Manager-scoped agent discovery via org hierarchy resolver


vivekkrishna:
## Problem Statement

The Agent Directory Service supports annotation-based search (e.g., \`owner.id: alice@example.com\`), but there is no way to ask: _\"give me all agents owned by anyone in Alice's org subtree.\"_

Today a user must:
1. Manually query their HR/directory system to enumerate every IC under a given manager
2. Issue one search per owner alias, or construct the full owner list themselves

This is impractical at scale — a manager may have hundreds of direct and indirect reports — and it puts the burden of org resolution on the caller.

## Proposed Solution

Introduce a pluggable **OrgResolver** interface in the ADS server that can expand a manager alias into a full list of aliases in their reporting chain. This enables a single query:

\`\`\`
dirctl search --manager alice@example.com
\`\`\`

Which internally:
1. Calls the configured OrgResolver with \`alice@example.com\`
2. Receives the full subtree: \`[alice, bob, carol, dave, ...]\`
3. Fans out to annotation search: \`owner.id IN [alice, bob, carol, dave, ...]\`
4. Returns all matching agent records

**OrgResolver interface (sketch):**
\`\`\`go
type OrgResolver interface {
    // Expand returns the manager and all direct/indirect reports under them.
    Expand(ctx context.Context, managerAlias string) ([]string, error)
}
\`\`\`

**Resolver implementations (pluggable, configured in server config):**
- `static` — file-based mapping (YAML/JSON), useful for testing and air-gapped environments
- `http` — generic REST adapter with configurable endpoint and response mapping, covers Workday, BambooHR, and other HR APIs without hard-coding vendor logic

The resolver is **optional** — when unconfigured, `--manager` returns an error and all other search behavior is unaffected.

**Annotation convention (recommended, not enforced):**
\`\`\`
annotations:
  owner.id: "alice@example.com"
  owner.type: "user"
\`\`\`

## Alternatives Considered

**Client-side resolution:** The CLI could call the HR system directly and pass the expanded owner list to annotation search. Simpler to implement, but leaks org structure to every client, requires every CLI user to have credentials for the HR system, and duplicates resolver logic across clients.

**Per-vendor integrations (Workday, BambooHR, LDAP):** Tight integrations with specific HR APIs would provide richer data but introduce vendor coupling and maintenance burden. A generic HTTP adapter with configurable field mapping covers the same cases without lock-in.

[...]

PR #1478, opened 2026-05-09
feat(dir): add ownership claim referrer, search, and CLI (v1)

### LINKED ISSUE #1559: [Feature] Support SPIFFE-based record identifiers
Add support for SPIFFE-based identification/signing/verification similar to the existing pubkey and oidc-based methods. Leverage existing `Signature` fields rather than adding other custom claims. Index the data into the local database to enable additional workflows such as ownership, groups, etc.

vivekkrishna:
## Summary

Implements SPIFFE-based record ownership using the existing \`api/sign/v1/Signature\` referrer type — as requested in the review: *"make it part of the standard Signature field instead of adding new claims object."*

Ownership claims are now pushed as \`SignatureReferrerType\` referrers with \`content_type = "agntcy.dir.ownership.v1"\` and \`annotations["owner_id"] = <SPIFFE ID>\`. No new proto message is needed.

## Design

| Field | Value |
|-------|-------|
| Referrer type | \`agntcy.dir.sign.v1.Signature\` (existing) |
| \`content_type\` | \`agntcy.dir.ownership.v1\` |
| \`annotations["owner_id"]\` | SPIFFE ID of the claiming principal |
| \`signed_at\` | RFC3339 timestamp |
| \`signature\` | Base64 ECDSA/RSA sig over SHA-256(owner_id + ":" + signed_at) |
| \`certificate\` | Base64 DER-encoded SPIFFE SVID X.509 certificate |

Unsigned claims (no \`--key\`/\`--cert\`) are accepted in dev environments; the reconciler logs a warning and indexes them without verification.

## Changes

**API**
- \`api/core/v1/referrer_types.go\`: \`OwnershipContentType = "agntcy.dir.ownership.v1"\` constant
- \`api/ownership/v1/sign.go\`: \`NewClaim\`, \`SignClaim\`, \`VerifyClaim\`, \`IsSigned\`, \`IsOwnershipClaim\`, \`GetOwnerID\` — all operating on \`*signv1.Signature\`
- Removed: \`proto/agntcy/dir/ownership/v1/claim.proto\` + generated files (no new proto needed)

**Search**
- \`proto/agntcy/dir/search/v1/record_query.proto\`: \`RECORD_QUERY_TYPE_OWNER = 20\`
- \`server/types/search.go\`: \`Owners []string\` filter + \`WithOwners()\`
- \`server/database/gorm/record.go\`: owner JOIN filter
- \`server/database/utils/utils.go\`: \`RECORD_QUERY_TYPE_OWNER\` → \`WithOwners()\`

**Database**
- \`server/database/gorm/owners.go\`: \`Owner\` model (\`record_cid\`, \`owner_id\`, \`claimed_at\`)
- \`server/database/gorm/migration.go\`: \`Owner{}\` in \`AutoMigrate\` list
- \`server/types/database.go\`: \`OwnershipDatabaseAPI\` interface (\`AddOwner\`, \`RemoveOwners\`)

**Ingestor**
- \`server/ingest/ingest.go\`: eagerly indexes ownership claims on push so search is immediate without waiting for the reconciler

**Reconciler**
- \`reconciler/tasks/ownership/task.go\`: walks \`SignatureReferrerType\` referrers, filters by \`IsOwnershipClaim\`, verifies SPIFFE signatures, syncs \`owners\` table
- \`reconciler/config/config.go\`: \`Ownership\` config field
- \`reconciler/service/service.go\`: ownership task registration

**CLI**
- \`cli/cmd/ownership/ownership.go\`: \`dirctl ownership claim [--key <pem> --cert <pem>]\`

## Test plan

- [x] All unit tests pass (\`go test ./...\` in api, server, reconciler, cli modules)
- [x] Lint clean (golangci-lint)
- [x] 7 ownership search e2e tests pass (exact match, wildcards, combined filter, negative cases)
- [x] 7 ownership signing e2e tests pass (signed claim indexed, identity mismatch rejected, idempotent re-push)

Generated with [Claude Code](https://claude.com/claude-code)


vivekkrishna:
@ramizpolic  Any comments on this?

ramizpolic:
Hey @vivekkrishna, thanks for the PR. It looks good from my end, the only issue is that we are working on simplifying the interfaces and API surface in this PR https://github.com/agntcy/dir/pull/1500, and I am not sure which order to take (merge that PR, adjust this, or something else).  I think that work will be done soon (end of next week). Can we hold this PR open until then and do a rebase together to merge this PR? 

Alternatively, I think you don't even need any code changes for your functionality since you can already push whichever referrer you want and it should work out of box. Code snippet below
https://github.com/agntcy/dir/blob/0c585155ec977518d2803496f419295b06a5b198/server/store/oci/referrers.go#L62-L75

The only issue would be that the CLI is not available, but this can be handled through a separate CLI for your needs built with DIR Go SDK. Lets sync tomorrow/when it works for you on Slack and we can decide next steps together.

vivekkrishna:
Sure @ramizpolic  , I will sync up over slack.

vivekkrishna:
Hi @ramizpolic . After analysis, merging as-is makes sense for these reasons.

The generic PushReferrer path (referrers.go:72-74) just stores a blob in OCI with no DB indexing — no owners table, no --owner search, and no foundation for manager subtree search. #1478's owners table + eager indexing + reconciler is exactly what makes ownership searchable and is the prerequisite for manager search (#1468).

On the #1500 question: planning to land #1478 on v1 PushReferrer as-is, then rebuild ownership onto the v2 ObjectStore + ObjectReferrer path once #1500 merges — does that sequencing work? Only immediate blocker is Target() added to StoreAPI in #1500 will break #1478's compile, so I'll add that method to the v1 OCI store when rebasing. Happy to hold if you'd rather do it right in v2 from the start — just didn't want to block manager search on #1500's timeline.

server/controller/store.go
```diff
@@ -252,6 +254,28 @@ func (s storeCtrl) pushReferrer(ctx context.Context, request *storev1.PushReferr
...
 	recordCID := request.GetRecordRef().GetCid()
 
+	// Enforce ownership: the caller's authenticated identity must match the owner_id in the claim.
+	var ownershipClaim *ownershipv1.Claim
+
+	if request.GetType() == corev1.OwnershipClaimReferrerType {
+		claim := &ownershipv1.Claim{}
+		if err := claim.UnmarshalReferrer(&corev1.RecordReferrer{Data: request.GetData()}); err != nil {
+			errMsg := fmt.Sprintf("failed to decode ownership claim: %v", err)
+
```
ramizpolic: this ownership claim is currently unverifiable since the following things can happen:
- user has write access to the OCI store (either directly or via sync)
- user writes an ownership claim to a non-owned resource through OCI referrers (either directly or through its own OCI that gets synced to a targeted node)
- target node indexes ownership through the reconciler
- the resource is reported as owned by someone who actually doesnt own it

this behavior enables targeted attacks across the network since the data has 2 access points:
- unsafe route through OCI that doesnt get verified in the reconciler
- secure route through the controller that does get verified

my concern here is that the ownership claims are not verifiable. we already support verifiable claims through a public key and keyless OIDC signing (`dirctl sign`) via COSIGN which gets verified and indexed into the database. I am not sure if we currently have a mechanism to search over this data, but it is present and may only need query params to be exposed on the client-side. perhaps @paralta can assist on this question 

i think a secure/verifiable way to enable this functionality is to add support for **SPIFFE-based signing**. this would provide all the things you need, and i would happily merge it. it would also be a super important feature for us, not just for local use-case, but across the (insecure) network.

could you please adjust the implementation according to the notes above and we can merge it right away

vivekkrishna: Thanks for the input @ramizpolic  , addressed with new commit.

vivekkrishna:
should we add this 1.5 milestone?

ramizpolic:
Hi @vivekkrishna, we will include this feature in v1.6, but we will need to make it part of the standard `Signature` field instead of adding new claims object. Are you okay if we cherry pick commits from this PR and continue working on it?

vivekkrishna:
Hi @ramizpolic , that works for me. Can you let me know the planned timeline for v1.6 release? Also see if you can add me or tag me in new PR for review or modify this same PR?. Also can you confirm search filter by --owner is still preserved after the signature approach, that is crucial for overall epic?

vivekkrishna:
@ramizpolic  Can you check if these updates look good to approve based on recent comments with signature?

vivekkrishna:
@ramizpolic  are we good to merge this?

ramizpolic:
Hi @vivekkrishna, apologies for a delay. I have used this PR as a base for https://github.com/agntcy/dir/pull/2125 to implement the identity and ownership claims. We need identity for a separate use-case and it goes hand-in-hand with the work that you have done. Once that PR is merged, you can leverage the feature as per your use-case.

PR #1499, opened 2026-05-14
refactor(dir/reconciler): enhance credential negotiation by introducing client config builder from context

### LINKED ISSUE #1470: [Bug]: dirctl daemon unable to sync with remote registries/apiservers
### Bug Description

The daemon mode by default runs insecure mode and using http protocol for serving the GRPC services.
If the sync operation points to a remote registry which using https protocol then a client sided error will pop up:

`Error: failed to create sync: failed to create sync: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: tls: first record does not look like a TLS handshake"`

### Expected Behavior

The sync would be created, record synced to local oci instance.

### Affected Version

main

### Steps to Reproduce

1. Get dirctl from main: `task cli:compile`
2. Start daemon mode: `dirctl daemon start`
3. Sync with a remote registry: `dirctl sync create https://dev.gateway.ads.outshift.io`

### Checklist

- [x] I have verified this does not duplicate an existing issue

paralta:
The regsync reconciler previously dialed every remote Directory node as `insecure`, which fails against TLS/OIDC-protected remotes. It now picks up the user's matching `dirctl` client context (auth mode, OIDC issuer, cached token, SPIFFE settings) when negotiating credentials.

- Added `buildClientConfigForRemote`: resolves a matching dirctl context first, falls back to the previous `authnConfig`-based insecure path.
- Added `resolveContextByServerAddress`: scans `~/.config/dirctl/config.yaml`, treats a missing file as no-match, errors on ambiguous matches.
- Added `canonicalServerAddress`: normalizes addresses (strips scheme/path, injects default `:80`/`:443`) so URL and `host:port` forms compare equal.

### DIFF
--- a/reconciler/tasks/regsync/credentials.go
+++ b/reconciler/tasks/regsync/credentials.go
@@ -5,10 +5,15 @@ package regsync
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
+	"sort"
+	"strings"
 
 	storev1 "github.com/agntcy/dir/api/store/v1"
 	"github.com/agntcy/dir/client"
+	clientconfig "github.com/agntcy/dir/client/config"
 	authnconfig "github.com/agntcy/dir/server/authn/config"
 )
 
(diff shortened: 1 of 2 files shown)


reconciler/tasks/regsync/credentials.go
```diff
@@ -86,3 +81,109 @@ func NegotiateCredentials(ctx context.Context, remoteDirectoryURL string, authnC
...
+
+	switch len(matches) {
+	case 0:
+		return "", nil
+	case 1:
+		return matches[0], nil
+	default:
+		sort.Strings(matches)
+
+		return "", fmt.Errorf("multiple client contexts match server_address %q: %v", serverAddress, matches)
```
ramizpolic: i dont think this needs to be an error. we dont really care about the context here in a sense that "executing X against context Y", but rather "executing X against DIR server address Y"

in that regard, we can assume that any context for a specific server address is valid since we only need client configs. consider simply defaulting here to the first item.

reconciler/tasks/regsync/credentials_test.go
```diff
@@ -0,0 +1,175 @@
...
+	authnconfig "github.com/agntcy/dir/server/authn/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCanonicalServerAddress(t *testing.T) {
+	tests := []struct {
+		name string
+		in   string
+		want string
```
ramizpolic, line 20: consider specifying config directly here as well so that we can control exactly the whole process in a single check
```suggestion
		name string
		in   string
		want string
		config string
```

```go
{"plain host:port", "localhost:8888", "localhost:8888", `
contexts:
  dev:
    server_address: dev.gateway.example.com:443
`},
```

ramizpolic (approved):

reconciler/tasks/regsync/credentials.go
paralta: good point!

PR #1501, opened 2026-05-16
feat(dir): record ranking

### LINKED ISSUE #1418: [Spike]: Record ranking
The records stored in the Agent Directory don't have any ranking system in them. When I search for some skills then it returns a bunch of CIDs without any ranking. So it is very hard to find good records between a lot of not so good records and we need a solution for this.

adamtagscherer:
Adds a deterministic, explainable ranking layer to search and routing search.
Results are scored on a 0–1000 composite of six signals — query relevance, trust, freshness, popularity, completeness, schema recency — sorted server-side, and each carries a structured RankExplanation with per-signal sub-scores and a ScoreStatus (FULL / PARTIAL / DEFAULTS_ONLY).
DEFAULTS_ONLY is the cross-peer case where a CID is known via the DHT but its body isn't local yet.

`dirctl search` and `dirctl routing search` gain `--explain` to render the breakdown.
Weights, saturation points, freshness half-life, and the candidate cap are all tunable via daemon config / DIRECTORY_DAEMON_* env vars.


ramizpolic:
Hi @adamtagscherer, I haven't had a chance to review this sooner, will provide my feedback by EOW and we can target this for **DIR v1.5** once it's ready.

ramizpolic:
We have moved this PR to v1.6 and there are some important concerns to consider here. I like the idea and the contract bindings from the API level, but the implementation is currently too tightly coupled across unrelated components. I would prefer we agree on the implementation and try to handle this as a standalone feature rather than having it surface across all the components. This will be problematic to maintain in the long run in the current form.

PR #1515, opened 2026-05-18
feat(cli): introduce the dirctl doctor network diagnose command

### LINKED ISSUE #1512: [Feature]: Add dirctl doctor diagnostics for Directory endpoints
## Summary

Add a top-level `dirctl doctor` command that diagnoses the effective Directory endpoint and related network services. The command should reuse the current `dirctl` context/config for Directory address and auth settings, then add doctor-specific bootstrap peer configuration for libp2p/DHT checks.

## Problem

Operators and developers currently have to combine several commands and manual checks to understand whether a Directory endpoint is usable. Failures across local daemon startup, remote gateway authentication, gRPC reachability, bootstrap peer configuration, and DHT reachability are hard to distinguish.

A separate `dirnet doctor` prototype already proves the desired behavior, but the feature should live in `dirctl` so it uses the same contexts, auth modes, token cache, and release path as the rest of the CLI.

## Goals

- Add `dirctl doctor` as a top-level command, not under the hidden `network` command.
- Reuse the effective `dirctl` client config for `server_address`, `auth_mode`, OIDC, TLS, and SPIFFE settings.
- Add a doctor-specific bootstrap peer setting for network diagnostics.
- Produce actionable human output and stable JSON output for CI/support workflows.
- Keep checks read-only.

## Proposed config model

Endpoint and auth should continue to come from existing context fields:

```yaml
current_context: dev
contexts:
  dev:
    server_address: dev.gateway.example.com:443
    auth_mode: oidc
    oidc_issuer: https://dev.idp.example.com
    oidc_client_id: dirctl
    doctor:
      bootstrap_peers:
        - /dns4/dev.routing.example.com/tcp/5555/p2p/12D3Koo...
```

The `doctor.bootstrap_peers` field is diagnostic-specific and should remain separate from the normal Directory API client config. A repeatable `--bootstrap-peer` flag should also be available for one-off diagnostics and CI overrides.

## Proposed checks

- `context_config`: resolve and report selected context/config source with sensitive values redacted.
- `directory_api_tcp`: check raw TCP reachability to the resolved `server_address`.
- `grpc_client_setup`: create a Directory client using the resolved config and existing auth behavior.
- `routing_list`: call `RoutingService.List` with `Limit=1` to verify the routing API path without mutating records.
- `bootstrap_peer_multiaddr`: validate each configured/supplied bootstrap peer multiaddr and require `/p2p/<peer-id>`.
- `bootstrap_peer_dial`: dial each valid bootstrap peer with a temporary libp2p host and report connectedness/protocol metadata.
- `dht_bootstrap_reachability`: bootstrap a temporary Kademlia DHT client using Directory's `dir` protocol prefix and report peer/routing-table reachability.

## Execution model

`doctor` should bypass the root command's automatic client setup and resolve/create the client inside the check runner. This lets config, auth, and client setup failures appear as structured doctor results instead of failing before the command can print output.

Recommended phases:

1. Resolve and validate config without creating a client.
2. Run checks that do not require the gRPC client, such as config reporting, TCP reachability, and bootstrap multiaddr validation.
3. Create the authenticated client using existing `client.New` behavior.
4. Run client-dependent checks such as `routing_list`.
5. Run libp2p/DHT checks using configured bootstrap peers.

If client creation fails, earlier checks should still be reported and client-dependent checks should be marked `skip` with an actionable reason.

## Output behavior

- Human output should be compact and actionable: resolved target/context, one line per check, details for warnings/failures, and a summary.
- JSON output should include `config`, `summary`, and `results`.
- JSON output must not include raw tokens, refresh tokens, private keys, or other secrets.
- Failed checks should cause a non-zero exit code.
- Warnings should not cause a non-zero exit code.
- Missing bootstrap peers should skip peer/DHT checks rather than failing the whole command.

## Acceptance criteria

- `dirctl doctor` is registered as a top-level command.
- The command resolves the same effective client configuration as normal API commands.
- The command supports at least human and JSON output.
- The command supports `--timeout` and repeatable `--bootstrap-peer` flags.
- The context config supports doctor-specific bootstrap peers.
- The command reports `pass`, `warn`, `fail`, and `skip` statuses.
- It is read-only and does not publish, pull, mutate, or create Directory records.
- OIDC cache and issuer ambiguity errors include actionable hints such as using `--oidc-issuer` or running `dirctl auth login`.
- Tests cover config failures, output summary, redaction, bootstrap peer validation, and exit behavior.
- Docs include examples for local daemon, context-based remote checks, OIDC-gateway-backed checks, and JSON CI output.

## Out of scope for the first milestone

- Record lifecycle E2E checks.
- Publish/pull/mutation-based diagnostics.
- Inspecting remote daemon internals or private routing tables.
- Adding a separate profile/config system similar to `dirnet.yaml`.
- Placing the command under the hidden `network` command.

tkircsi:
Closes #1512

### DIFF
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -50,6 +50,12 @@ type Context struct {
 	OIDCIssuer       string `yaml:"oidc_issuer"`
 	OIDCClientID     string `yaml:"oidc_client_id"`
 	AuthToken        string `yaml:"auth_token"`
+	Doctor           Doctor `yaml:"doctor"`
+}
+
+// Doctor holds diagnostic-only settings for dirctl doctor.
+type Doctor struct {
+	BootstrapPeers []string `yaml:"bootstrap_peers"`
 }
 
 // ResolveOptions controls context and client configuration resolution.
@@ -203,6 +209,37 @@ func Resolve(opts ResolveOptions) (*dirclient.Config, *ResolvedContext, error) {
 		if err := validateClientConfig(cfg); err != nil {
 			return nil, nil, err
 		}
+	}
+
+	return cfg, &ResolvedContext{
+		Name:   contextName,
+		Source: source,
+		Path:   path,
+	}, nil
+}
+
+// ResolveDoctor resolves diagnostic-only settings for the selected context.
+func ResolveDoctor(opts ResolveOptions) (*Doctor, *ResolvedContext, error) {
+	path, explicitPath, err := resolvePath(opts.Path)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	file, err := loadOptionalFile(path, explicitPath, opts.AllowUnknownFields)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	contextName, source := selectedContextName(opts, file)
+	cfg := &Doctor{}
+
+	if contextName != "" {
+		contextConfig, ok := file.Contexts[contextName]
+		if !ok {
+			return nil, nil, fmt.Errorf("unknown client context %q in %s", contextName, path)
+		}
+
+		cfg.BootstrapPeers = append([]string(nil), contextConfig.Doctor.BootstrapPeers...)
 	}
 
 	return cfg, &ResolvedContext{
--- a/cli/cmd/doctor/api_checks.go
+++ b/cli/cmd/doctor/api_checks.go
@@ -1,0 +1,160 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package doctor
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"strings"
+	"time"
+
+	routingv1 "github.com/agntcy/dir/api/routing/v1"
+	"github.com/agntcy/dir/client"
+)
+
+func directoryAPI(ctx context.Context, addr string, timeout time.Duration) checkResult {
+	checkCtx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	start := time.Now()
+
+	var dialer net.Dialer
+
+	conn, err := dialer.DialContext(checkCtx, "tcp", addr)
+
+	elapsed := time.Since(start)
+	if err != nil {
+		return checkResult{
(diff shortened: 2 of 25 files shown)


client/config/config.go
```diff
@@ -50,6 +50,12 @@ type Context struct {
 	OIDCIssuer       string `yaml:"oidc_issuer"`
 	OIDCClientID     string `yaml:"oidc_client_id"`
 	AuthToken        string `yaml:"auth_token"`
+	Doctor           Doctor `yaml:"doctor"`
+}
+
+// Doctor holds diagnostic-only settings for dirctl doctor.
+type Doctor struct {
+	BootstrapPeers []string `yaml:"bootstrap_peers"`
 }
```
ramizpolic, line 59: why do we need this on the client side? the client is currently not aware of the DHT that is used under the hood of a given APIserver.

maybe the approach here would be:
- the client can be: RoutingClient, ApiClient, or Client (which includes both)
- the config for the ApiClient are what we have now (server/issuer/etc)
- the config for the RoutingClient are things needed to interact with a specific network/DHT (routing address, routing keypath, etc)

this would nicely decouple the client into API vs DHT without having to define anything related for diagnostics (since it can be inferred from the configs natively)

tkircsi: this is a low level and basic DHT/libp2p network diagnostic tool not only for the client. besides the TCP/gRPC API checks it can help to the operators to validate the DHT part. It checks: 
1. is the multiaddress is valid?
2. does this bootstrap peer accept libp2p connection and expose peer metadata?
3. can a temporary DHT client reached bootstrap peer and communicate with it? (ie. populate routing table)

ramizpolic (approved):

client/config/config.go
ramizpolic: how is that any different from a reusable client for DHT operations (interaction with the low-level DHT outside of the apiserver scope)? defining a DHT-only client helps with decoupling and reusability not only in the Go SDK (client) but also in the CLI and other places.

Im okay not doing it right now, i was only pointing out that the client can be decoupled for DHT-only operations, and probably we could fully reuse the libp2p go client for this, including the config.

PR #1516, opened 2026-05-18
feat(cli): auth status refresh expired tokens

### LINKED ISSUE #1514: [Feature]: Refresh expired OIDC tokens in dirctl auth status
## Summary

Update `dirctl auth status` so it reflects whether the user can currently authenticate, not only whether the cached access token on disk is expired. When a cached OIDC access token is expired but a refresh token is available, `auth status` should attempt to refresh the token, save the refreshed cache, and report the resulting valid status.

## Problem

Today `dirctl auth status` reads the cached token and reports `Token: Expired` when the access token is expired. However, normal API commands using `client.New` can refresh an expired cached access token automatically when a refresh token exists.

This creates confusing UX:

1. `dirctl auth status` says the token is expired and asks the user to run `dirctl auth login`.
2. The next API command can still succeed because it refreshes the token internally.

`auth status` should align with the behavior users experience when running normal authenticated `dirctl` commands.

## Desired behavior

- If the cached access token is valid, keep the current behavior and report `Token: Valid`.
- If the cached access token is expired and a refresh token exists, attempt to refresh it.
- If refresh succeeds, save the refreshed token cache and report a valid/refreshed status.
- If refresh fails, report `Token: Expired` and tell the user to run `dirctl auth login`.
- If no refresh token is available, report `Token: Expired` and tell the user to run `dirctl auth login`.
- If multiple issuer-scoped token caches exist, preserve the existing ambiguity guidance to use `--oidc-issuer` or `DIRECTORY_CLIENT_OIDC_ISSUER`.

## Suggested output

For a successful refresh:

```text
Status: Authenticated
  Provider: oidc
  Subject: <subject>
  Issuer: <issuer>
  Principal type: human
  Cached at: <new-cache-time>
  Token: Refreshed ✓
  Expires: <new-expiry>
  Cache file: <path>
```

For refresh failure or missing refresh token:

```text
Status: Authenticated
  ...
  Token: Expired ✗

Run 'dirctl auth login' to re-authenticate.
```

## Acceptance criteria

- `dirctl auth status` refreshes an expired cached OIDC access token when a refresh token exists and refresh succeeds.
- The refreshed token is saved atomically to the existing issuer-scoped token cache.
- The command reports a valid/refreshed token after successful refresh.
- The command reports expired only when the token cannot be refreshed or refresh fails.
- The command keeps existing no-cache and ambiguous-issuer behavior.
- Tests cover valid token, expired refreshable token, expired token without refresh token, refresh failure, and ambiguous issuer cases.

## Notes

This is intentionally separate from `dirctl doctor`. `doctor` should continue to reuse the normal client auth path; this issue is about making `auth status` communicate the same effective authentication behavior users get from normal API commands.

tkircsi:
Closes #1514

### DIFF
--- a/cli/cmd/auth/auth.go
+++ b/cli/cmd/auth/auth.go
@@ -43,10 +43,11 @@ func resolveAuthConfig(cmd *cobra.Command, _ []string) error {
 	}
 
 	cfg, _, err := clientconfig.Resolve(clientconfig.ResolveOptions{
-		Context:        config.Context,
-		Overrides:      overrides,
-		OverrideFields: fields,
-		SkipValidation: true,
+		Context:            config.Context,
+		Overrides:          overrides,
+		OverrideFields:     fields,
+		SkipValidation:     true,
+		AllowUnknownFields: true,
 	})
 	if err != nil {
 		return fmt.Errorf("failed to resolve auth config: %w", err)
--- a/cli/cmd/auth/logout.go
+++ b/cli/cmd/auth/logout.go
@@ -29,8 +29,7 @@ Examples:
 func runLogout(cmd *cobra.Command, _ []string) error {
 	cache, err := client.ResolveTokenCacheForIssuer(config.Client.OIDCIssuer)
 	if err != nil {
-		var ambiguousErr *client.AmbiguousTokenCacheError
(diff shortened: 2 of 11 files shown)


muscariello (approved):
LGTM

PR #1523, opened 2026-05-20
fix(tests): log errors in test suite initializer

csirmazbendeguz:
adding logs to the test suite function since using assertions here causes panic and I would like to know what the error is

### DIFF
--- a/tests/e2e/client/client_suite_test.go
+++ b/tests/e2e/client/client_suite_test.go
@@ -27,11 +27,15 @@ func TestClientE2E(t *testing.T) {
 
 	// Load configuration
 	cfg, err := config.LoadConfig()
-	gomega.Expect(err).NotTo(gomega.HaveOccurred())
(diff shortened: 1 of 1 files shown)


tests/e2e/client/client_suite_test.go
```diff
@@ -27,11 +27,19 @@ func TestClientE2E(t *testing.T) {
 
 	// Load configuration
 	cfg, err := config.LoadConfig()
-	gomega.Expect(err).NotTo(gomega.HaveOccurred())
+	if err != nil {
+		t.Errorf("config.LoadConfig failed: %+v", err)
```
ramizpolic: would `t.Failf` make more sense here?

csirmazbendeguz: I think you mean `t.Fatalf` and yes that could work too

ramizpolic (approved):

PR #1550, opened 2026-05-25
fix(dir): handle mTLS healthchecks in e2e tests

csirmazbendeguz:
while working on the SPIRE test environment (https://github.com/agntcy/csit/pull/196), I noticed the e2e tests only handle insecure authentication at this time

this PR adds mTLS healthchecks to e2e tests' `WaitForGrpcServerReady` function

### DIFF
--- a/tests/e2e/client/client_suite_test.go
+++ b/tests/e2e/client/client_suite_test.go
@@ -48,7 +48,11 @@ func TestClientE2E(t *testing.T) {
 }
 
 var _ = ginkgo.BeforeSuite(func(ctx context.Context) {
-	utils.WaitForGrpcServerReady(ctx, testEnv.Config.ClientOptions.ServerAddress)
+	utils.WaitForGrpcServerReady(
+		ctx,
+		testEnv.Config.ClientOptions.ServerAddress,
+		testEnv.Config.ClientOptions.SpiffeSocketPath,
+	)
 })
 
 var _ = ginkgo.AfterSuite(func() {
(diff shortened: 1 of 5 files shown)


paralta (approved):

ramizpolic:
do we need authentication for gRPC health endpoints? afaik we dont require any auth when doing healtchecks, and thats why we didnt need to handle it. 

@paralta can you please confirm

paralta:
From a quick look at the code, we do skip auth for healthchecks using `IsHealthCheckEndpoint` in authn/x509.go and authz/interceptor.go. @csirmazbendeguz can you please investigate this?

csirmazbendeguz:
It's true that the auth interceptors are skipped for healthcheck requests

The issue is, if authentication is enabled, the server will only accept mTLS connections:

https://github.com/agntcy/dir/blob/8b830b3fedcb7966707e8826c75268af6a25ee40/server/authn/service.go#L233-L239

So if `authn.mode: x509`, the client may only connect with mTLS. In this case, the client is the `WaitForGrpcServerReady` function in the e2e tests. The interceptors will be skipped, yes, but the server will still require mTLS

csirmazbendeguz:
@ramizpolic , does that answer your question, can we proceed ?

ramizpolic:
LGTM, lets merge this

@csirmazbendeguz can you create an issue to remove mTLS and only use TLS instead? this was something we were discussing long ago to remove lots of issues with mTLS due to connection handling (spiffe, k8s ingress, etc)

ramizpolic (approved):

csirmazbendeguz:
> @csirmazbendeguz can you create an issue to remove mTLS and only use TLS instead? this was something we were discussing long ago to remove lots of issues with mTLS due to connection handling (spiffe, k8s ingress, etc)

I see. I suppose we could remove mTLS support

PR #1571, opened 2026-06-02
chore(server)!: drop spiffe mtls-x509 for tls-jwt only authentication

ramizpolic:


ramizpolic:
dropping this PR since mTLS is still useful for internal communication. for the public testbed, we only allow TLS-based transport

PR #1614, opened 2026-06-09
fix(gateway): enable authenticated client usage in gateway

### LINKED ISSUE #1574: [Feature]: Implement an in-process HTTP/REST gateway
### Problem Statement

The AI Finder spec (§3.5, §7) mandates a standard HTTP REST interface for federated discovery, but DIR exposes only gRPC. DIR needs a generic mechanism to expose annotated gRPC services as HTTP/JSON while preserving the existing middleware stack (authn, authz, rate-limit, logging, metrics).

### Proposed Solution

- Add a `gateway` package implementing an in-process grpc-gateway. It dials the gRPC server over loopback so existing interceptors still apply to HTTP requests, and supports raw-bytes responses for export endpoints.
- Add an `HTTPGatewayConfig` (`enabled`, `listen_address`, default `:8889`, disabled by default) wired into `server.Run`, so the sidecar is opt-in and a pre-gateway deployment is unchanged when disabled.
- Provide a `RegisterHandlers` hook so services register themselves (e.g. `RegisterAIFinder`) without the gateway package depending on every service.
- Tests for construction, registration-failure handling, and shutdown.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

ramizpolic:


akijakya (approved):
LGTM

PR #1640, opened 2026-06-15
feat(dir): add SKILL.md file

### LINKED ISSUE #1612: [Feature]: Native Directory SKILL.md
### Problem Statement

This problem is twofold:
- there is no standard DIR skill around the usage of the API/SDK/CLI increasing the usage complexity
- distribution is currently handled through the testbed decreasing the overall usage

### Proposed Solution

- Define a DIR skill.md centered around features, integrations, and tooling (similar to what we have for the MCP server)
- Embed/push this DIR skill.md directly into DIR during service startups so that it is always available. For a fresh/empty DIR that starts, there should be 1 record already availble: DIR Skill.MD. For other existing DIR instances, it will be available directly after version upgrades.

### Alternatives Considered

_No response_

### Additional Context

For the initial support, we can use our existing DIR MCP server and convert it into a single SKILL.md file without all the boilerplate. Try the prompt: `Take this MCP server and convert it into a single SKILL.md file`.

### Checklist

adamtagscherer:
Closes #1612.

On apiserver startup, DIR now embeds its own `SKILL.md` and publishes it as a discoverable OASF 1.0.0 record (`name: agntcy.dir/skill`). Publish runs in a background goroutine — failures are logged but do not block startup. Idempotent on restart by name + version.

### How the SKILL.md content is stored

Stored inline at `modules[0].artifact.data` (base64) on the `core/language_model/agentskills` module.

### How an agent consumes it (and the discovery gap)

The mechanical flow is straightforward:

```
search by name "agntcy.dir/skill"  →  CID
pull CID                            →  OASF record
base64-decode modules[0].artifact.data
verify length == artifact.size, sha256 == artifact.digest
hand the markdown to the LLM as a skill (per the Agent Skills spec)
```

The catch is the very first step: the agent has to *know* to search for the name `agntcy.dir/skill`. This PR doesn't bridge that gap — the calling system (prompt, SDK, or surrounding tool) has to inject the convention.

### DIFF
--- a/server/skill/publisher.go
+++ b/server/skill/publisher.go
@@ -16,22 +16,12 @@ import (
 
 var logger = logging.Logger("server/skill")
 
-// Publish ensures the embedded SKILL.md is available as an OASF record.
-// Idempotent on name+version: a record matching the current build is not
-// re-pushed. Callers should treat errors as non-fatal — DIR must come up
-// even if the skill record cannot be published.
+// Publish pushes the embedded SKILL.md as an OASF record. Idempotent on
+// name+version. Callers must treat errors as non-fatal — DIR has to start
+// even if publishing fails.
 func Publish(ctx context.Context, store types.StoreAPI, db types.DatabaseAPI, validator corev1.Validator) error {
 	if store == nil || db == nil {
 		return errors.New("store and db must be provided")
-	}
-
-	record, err := BuildRecord(time.Now())
-	if err != nil {
-		return fmt.Errorf("build skill record: %w", err)
-	}
-
-	if err := validateRecord(ctx, record, validator); err != nil {
-		return err
 	}
 
 	if existingCID, ok := findExisting(db); ok {
@@ -44,13 +34,22 @@ func Publish(ctx context.Context, store types.StoreAPI, db types.DatabaseAPI, va
 		return nil
 	}
 
+	record, err := BuildRecord(time.Now().UTC())
+	if err != nil {
+		return fmt.Errorf("build skill record: %w", err)
+	}
+
+	if err := validateRecord(ctx, record, validator); err != nil {
+		return err
+	}
+
 	ref, err := store.Push(ctx, record)
 	if err != nil {
 		return fmt.Errorf("push skill record: %w", err)
 	}
 
-	// Mirror the gRPC store controller: keep the search index in sync so the
-	// record is discoverable immediately, without waiting for an external push.
+	// Update the search index in line with the gRPC store controller, so the
+	// record is discoverable without waiting for an external push.
 	decoded, decodeErr := record.Decode()
 	if decodeErr != nil {
 		logger.Warn("DIR skill record pushed but could not be decoded for search index",
@@ -68,7 +67,7 @@ func Publish(ctx context.Context, store types.StoreAPI, db types.DatabaseAPI, va
 		"cid", ref.GetCid(),
 		"name", RecordName,
 		"version", RecordVersion(),
-		"sha256", ContentSHA256(),
+		"digest", ContentDigest(),
 	)
 
 	return nil
@@ -98,8 +97,8 @@ func findExisting(db types.DatabaseAPI) (string, bool) {
 		types.WithLimit(1),
 	)
 	if err != nil {
-		// Treat as "not found"; a re-push of unchanged content is harmless
-		// because the store is content-addressed.
+		// Fall through and re-push; the store is content-addressed so an
+		// unchanged record is a no-op.
 		logger.Debug("skill record lookup failed, will attempt to publish anyway", "error", err)
 
 		return "", false
--- a/server/skill/publisher.go
+++ b/server/skill/publisher.go
@@ -1,0 +1,113 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package skill
+
+import (
+	"context"
+	"errors"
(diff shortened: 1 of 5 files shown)


server/skill/publisher.go
```diff
@@ -0,0 +1,113 @@
...
+// Publish ensures the embedded SKILL.md is available as an OASF record.
+// Idempotent on name+version: a record matching the current build is not
+// re-pushed. Callers should treat errors as non-fatal — DIR must come up
+// even if the skill record cannot be published.
+func Publish(ctx context.Context, store types.StoreAPI, db types.DatabaseAPI, validator corev1.Validator) error {
+	if store == nil || db == nil {
+		return errors.New("store and db must be provided")
+	}
+
+	record, err := BuildRecord(time.Now())
```
akijakya: Comment above BuildRecord says `time.Now().UTC()` should be used

server/skill/publisher.go
```diff
@@ -0,0 +1,113 @@
...
+
+	if existingCID, ok := findExisting(db); ok {
+		logger.Debug("DIR skill record already present, skipping publish",
+			"cid", existingCID,
+			"name", RecordName,
+			"version", RecordVersion(),
+		)
+
+		return nil
+	}
```
akijakya, line 35: This could be placed before before building & validating the record so we can return earlier

server/skill/skill.go
```diff
@@ -0,0 +1,136 @@
...
+// BuildRecord constructs the OASF record. `now` is injected for deterministic
+// testing; production callers should pass time.Now().UTC().
+func BuildRecord(now time.Time) (*corev1.Record, error) {
+	recordVersion := RecordVersion()
+	contentBytes := []byte(markdown)
+
+	digest := "sha256:" + ContentSHA256()
+	encodedPayload := base64.StdEncoding.EncodeToString(contentBytes)
+
+	doc := map[string]any{
```
akijakya: Could we use the output of the skillToRecord translator function of OASF SDK here?

server/skill/skill.go
```diff
@@ -0,0 +1,136 @@
...
+		// `domains` is Recommended, not Required; emit empty to silence the
+		// validator warning without binding to a specific domain.
+		"domains": []any{},
+		"skills": []map[string]any{
+			{
+				"name": "natural_language_processing/natural_language_understanding/contextual_comprehension",
+				"id":   skillContextualComprehensionID,
+			},
+		},
+		"modules": []map[string]any{
```
akijakya: Could we also add an MCP module for the directory MCP server here, so we have both in one place?

server/skill/skill.go
```diff
@@ -0,0 +1,136 @@
...
+		return nil, fmt.Errorf("marshal skill record: %w", err)
+	}
+
+	record, err := corev1.UnmarshalRecord(raw)
+	if err != nil {
+		return nil, fmt.Errorf("decode skill record: %w", err)
+	}
+
+	return record, nil
+}
```
ramizpolic, line 162: we should use explicit types to define records in the code. this is prone to errors due to free-form definition, lets use proper things we already have in the code:

```suggestion
// DirSkill is the content of the Directory SKILL.md file, which describes to use DIR
// tooling for Agent Skills integrations.
//
//go:embed SKILL.md
var DirSkill string

// DirRecord is the OASF record describing how an AI agent can interact with this Directory instance.
// It includes all the information, instructions, and artifacts for out-of-box usage.
// It is published to the catalog at startup and can be pulled by agents to learn how to use this Directory.
var DirRecord = corev1.New(&typesv1.Record{
	Name:          "org.agntcy/directory",
	SchemaVersion: "1.0.0",
	Version:       version.String(),
	Description:   "Self-describing usage guide for this AGNTCY Directory (DIR) instance: how an AI agent can publish, search, pull, sign, and verify OASF agent records.",
	Authors:       []string{"https://github.com/agntcy/dir"},
	CreatedAt:     time.Now().UTC().Format(time.RFC3339),
	Skills: []*typesv1.Skill{
		{Name: "natural_language_processing/natural_language_understanding/contextual_comprehension"},
	},
	Modules: []*typesv1.Module{
		// TODO: add MCP module
		// Skill module
		{
			Name: "core/language_model/agentskills",
			Data: &structpb.Struct{
				Fields: map[string]*structpb.Value{
					"skill_file": structpb.NewStringValue("SKILL.md"),
					"skill_manifest": structpb.NewStructValue(&structpb.Struct{
						Fields: map[string]*structpb.Value{
							"name":        structpb.NewStringValue("agntcy-dir"),
							"description": structpb.NewStringValue("Use this skill to interact with an AGNTCY Directory (DIR) instance."),
							"version":     structpb.NewStringValue(version.String()),
						},
					}),
				},
			},
			Artifact: &typesv1.Descriptor{
				MediaType: catalogv1.ProtocolAgentSkillsMdMediaType,
				Size:      uint64(len(DirSkill)),
				Digest:    digest.FromString(DirSkill).String(),
				Data:      []byte(base64.StdEncoding.EncodeToString([]byte(DirSkill))),
			},
		},
	},
})
```

server/skill/skill.go
adamtagscherer: I went with @ramizpolic's suggestion, because:
- the translator emits `*structpb.Struct` which is exactly what he flagged
- it emits `skills: []` and we'd still need to tinker with it afterwards

ramizpolic (approved):

akijakya (approved):
💯

PR #1708, opened 2026-06-26
feat(search): track record usage metrics and add popularity sort modes

### LINKED ISSUE #1695: [Feature]: Track download counts and provider distribution to power the most-popular sort mode
### Problem Statement

There is no signal for how valuable or widely used a record is. No usage counters exist, and there is no way to distinguish a widely-used record from one that was just pushed. Well-used records cannot be surfaced first, and the direction server has no data to back a "most popular" sort mode.

### Proposed Solution

Track real usage across the gRPC operations the directory exposes, persist counters in a dedicated DB table, and use them to power "most popular":

- **Usage signals to track:**
  - **Pulls / downloads** — `pull_count` is incremented directly in `StoreService.Pull` (store controller) on each successful pull. Sync operations go through the OCI layer and bypass the gRPC controller entirely, so they are never counted.
  - **Provider count** — the number of directories (local node + remote peers) currently announcing a given record CID in the DHT. A record hosted by many nodes has been pulled and retained by many independent parties — a strong organic popularity signal. Updated periodically by the **reconciler**, which queries the routing layer for each locally known CID and persists the result as a point-in-time gauge.
  - **Lookups** — `StoreService.Lookup` (metadata-only interest; minor signal, lower weight than pulls).
- **Persist counters in a `record_usage_metrics` DB table**, separate from the main record index. Stores cumulative counters (`pull_count`, `lookup_count`) and the reconciler-computed `provider_count` gauge, plus `last_used_at`.
- **Introduce a "Usage metrics" referrer type** — a structured referrer attached to each record that carries the per-record counters. This allows usage data to travel with the record when synced: when a directory syncs a record from a peer, it copies the peer's usage metrics referrer, giving nodes a baseline popularity signal for records they did not originate.
- **Combine into a popularity value** (a weighted blend of pull count, provider count, and optionally lookup count, time-decayed) that backs the "most popular" sort mode.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

paralta:
This PR introduces a `record_usage_metrics` table that accumulates per-record usage signals (`pull_count`, `lookup_count`, `provider_count`) and exposes them as sort modes on `SearchService`. 
* Pull counts are incremented directly in `StoreService.Pull`; lookup counts in `StoreService.Lookup`. 
* Provider count — the number of DHT peers currently announcing a CID — is a point-in-time gauge refreshed by a new `metrics` reconciler task. Because the routing layer uses an embedded Badger store that only supports a single writer, the standalone reconciler queries provider count over a dedicated `RoutingService.GetProviderCount` gRPC RPC rather than reading the datastore directly; it requires `server_address` to be configured for this task to run.


proto/agntcy/dir/routing/v1/routing_service.proto
```diff
@@ -44,6 +44,11 @@ service RoutingService {
   // that match the given parameters.
   // This operation does not interact with the network.
   rpc List(ListRequest) returns (stream ListResponse);
+
+  // GetProviderCount returns the number of distinct peers (including this node)
+  // currently announcing the given CID in the routing layer.
+  // Used by the reconciler to populate the provider_count popularity signal.
+  rpc GetProviderCount(GetProviderCountRequest) returns (GetProviderCountResponse);
```
ramizpolic, line 51: lets not expose this part through the API, we can already access the RoutingAPI through the Go interface, we dont need to do it publically. Only the reconciler needs the method exposed by the go code, no other components need it

reconciler/config/config.go
```diff
@@ -44,6 +45,14 @@ type Config struct {
 	// LocalRegistry holds configuration for the local OCI registry.
 	LocalRegistry ociconfig.Config `json:"local_registry" mapstructure:"local_registry"`
 
+	// ServerAddress is the gRPC address of the apiserver (e.g. "localhost:8888").
+	// Required by the metrics task when the reconciler runs as a standalone process,
+	// because the routing layer (Badger datastore) is embedded in the server and
+	// cannot be shared across process boundaries. The metrics task calls
+	// RoutingService.GetProviderCount over gRPC instead.
+	// Leave empty in daemon mode — the in-process routing API is used directly.
+	ServerAddress string `json:"server_address" mapstructure:"server_address"`
```
ramizpolic, line 55: AFAIK we are persisting the data, read-only access should be possible to badger instance between separate processes

proto/agntcy/dir/routing/v1/routing_service.proto
paralta: That makes sense for the daemon case where the reconciler runs in the same process and uses the Go interface directly. However, when the reconciler runs as a standalone process (docker compose, kind), it has no access to the routing layer (check the datastore comment in the description). 

Without this RPC, the standalone reconciler has no way to query provider counts.

reconciler/config/config.go
paralta: This read-only Badger approach works on the same filesystem, but in a containerized deployment (kind) the apiserver and reconciler run in separate pods with isolated filesystems.

ramizpolic: you are right, it does not support concurrency. a separate problem that we have with [badger](https://github.com/ipfs/go-ds-badger) right now is that is no longer maintained (https://github.com/ipfs/kubo/issues/11186), so we will need to eventually switch to something else (luckily our datastore is only a cache, so no migration will be needed).

alternative to consider: github.com/ipfs/go-ds-pebble

i tried to do a concurrent read-write with pebble, there are some hacky solutions, but lets keep this approach you added

ramizpolic: **one important note**: if the server is running with auth, we need to ensure that reconciler uses authenticated client to connect with the server, otherwise this wont work

ramizpolic: this was similar to an issue we had with HTTP gateway that was fixed here https://github.com/agntcy/dir/pull/1614

reconciler/tasks/metrics/grpc.go
```diff
@@ -0,0 +1,54 @@
...
+// reconciler process, which has no direct access to the routing layer.
+type GRPCProviderCounter struct {
+	client routingv1.RoutingServiceClient
+	conn   *grpc.ClientConn
+}
+
+// NewGRPCProviderCounter dials the apiserver at addr and returns a
+// GRPCProviderCounter. The caller is responsible for calling Close when done.
+func NewGRPCProviderCounter(addr string) (*GRPCProviderCounter, error) {
+	conn, err := grpc.NewClient(addr, grpc.WithTransportCredentials(insecure.NewCredentials()))
```
ramizpolic: this will not work, check fixes in https://github.com/agntcy/dir/pull/1614 for a similar issue we had

ramizpolic (changes requested):

reconciler/tasks/metrics/grpc.go
paralta: thanks for catching this! fixing now

ramizpolic: the solution can be adjusted and fully reused for the current code as well (adjusting the GetClientOpts to be a `authn` package function rather than a method on the authn service)

ramizpolic (approved):

PR #1732, opened 2026-06-30
refactor(config): introduce config module

### LINKED ISSUE #1490: [Feature]: Deduplicate shared infrastructure settings in daemon config (OCI registry, database)
### Problem Statement

The reference daemon configuration (`cli/cmd/daemon/daemon.config.yaml`) currently forces operators to declare the same infrastructure settings in two places:

1. **OCI registry**
   - `server.store.oci` (lines 26-34) — used by the apiserver to read/write records
   - `reconciler.local_registry` (lines 78-82) — used by the reconciler tasks (regsync, indexer, signature, name) to read those same records

2. **Database**
   - `server.database` (lines 45-53) — used by the apiserver for its search/sync state
   - `reconciler.database` (lines 83-86) — used by the reconciler to operate on the same state the apiserver writes

In the default single-process daemon topology these two pairs **must** point at the same registry and the same database, but nothing in the schema, defaults, or loader enforces it. The result:

- **Silent drift.** Updating one side (e.g. `server.store.oci.registry_address` or `server.database.postgres.host`) without the other lets the daemon start cleanly while the reconciler reads from a different — often empty — registry/database. Records appear to be written but are never indexed, signed, or synced.
- **Doubled defaults.** `registerServerDefaults` and `registerReconcilerDefaults` in `cli/cmd/daemon/config.go` set the same constants twice.
- **Doubled env vars.** One logical setting requires two overrides, e.g. `DIRECTORY_DAEMON_SERVER_STORE_OCI_REGISTRY_ADDRESS` **and** `DIRECTORY_DAEMON_RECONCILER_LOCAL_REGISTRY_REGISTRY_ADDRESS`; same for every database field (host, port, credentials, ssl_mode, …).
- **Mirrored everywhere.** The same duplication lives in every E2E config (`tests/e2e/**`) and Helm chart (`install/charts/dir/**`), multiplying the surface area for drift.

The result is an error-prone configuration model where the only path that the operator actually wants to express ("the daemon's registry and database") is not first-class, and the wrong path (independent server/reconciler settings) is the default.

### Proposed Solution

Introduce a single source of truth for the shared infrastructure settings in the daemon config and have the `server` and `reconciler` sections inherit from it.

### 1. New top-level `shared` section (default)

Add a top-level section to `daemon.config.yaml` that holds the settings the two services must agree on:

```yaml
shared:
  registry:
    local_dir: "store"            # used by embedded zot if set
    registry_address: "localhost:5555"
    repository_name: "dir"
    auth_config:
      insecure: true
  database:
    type: "sqlite"
    sqlite:
      path: "dir.db"
    postgres:
      host: "localhost"
      port: 5432
      database: "dir"
      ssl_mode: "disable"
```

The daemon's config loader (`cli/cmd/daemon/config.go::loadConfig`) is responsible for projecting `shared.registry` into both `server.store.oci` and `reconciler.local_registry`, and projecting `shared.database` into both `server.database` and `reconciler.database`, **after** reading the file but **before** unmarshalling into `DaemonConfig`. This keeps the in-process service config structs (from `server/config` and `reconciler/config`) untouched, so this change is purely additive at the daemon CLI layer.

### 2. `shared.*` overwrites the corresponding per-service fields

`server.store.oci`, `server.database`, `reconciler.local_registry`, and `reconciler.database` remain first-class struct fields. `shared.*` is an override that the daemon's loader fans out into both locations before unmarshalling into `DaemonConfig`:

- `shared.registry` set → overwrites `server.store.oci` and `reconciler.local_registry`.
- `shared.database` set → overwrites `server.database` and `reconciler.database`.
- `shared.*` unset → per-service fields are used as-is; if the two halves disagree, the loader refuses to start with an error naming both keys (the current architecture requires them to match).

### 3. Single env-var path for shared settings

Bind the shared keys so a single environment variable updates both services:

- `DIRECTORY_DAEMON_SHARED_REGISTRY_REGISTRY_ADDRESS`
- `DIRECTORY_DAEMON_SHARED_REGISTRY_AUTH_CONFIG_USERNAME` / `_PASSWORD`
- `DIRECTORY_DAEMON_SHARED_DATABASE_TYPE`
- `DIRECTORY_DAEMON_SHARED_DATABASE_POSTGRES_HOST` / `_PORT` / `_USERNAME` / `_PASSWORD` / `_DATABASE` / `_SSL_MODE`
- `DIRECTORY_DAEMON_SHARED_DATABASE_SQLITE_PATH`

The existing `DIRECTORY_DAEMON_SERVER_STORE_OCI_*`, `DIRECTORY_DAEMON_SERVER_DATABASE_*`, `DIRECTORY_DAEMON_RECONCILER_LOCAL_REGISTRY_*`, and `DIRECTORY_DAEMON_RECONCILER_DATABASE_*` variables continue to work and follow the same override semantics as the file-level keys: when the matching `SHARED_*` variable is set, it overwrites both per-service fields.

### 4. Validation at load time

- `shared.*` set → it overwrites both per-service fields silently.
- `shared.*` not set, per-service halves agree → use them as-is.
- `shared.*` not set, per-service halves disagree (e.g. `server.store.oci.registry_address` ≠ `reconciler.local_registry.registry_address`) → refuse to start with an explicit error naming both keys; this configuration cannot work in the current architecture.

### 5. Clean up the reference config and tests

- Update `cli/cmd/daemon/daemon.config.yaml` to use the new `shared` section as the documented default and drop the duplicated `reconciler.local_registry` / `reconciler.database` blocks from the reference file (the fields still exist on the struct; they just no longer need to be repeated in the YAML).
- Update the test fixtures under `tests/e2e/**` and the Helm values in `install/charts/dir/` to use `shared.*`, removing the duplication at the source.
- Update the env-var examples in the header comment of `daemon.config.yaml` (lines 10-18) to point at the new `SHARED_*` variables.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

paralta:
This PR establishes a new `config/` Go module as the single source of truth for all configuration types across the monorepo. Previously, each service (apiserver, reconciler, daemon) owned fragmented config packages — `server/config`, `server/database/config`, `server/store/oci/config`, `reconciler/config, and a dozen sub-service `config.go` files — that duplicated type definitions, viper wiring, and defaults. All of those are now deleted and replaced by the `config` module plus a `config/reconciler sub-package for reconciler task types.


paralta:
These changes will not be included in https://github.com/agntcy/dir/milestone/14, thus keeping this PR as draft.

paralta:
https://github.com/agntcy/dir/issues/1490#issuecomment-5451742722

PR #1746, opened 2026-07-03
feat(utils/scanner): add SkillRunner to scan agent skill bundles

### LINKED ISSUE #1724: [Feature]: Add skill-scanner support
### Problem Statement

Records with OASF `skills` fields are never scanned today, even though [`skill-scanner`](https://github.com/cisco-ai-defense/skill-scanner) exists for exactly this purpose. Skill-based agents represent a different attack surface from MCP servers, and leaving them unscanned is a security blind spot.

### Proposed Solution

Add a `SkillRunner` implementation to `utils/scanner`. Scanning logic auto-detects which runner(s) to invoke based on record content: `MCPRunner` runs when the record has MCP extensions, `SkillRunner` runs when the `skills` field is non-empty. Both can run for the same record. Results are tagged by scanner type so they can be stored and surfaced independently.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

paralta:
This PR implements `SkillRunner` in `utils/scanner`, the counterpart to the existing `MCPRunner`, completing security scanning support for records that carry agent skill artifacts. The runner extracts the `core/language_model/agentskills` module from a record, unpacks it via `exportfmt.ExtractSkillBundleArchive`, and runs `skill-scanner scan` to produce a `ScanResult` with per-finding severity. `SKILL_SCANNER_LLM_*` environment variables are derived automatically from the `AZURE_OPENAI_*` equivalents so CI doesn't need duplicate config.

- `ExtractSkillBundleArchive` is extended to transparently handle both gzip-tar bundles and plain-text `SKILL.md` manifests via a magic-byte check, eliminating format-detection logic from the scanner itself.
- `utils/go.mod` switches from the pinned `api v1.5.0` to a local `replace` directive so the scanner can use the updated `exportfmt` package before a new tag is cut.
- `trimToSkillJSON` looks for either `{` or `[` as the JSON start, unlike the MCP equivalent which only accepts arrays — skill-scanner's single-scan mode emits a top-level object.
- Safety verdict is taken directly from the scanner's `is_safe` field; findings are always surfaced regardless of that verdict so callers have full context.


ramizpolic (approved):
LGTM

PR #1757, opened 2026-07-06
feat(security): add automated security scanning, scan search filters, and pull support

paralta:
This PR introduces end-to-end security scanning for records in the directory. A new reconciler task periodically runs `mcp-scanner` and `skill-scanner` against unscanned records, persisting each scanner's result both as an OCI referrer (full report with findings) and as a denormalized `scan_reports` DB row (for efficient search queries). The server also upserts the DB row immediately when a `ScanReport` referrer is pushed directly via the API, so the search index reflects external scan results without waiting for the next reconciler cycle.

Two new search filters are exposed on `dirctl search`: `--safe` (all scanners reported `is_safe=true`, analogous to `--trusted` for signatures) and `--scan-severity <threshold>` (max severity ≥ the given level). There is intentionally no bare "has been scanned" filter — absence of a scan result is not semantically meaningful in the same way as a clean or dirty result. `dirctl pull --scan-report` fetches the full scan reports for a record.

- The `scan_reports` table uses a composite primary key `(record_cid, scanner_type)` so each runner's result is tracked independently — a record can be safe per MCP and unsafe per skill simultaneously.
- The indexer task is fixed to silently skip referrer blobs (whose CIDs appear as OCI tags) rather than logging them as indexing errors.
- Scanner CLIs (`mcp-scanner`, `skill-scanner`) are installed via `uv` and added to `Taskfile.deps.yml` as shared tasks; the daemon config enables the scan task by default.

PR #1760, opened 2026-07-06
feat(gateway): surface security scan results in the AI Finder UI

### LINKED ISSUE #1731: [Feature]: AI Finder UI with scanned badge and findings
### Problem Statement

The AI Finder UI gives users no security signal when browsing agents. Users cannot tell whether a record has been security-scanned, which scanner ran, or what findings were produced — making it impossible to make an informed deployment decision from the UI.

### Proposed Solution

Surface scan metadata on record cards and in the search panel:

- **Card chip**: display a "Security Scanned" chip when `scan_reports` is non-empty. Severity drives the chip color: green (NONE/LOW), amber (MEDIUM), red (HIGH/CRITICAL).
- **Findings drawer**: clicking the chip opens a side drawer listing scanner type, scanned timestamp, and each finding (rule ID, severity, message).
- **Search panel filter**: add a "Scanned" toggle and a severity dropdown, wired to `SCANNED` and `SCAN_SEVERITY` from S6.

The gateway layer attaches scan report summaries to record list responses via a lightweight DB lookup (not an OCI fetch), so no extra round-trips are needed per card.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

paralta:
This PR wires the reconciler's security scan results into the AI Finder UI. Scan summaries stored in the database are now preloaded alongside catalog entries and embedded into each record's `metadata.scanManifest` field — using the existing `map<string, Value>` proto field, so no schema changes are needed. The UI consumes this to show a severity-colored "Security Scanned" / "Issues Found" chip inline on each card title, a per-scanner breakdown in the detail modal, and a "Safe" checkbox filter in the sidebar.

- Scan metadata is embedded at the catalog query layer rather than fetched separately, because `PullReferrer` is a gRPC bidirectional stream with no HTTP gateway binding and therefore unreachable from the browser.
- The Status filter is now checkboxes (not radio buttons) because a record can simultaneously be trusted and verified — the two are not mutually exclusive. "Safe" is co-located in the same Status group rather than a separate Security section.
- The scan badge color encodes the highest reported severity across all scanners (green for safe/none, blue for low, amber for medium, orange for high, red for critical), with a tooltip detailing scanner count or max severity.

image

image


server/gateway/ui/src/lib/components/ScanBadge.svelte
```diff
@@ -0,0 +1,37 @@
...
+			case 'CRITICAL': return 'bg-red-100 text-red-700';
+			case 'HIGH': return 'bg-orange-100 text-orange-700';
+			case 'MEDIUM': return 'bg-amber-100 text-amber-700';
+			case 'LOW': return 'bg-blue-100 text-blue-700';
+			default: return 'bg-emerald-100 text-emerald-700';
+		}
+	}
+
+	let colorClass = $derived(severityColorClass(scan.maxSeverity));
+	let label = $derived(scan.isSafe ? 'Security Scanned' : 'Issues Found');
```
ramizpolic, line 26: can we maybe have this card display something like: 

`Security Posture: No issues found` or `Security Posture: X issues found`

the other part is that it looks a bit strange the title of the card like: `Security scanned Security Scan Results`. This would be nicer a bit more unified like above

server/gateway/ui/src/lib/components/AICard.svelte
```diff
@@ -28,11 +30,14 @@
...
 		<div class="min-w-0">
-			<div class="flex items-center gap-1.5 min-w-0">
+			<div class="flex items-center gap-1.5 min-w-0 flex-wrap">
 				<h3 class="font-semibold text-ink-strong truncate">{aicard.displayName || 'Unnamed AI card'}</h3>
 				{#if verified}
 					<VerifiedBadge />
 				{/if}
+				{#if scanManifest}
+					<ScanBadge scan={scanManifest} />
+				{/if}
```
ramizpolic, line 40: can we put this below instead or below the stars/rating? seems too packed as part of the title

we can use the footer of the card on the dashboard to put extra things like security, metrics, and other details

api/catalog/v1/catalog.go
```diff
@@ -341,6 +364,57 @@ func catalogSignatures(cid string, signatures []coretypes.ObjectSignature) *Trus
...
+	val, err := buildScanManifestValue(reports)
+	if err != nil {
+		return
+	}
+
+	if entry.Metadata == nil {
+		entry.Metadata = make(map[string]*structpb.Value)
+	}
+
+	entry.Metadata["scanManifest"] = val
```
ramizpolic: the metadata keys per AI catalog spec require a bit more specific prefix to indicate the protocol that is injecting it. i would maybe consider having something like the protobuf name here of the object, ie. `agntcy.dir.security.v1.ScanResult`

ramizpolic (approved):
LGTM

api/catalog/v1/catalog.go
```diff
@@ -66,8 +67,17 @@ var catalogModules = map[string]catalogModuleProjection{
 	},
 }
 
+// ScanReportSummary summarizes one scanner's result for embedding in catalog metadata.
+type ScanReportSummary interface {
+	GetScannerType() string
+	GetIsSafe() bool
+	GetMaxSeverity() string
+	GetUpdatedAt() time.Time
+}
```
ramizpolic, line 76: can we have the reason for the severities?

server/gateway/ui/src/lib/components/ScanBadge.svelte
paralta: Let me improve this, agreed 👍 the MCP behaviour scanner does not return findings (just `safe` or `not safe`) so `Security Posture: X issues found` is not a good fit here.

server/gateway/ui/src/lib/components/AICard.svelte
paralta: My plan was to add the badge close to the trusted badge, but it turned out a bit bloated indeed

api/catalog/v1/catalog.go
paralta: The scanners currently return a list of raw findings which might lead to an extensive text for a new `reason` field, which would be hard to visualise in this UI. Check the skill example at https://github.com/agntcy/dir/pull/1757#issuecomment-4892001269.

paralta:
Screenshot 2026-07-07 at 15 41 42
Screenshot 2026-07-07 at 15 41 50
Screenshot 2026-07-07 at 15 44 50

@ramizpolic wdyt about these updates?

ramizpolic:
@paralta looks better! my slight concern is that it may get mixed up with verifiable status 

for next iteration, we can include a filter in the search like `include=security-details` to include the messages as well since right now we can not inspect the details of the security scan, but its enough for v1

PR #1763, opened 2026-07-07
feat(dir): batch install

### LINKED ISSUE #1742: feat(cli): batch install for `dirctl install` (search-filter selection, like pull/export)
## Summary

Add a batch mode to `dirctl install` that selects multiple records via search filters and installs each into the detected AI coding agents — mirroring the batch mode `dirctl pull` and `dirctl export` already have.

`dirctl install` (#1737) is currently single-record only: it takes one `<cid-or-name[:version][@digest]>` positional. `pull` and `export` gained a search-filter batch mode in #1707, backed by the shared `cli/util/records` package (`SearchAndPull`, `CollectCIDs`, `LatestByName`, `SanitizeName`). Batch install should reuse that same backbone.

## Motivation

Installing a set of records (e.g. "every record carrying an MCP module", or a naming glob) currently means running `dirctl install <name>` N times. A batch mode makes it one command, consistent with `pull`/`export`.

## Proposed design

- **Trigger:** search filters present and **no positional argument** (mutually exclusive, exactly like `pull`/`export`). Note `--output-dir` does **not** apply — `install` writes into agent configs, not files — so batch mode is keyed purely on "filters given, no CID/name".

  ```
  dirctl install --module integration/mcp --name "web*" --agents all
  ```

- **Record selection:** reuse `records.SearchAndPull(ctx, c, search.BuildQueries(&filters), limit)` then `records.LatestByName` by default, with `--all-versions` to keep every version. Register `search.RegisterFilterFlags` + `--limit` on the command.

- **Per record:** run the existing derive → install flow into the selected agents; aggregate a single plan and a single summary across all records; one confirmation for the whole batch (`--yes` to skip).

- **No-installable-module / a2a-only records in a batch:** unlike single-install (which hard-errors), a batch should **skip such records with a warning** so one unsuitable record does not abort the whole run. The summary should list what was skipped and why.

- **Identity is still per-record:** each record keeps its own slug-derived skill folder/file/managed-block and its MCP server key(s), so batch-installed records coexist exactly as if installed one by one (see #1737).

## Cleanup to fold in

- `cli/cmd/install`'s `sanitizeSlug` overlaps with `records.SanitizeName`. Point the slug at the shared helper (keeping the extra leading/trailing `.`/`-` trim that guards against `.`/`..` path traversal) to avoid drift.
- Batch plans can be large (N records × M agents) — group the plan/summary output by record for readability.

## Acceptance criteria

- `dirctl install --<filter> …` (no positional) installs all matched records' artifacts into the detected agents; positional + filters are mutually exclusive.
- `--all-versions` keeps every version; default is latest-per-name.
- Records with no installable module (or a2a-only) are skipped-with-warning in batch mode, not fatal; reported in the summary.
- Single confirmation for the whole batch; `--dry-run` and `--yes` honored.
- Record selection reuses `cli/util/records`; no duplicated search/pull logic.
- Unit tests for: filter→records selection, latest-per-name vs `--all-versions`, skip-with-warning for unsuitable records, aggregated plan/summary.

## Dependencies / references

- Builds on `dirctl install` (#1737).
- Reuses the batch backbone from #1707 (`cli/util/records`).

csirmazbendeguz:
If install receives filters, use those filters to fetch the records and install them

If a tool doesn't support skill directories, skip installing

Note, this PR changes the Copilot and Cursor paths so skills are stored in directories instead of markdown files

### DIFF
--- a/api/exportfmt/skill_bundle_extract.go
+++ b/api/exportfmt/skill_bundle_extract.go
@@ -6,14 +6,178 @@ package exportfmt
 import (
 	"archive/tar"
 	"bytes"
-	"compress/gzip"
 	"fmt"
-	"io"
+	"io/fs"
 	"os"
 	"path/filepath"
 )
 
 const skillManifestFile = "SKILL.md"
+
+// SkillMarkdownFromArchive returns the SKILL.md content from a skill artifact.
+// Plain-text archives are returned as-is; gzip tar bundles must contain SKILL.md.
+func SkillMarkdownFromArchive(archive []byte) (string, error) {
+	if len(archive) == 0 {
+		return "", fmt.Errorf("skill bundle archive is empty")
+	}
+
+	if !isGzipArchive(archive) {
+		return string(archive), nil
+	}
+
+	iterator, err := NewTarIterator(archive, WithTypeflag(tar.TypeReg))
+	if err != nil {
+		return "", fmt.Errorf("invalid gzip archive: %w", err)
+	}
+
+	for entry, err := range iterator {
+		if err != nil {
+			return "", fmt.Errorf("read tar entry: %w", err)
+		}
+
+		rel, err := localTarEntryPath(entry.header.Name)
+		if err != nil {
+			return "", fmt.Errorf("invalid tar entry %q: %w", entry.header.Name, err)
+		}
+
+		if rel != skillManifestFile {
+			continue
+		}
+
+		return string(entry.payload), nil
+	}
+
+	return "", fmt.Errorf("archive does not contain %q", skillManifestFile)
+}
+
+// SkillBundleMatchesDir reports whether dir already contains the same files as archive.
+func SkillBundleMatchesDir(archive []byte, dir string) (bool, error) {
+	if len(archive) == 0 {
+		return false, fmt.Errorf("skill bundle archive is empty")
+	}
+
+	if !isGzipArchive(archive) {
+		return matchesSkillManifest(dir, archive)
+	}
+
+	return matchesGzip(dir, archive)
+}
+
+func matchesSkillManifest(dir string, b []byte) (bool, error) {
+	content, err := os.ReadFile(filepath.Join(dir, skillManifestFile))
+
+	if os.IsNotExist(err) {
+		return false, nil
+	}
+
+	if err != nil {
+		return false, fmt.Errorf("read %s: %w", skillManifestFile, err)
+	}
+
+	return bytes.Equal(content, b), nil
+}
+
+func matchesGzip(dir string, b []byte) (bool, error) {
+	iterator, err := NewTarIterator(b, WithTypeflag(tar.TypeReg))
+	if err != nil {
+		return false, fmt.Errorf("invalid gzip archive: %w", err)
+	}
+
+	paths := make(map[string]bool)
+
+	for entry, err := range iterator {
+		if err != nil {
+			return false, fmt.Errorf("read tar entry: %w", err)
+		}
+
+		path, err := localTarEntryPath(entry.header.Name)
+		if err != nil {
+			return false, fmt.Errorf("invalid tar entry %q: %w", entry.header.Name, err)
+		}
+
+		match, err := matchesTarEntry(dir, entry)
+		if err != nil {
+			return false, err
+		}
+
+		if !match {
+			return false, nil
+		}
+
+		paths[path] = true
+	}
+
+	if len(paths) == 0 {
+		return false, fmt.Errorf("archive contains no regular files")
+	}
+
+	match, err := matchesPaths(dir, paths)
+	if err != nil {
+		return false, err
+	}
+
+	if !match {
+		return false, nil
+	}
+
+	return true, nil
+}
+
+func matchesPaths(dir string, paths map[string]bool) (bool, error) {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return false, nil
+	} else if err != nil {
+		return false, fmt.Errorf("stat %s: %w", dir, err)
+	}
+
+	result := true
+
+	err := filepath.WalkDir(dir, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if d.IsDir() {
+			return nil
+		}
+
+		rel, err := filepath.Rel(dir, path)
+		if err != nil {
+			return fmt.Errorf("rel path for %q: %w", path, err)
+		}
+
+		if _, ok := paths[rel]; !ok {
+			result = false
+
+			return fs.SkipAll
+		}
+
+		return nil
+	})
+	if err != nil {
+		return false, fmt.Errorf("walk %s: %w", dir, err)
+	}
+
+	return result, nil
+}
+
+func matchesTarEntry(dir string, entry *TarEntry) (bool, error) {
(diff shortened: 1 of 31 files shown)


api/exportfmt/skill_bundle_extract.go
```diff
@@ -6,15 +6,121 @@ package exportfmt
...
+
+	if !isGzipArchive(archive) {
+		return matchesSkillManifest(dir, archive)
+	}
+
+	return matchesGzip(dir, archive)
+}
+
+func matchesSkillManifest(dir string, b []byte) (bool, error) {
+	content, err := os.ReadFile(filepath.Join(dir, skillManifestFile))
```
akijakya, line 67: **Latent bug: fresh install fails on the plain-text branch.**

This returns an *error* (not `false`) when `SKILL.md` doesn't exist yet. `InstallSkillBundle` treats any error from `SkillBundleMatchesDir` as fatal (`failOutcome`), so a first-time install via this branch would fail instead of proceeding to `ActionAdded`.

The gzip branch (`matchesTarEntry`) handles this correctly with `if os.IsNotExist(err) { return false, nil }`. This branch should match:

```go
content, err := os.ReadFile(filepath.Join(dir, skillManifestFile))
if os.IsNotExist(err) {
	return false, nil
}
if err != nil {
	return false, fmt.Errorf("read %s: %w", skillManifestFile, err)
}
```

Currently latent (only reachable if `arts.skillBundle` ever holds non-gzip content — today the bundle formatter always emits gzip), but the two branches of a function documented to handle both formats disagree on the missing-file case.

api/exportfmt/skill_bundle_extract.go
```diff
@@ -6,15 +6,121 @@ package exportfmt
...
+	if err != nil {
+		return false, fmt.Errorf("read %s: %w", skillManifestFile, err)
+	}
+
+	return bytes.Equal(content, b), nil
+}
+
+func matchesGzip(dir string, b []byte) (bool, error) {
```
akijakya, line 80: **Update idempotency doesn't detect or clean stale files.**

`matchesGzip` verifies every *archive* file exists and matches on disk, but never checks for on-disk files that are **not** in the archive. Combined with `ExtractSkillBundleArchive` extracting over the existing directory without clearing it first, this means: if a skill update removes a file, the stale file lingers — and since the remaining files still match, `SkillBundleMatchesDir` can even report `ActionUnchanged` incorrectly.

Uninstall is fine (`removeSkillFolder` does `os.RemoveAll` on the owned slug dir). Consider clearing the dest dir before extract, or documenting the limitation.

cli/internal/agentcfg/registry.go
```diff
@@ -44,10 +44,9 @@ func Registry() []Agent {
 			Detect: detectByMarker(cursorMarker),
 			MCP:    jsonMCP(cursorMCPPath, "mcpServers"),
 			Skill: &SkillTarget{
-				Strategy:    DedicatedFile,
-				Path:        cursorNoGlobalSkill, // no global rules mechanism
+				Strategy:    SkillFolder,
+				Path:        cursorUserSkillPath,
```
akijakya, line 48: **Migration: old single-file installs become orphans.**

Cursor and Copilot move from single files (`.cursor/rules/<slug>.mdc`, `prompts/<slug>.instructions.md`) to folders (`.cursor/skills/<slug>/SKILL.md`, `.copilot/skills/...`). Anyone who installed with the prior version can't `uninstall` the old artifacts through the new code — the paths no longer match, so the old files are left behind. Worth calling out in a release note.

cli/cmd/install/flags.go
```diff
@@ -27,6 +28,15 @@ func addSelectionFlags(cmd *cobra.Command, opts *options) {
 	flags.BoolVarP(&opts.yes, "yes", "y", false, "Skip the confirmation prompt")
 }
 
+func addBatchFlags(cmd *cobra.Command, opts *options) {
+	flags := cmd.Flags()
+
+	flags.Uint32Var(&opts.limit, "limit", 100, "Maximum number of records to install in batch mode") //nolint:mnd
```
akijakya: Nit: `--limit` caps the *search/pull* count, but `selectRecords` then applies `LatestByName` dedup afterward, so the number actually installed can be lower than `--limit`. The help text "Maximum number of records to install" slightly overstates it.

api/exportfmt/tar_iterator.go
```diff
@@ -0,0 +1,97 @@
...
+			if !yield(entry, nil) {
+				return
+			}
+		}
+	}, nil
+}
+
+func readTarEntry(tarReader *tar.Reader, header *tar.Header) (*TarEntry, error) {
+	reader := io.Reader(tarReader)
+	if header.Size >= 0 {
```
akijakya, line 84: Two minor points here:
- The typeflag filter (`o.typeflags` check, line ~71) is applied *after* `readTarEntry` has already read the full payload into memory. Filtering before reading would avoid buffering entries that get discarded (cheap for dirs, but tidier).
- `if header.Size >= 0` is effectively always true (tar sizes are non-negative), so the non-`LimitReader` branch is dead.

api/exportfmt/tar_iterator.go
```diff
@@ -0,0 +1,97 @@
...
+
+func WithTypeflag(typeflag byte) TarIteratorOption {
+	return func(opts *TarIteratorOptions) {
+		opts.typeflags = append(opts.typeflags, typeflag)
+	}
+}
+
+type TarIterator iter.Seq2[*TarEntry, error]
+
+type TarEntry struct {
```
akijakya, line 31: Nit: `TarEntry` is exported but both fields (`header`, `payload`) are unexported, so consumers outside `exportfmt` can hold a `*TarEntry` but can't read it. Fine within the module; if it's intended as package API, add accessors.

akijakya (commented):
Overall a solid PR — well-decomposed, the `TarIterator` refactor reads cleanly, and coverage is good (all changed packages' tests pass locally). Incorporating skill-bundle support into `install` is the right call. A few things worth addressing before merge — no blockers. Comments inline.

api/exportfmt/skill_bundle_extract.go
csirmazbendeguz: fixed

api/exportfmt/skill_bundle_extract.go
csirmazbendeguz: This is a good catch and I added the following adjustments:

- Check if additional files are present when checking for match
- When updating a skill, make sure to extract into a new directory and replace the old directory with a new directory - this way there won't be stale files

cli/internal/agentcfg/registry.go
csirmazbendeguz: Yeah. I think this is fine

cli/cmd/install/flags.go
csirmazbendeguz: It's fine

api/exportfmt/tar_iterator.go
csirmazbendeguz: it's fine

api/exportfmt/tar_iterator.go
csirmazbendeguz: it's fine

akijakya (approved):
LGTM! 🚀

Issue #1386, opened 2026-04-22
[Feature]: Add MCP export format for Claude Code


paralta:
### Problem Statement

The Directory currently supports exporting OASF records to GitHub Copilot MCP configuration format (`--format mcp-ghcopilot`). Claude Code uses a different configuration schema for its MCP servers, so users who work with Claude Code cannot directly consume exported MCP configurations without manual conversion.

Claude Code supports three ways to configure MCP servers:

1. **CLI commands** (`claude mcp add`) — the most common user-facing method.
2. **Project-scoped `.mcp.json`** — shared via version control.
3. **User config in `~/.claude.json`** — private per-user settings.

The `.mcp.json` format (project-scoped, shareable) uses this structure:

```json
{
  "mcpServers": {
    "server-name": {
      "command": "/path/to/server",
      "args": [],
      "env": {}
    }
  }
}
```

For remote HTTP servers:

```json
{
  "mcpServers": {
    "server-name": {
      "type": "http",
      "url": "https://api.example.com/mcp",
      "headers": {
        "Authorization": "Bearer ${API_KEY}"
      }
    }
  }
}
```

**Reference:** https://code.claude.com/docs/en/mcp

### Step 1: Add `RecordToClaudeCode` translator in oasf-sdk

Create a new translator function in [`github.com/agntcy/oasf-sdk/pkg/translator`](https://github.com/agntcy/oasf-sdk) following the existing `RecordToGHCopilot` pattern ([`pkg/translator/mcp.go` line 279](https://github.com/agntcy/oasf-sdk/blob/main/pkg/translator/mcp.go#L279)).

The new function should:

1. Define a `ClaudeCodeMCPConfig` struct that models the Claude Code `.mcp.json`
   schema:
   ```go
   type ClaudeCodeMCPConfig struct {
       MCPServers map[string]ClaudeCodeMCPServer `json:"mcpServers"`
   }

   type ClaudeCodeMCPServer struct {}
   ```
2. Implement `RecordToClaudeCode(record *structpb.Struct) (*ClaudeCodeMCPConfig, error)`.
3. Add unit tests covering all supported OASF versions and connection types.

### Step 2: Add `mcp-claudecode` format to the export command in this repository

1. Create `cli/cmd/export/format/mcp_claudecode.go` following the existing
   `mcp.go` (GHCopilot formatter) pattern:
   ```go
   func init() {
       RegisterFormatter("mcp-claudecode", &mcpClaudeCodeFormatter{})
   }
[...]

PR #1766, opened 2026-07-08
feat(dir): add mcp-claudecode export format

conorbronsdon:
Adds `--format mcp-claudecode` to `dirctl export`, closing #1386.

Pulls an OASF record's `integration/mcp` module (1.0.0 `connections` format) and emits Claude Code's project-scoped `.mcp.json` shape. Batch export (`--output-dir`) merges all matched servers into one `mcp.json`, mirroring the existing `mcp-ghcopilot` behavior.

Judgment calls where the issue was silent, called out for review:
- stdio entries omit `type`, matching what `claude mcp add` actually emits (the issue's stdio example also omits it).
- OASF's `streamable-http` transport maps to Claude Code's `"type": "http"`; `sse` passes through.
- Env vars without a `default_value` emit `"NAME": "${NAME}"`, Claude Code's documented shell-environment expansion. There's no Claude Code equivalent of VS Code's `${input:NAME}` prompt that the ghcopilot exporter uses.

Scope notes:
- Issue Step 1 targets oasf-sdk's translator package. This PR keeps the translator dir-local (`api/exportfmt/mcp_claudecode.go`, exported as `exportfmt.RecordToClaudeCode`) so the feature doesn't block on an oasf-sdk release and dependency bump. Happy to upstream it as a follow-up if you'd rather have the two-repo split.
- Step 2.4 references `mcp/tools/export_record.go`, which doesn't exist on current main; skipped.
- OASF 1.0.0 `connections` only; legacy 0.7.0/0.8.0 `servers` format isn't handled.

Verified: `go build ./...` and the exportfmt + cmd/export test suites pass on both the `api` and `cli` modules. Tests mirror the ghcopilot formatter's coverage (stdio, remote with headers, nil data, missing module, batch merge, empty batch, registration). Commits are DCO signed.

### DIFF
--- a/api/exportfmt/exportfmt.go
+++ b/api/exportfmt/exportfmt.go
@@ -40,6 +40,7 @@ const (
 	FormatAgentSkillBundle = "agent-skill-bundle"
 	FormatSkill            = "skill"
 	FormatMCPGHCopiot      = "mcp-ghcopilot"
+	FormatMCPClaudeCode    = "mcp-claudecode"
 )
 
 var (
--- a/api/exportfmt/exportfmt_test.go
+++ b/api/exportfmt/exportfmt_test.go
@@ -209,6 +209,12 @@ func TestGetFormatter(t *testing.T) {
 		assert.NotNil(t, f)
 	})
 
+	t.Run("returns mcp-claudecode formatter", func(t *testing.T) {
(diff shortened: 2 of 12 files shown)


ramizpolic (approved):
LGTM

Issue #1385, opened 2026-04-22
[Feature]: Add MCP export format for Cursor IDE


paralta:
### Problem Statement

The Directory currently supports exporting OASF records to GitHub Copilot MCP configuration format (`--format mcp-ghcopilot`). Cursor IDE uses a different configuration schema for its MCP servers, so users who work with Cursor cannot directly consume exported MCP configurations without manual conversion.

Cursor stores MCP server configurations in `.cursor/mcp.json` (project-scoped) or `~/.cursor/mcp.json` (global) using the following structure:

```json
{
  "mcpServers": {
    "server-name": {
      "command": "npx",
      "args": ["-y", "mcp-server"],
      "env": {
        "API_KEY": "value"
      }
    }
  }
}
```

For remote (HTTP/SSE) servers, the format uses a `url` field instead:

```json
{
  "mcpServers": {
    "server-name": {
      "url": "http://localhost:3000/mcp",
      "headers": {
        "API_KEY": "value"
      }
    }
  }
}
```

**Reference:** https://cursor.com/docs/mcp#installing-mcp-servers

### Step 1: Add `RecordToCursor` translator in oasf-sdk

Create a new translator function in [`github.com/agntcy/oasf-sdk/pkg/translator`](https://github.com/agntcy/oasf-sdk) following the existing `RecordToGHCopilot` pattern ([`pkg/translator/mcp.go` line 279](https://github.com/agntcy/oasf-sdk/blob/main/pkg/translator/mcp.go#L279)).

The new function should:

1. Define a `CursorMCPConfig` struct that models the Cursor `mcp.json` schema:
   ```go
   type CursorMCPConfig struct {
       MCPServers map[string]CursorMCPServer `json:"mcpServers"`
   }

   type CursorMCPServer struct {}
   ```
2. Implement `RecordToCursor(record *structpb.Struct) (*CursorMCPConfig, error)`
3. Add unit tests covering all supported OASF versions and connection types.

### Step 2: Add `mcp-cursor` format to the export command in this repository

1. Create `cli/cmd/export/format/mcp_cursor.go` following the existing`mcp.go` (GHCopilot formatter) pattern:
   ```go
   func init() {
       RegisterFormatter("mcp-cursor", &mcpCursorFormatter{})
   }
   ```
2. The formatter should call `translator.RecordToCursor(data)` and marshal the result to indented JSON.
3. Add tests in `cli/cmd/export/format/mcp_cursor_test.go`.
[...]

PR #1767, opened 2026-07-08
feat(dir): add mcp-cursor export format

conorbronsdon:
Adds `--format mcp-cursor` to `dirctl export`, closing #1385. Sibling of #1766 — the two branch from the same base, and whichever merges second takes a trivial keep-both-lines conflict in the shared registration lines.

Emits Cursor's `mcp.json` shape, verified against cursor.com/docs/context/mcp:
- No `type` field anywhere — Cursor infers transport from the config shape (`command` = stdio, `url` = remote). This is the main divergence from the Claude Code format in #1766.
- Env vars without a `default_value` use Cursor's documented `${env:NAME}` interpolation (note: different syntax from Claude Code's `${NAME}`).
- Cursor's `envFile` and OAuth `auth` fields aren't emitted; OASF records carry no data for them.

Batch export (`--output-dir`) merges all matched servers into one `mcp.json`, mirroring `mcp-ghcopilot`. Same scope notes as #1766: translator kept dir-local (`exportfmt.RecordToCursor`) rather than in oasf-sdk, happy to upstream as a follow-up; issue Step 2.4's `mcp/tools/export_record.go` doesn't exist on current main; OASF 1.0.0 `connections` only.

Verified: `go build ./...` plus the exportfmt and cmd/export test suites green on both modules. Commits are DCO signed.


ramizpolic (approved):
LGTM, thanks @conorbronsdon!

ramizpolic (approved):

PR #1775, opened 2026-07-08
feat(dir): expand MCP scanner coverage with yara/readiness and remote runner

### LINKED ISSUE #1740: [Feature]:  Expand MCP scanner coverage: YARA/readiness analyzers and live server scanning
### Problem Statement

`MCPRunner` in `utils/scanner/mcp.go` runs `mcp-scanner behavioral --raw <dir>` with no `--analyzers` flag, relying on the binary's defaults. This leaves two coverage gaps:

1. **Missing analyzers on source code scans.** `mcp-scanner` supports two zero-dependency analyzers — `yara` (YARA pattern matching) and `readiness` (static checks for missing timeouts, retries, error handling) — that never run because `--analyzers` is never passed. The LLM and Cisco AI Defense API analyzers are similarly opt-in but are not wired up.

2. **Live server endpoints are never scanned.** Records can contain `remotes[].url` entries pointing to running MCP endpoints. `mcp-scanner` can scan tools, prompts, resources, and system instructions against a live server (`remote`, `prompts`, `resources`, `instructions` subcommands), but no runner for this exists today. Records with no `source_code` locator but a valid remote URL are silently skipped entirely.

### Proposed Solution

**1. Explicit `--analyzers` in `MCPRunner`** — pass `--analyzers yara,readiness` always so both zero-dependency analyzers run regardless of environment:

```go
// utils/scanner/mcp.go
cmd := exec.CommandContext(ctx, cliPath, "behavioral", "--analyzers", "yara,readiness", "--raw", scanDir)
```

> **Note:** `mcp-scanner` also supports `llm` (LLM-as-judge) and `api` (Cisco AI Defense) analyzers that could significantly deepen findings. These require third-party credentials and are out of scope here, but are worth revisiting once the zero-dependency baseline is in place.

**2. New `RemoteRunner`** — add `utils/scanner/remote.go` implementing the `Runner` interface. It extracts `remotes[].url` from the record, then runs `remote`, `prompts`, `resources`, and `instructions` subcommands against each endpoint, merging findings into a single `ScanResult`. Network errors (unreachable, auth failure) are treated as skipped with a warning, not hard errors. Register it alongside `MCPRunner` wherever `RunAll` is called.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

conorbronsdon:
Closes #1740.

1. MCPRunner now passes `--analyzers yara,readiness` explicitly to the behavioral scan, so both zero-dependency analyzers always run instead of relying on the binary's undocumented defaults. Global flags are placed before the subcommand, matching mcp-scanner's CLI contract (its docs reject flags placed after the subcommand). Analyzers metadata on the result now names the analyzers that actually ran ("yara", "readiness") rather than the subcommand name.

2. New RemoteRunner (utils/scanner/remote.go) scans live MCP server endpoints. It extracts connection URLs from `modules[*].data.connections[]` (transport type "sse" or "streamable-http"; "stdio" connections are local and skipped), runs mcp-scanner's remote/prompts/resources/instructions subcommands against each endpoint via `--server-url`, and merges results with the existing package-level merge(). Network errors or unparsable output for a given endpoint/subcommand pair are recorded as a skipped sub-scan with a warning, never a hard error, so one bad endpoint doesn't block the rest. Registered alongside MCPRunner in the scan reconciler task.

Two deviations from the issue text, called out for review:

- The issue describes extracting `remotes[].url` from the record. That field doesn't exist in the OASF schema (checked against schema.oasf.outshift.com/1.0.0 for mcp_data and mcp_server_connection): the real field is the module's `connections[]`, where each connection carries a type and, for the two remote transports, a url. RemoteRunner reads that field instead, the same path the mcp-claudecode/mcp-cursor exporters (#1766/#1767) use. Happy to adjust if the issue meant something else by "remotes".
- On the same note: a module's `data` is itself the mcp_data object in decoded records (verified against buildMCPModule in server/skill and the record_100.json e2e fixture, whose MCP module data has top-level keys `connections`/`name` and no `mcp_data` wrapper). An earlier revision of this branch read `data.mcp_data.connections` and would never have matched a real record; the last commit corrects both the new connection extraction and the pre-existing `extractSubfolder`, which had the same nonexistent `data.mcp_data.*` path and always returned "".

Follow-ups noted rather than built, since the issue doesn't settle them:

- llm/api analyzers stay opt-in (they need third-party credentials, out of scope per the issue's own note).
- A SCANNER_TYPE_REMOTE proto enum needs a scanv1 proto change + buf generate cycle, left for a separate change. Until then the scan_reports row says "remote" but the ScanReport referrer carries SCANNER_TYPE_UNSPECIFIED, so the AI Finder UI (#1760) can't label remote scans yet.
- Endpoints are not de-duplicated across connections/modules; a URL listed twice is scanned twice.

Verified: go build/vet/test clean on the utils, reconciler, and server modules. Runner exec paths are covered by unit tests via a fake CLI binary (testdata/fakecli), including RemoteRunner end-to-end and MCPRunner.Run. One caveat for reviewers: the fake CLI matches on argument substrings, so the exact flag ordering contract (`--raw <subcommand> --server-url <url>`) is grounded in mcp-scanner's documentation rather than exercised against the real binary — a live smoke run would be a good double-check before merge.


paralta:
Good work here, @conorbronsdon 🚀  Overall looks good to me, my only note would be that we could add some URL validation - the URLs come directly from externally-published records, so a scheme allowlist (e.g. `https://` only) and rejecting private/loopback ranges before passing to `mcp-scanner` would be a good safeguard.

One thing worth discussing: `connections[]` is MCP-specific in the OASF schema, so `RemoteRunner` will only ever find endpoints in `integration/mcp` modules. So, this `mcp` and `remote` separation might not be adequate here, and it creates the `SCANNER_TYPE_REMOTE` gap you noted as a follow-up. I realise this retracts my original suggestion from #1740.

WDYT about including this remote scanning into `MCPRunner` as a second step — run source code scan, run live endpoint scan,  and merge findings?

conorbronsdon:
Good thought - let me take a look

conorbronsdon:
Thanks @paralta, both land, and I'd rather take the retraction than ship the split. Pushed as `e176d4ba` and `0fdd2a07`, rebased onto main.

**Merging remote into `MCPRunner`:** agreed. I checked the schema first: `connections` is an attribute of `mcp_data` in `schema/objects/mcp_data.json` and nothing else under `schema/objects` declares it, so `RemoteRunner` walked every module but could only match `integration/mcp`.

Folding it in fixed more than ergonomics. `toProtoScannerType` switches on `mcp`/`skill`/`a2a` with no `remote` case, so the standalone runner was already writing every remote report as `SCANNER_TYPE_UNSPECIFIED`. Two properties kept deliberately: the source scan still fails hard while the endpoint phase is skip-with-warning, and the endpoint phase has its own toggle since it dials third-party servers.

**URL validation:** agreed, and worth being blunt about how it first landed. My initial pass used `net.ParseIP` + `IsPrivate`, and a review of my own branch showed it was bypassable by `https://2852039166/` — 169.254.169.254 in decimal — plus the octal, hex, short, and trailing-dot spellings. `net.ParseIP` accepts only canonical forms, so everything else fell through to a "then it must be a hostname" branch while the resolver read them as addresses. The commit message asserted the metadata address was blocked when it wasn't, which is worse than no validation.

The rewrite normalizes the trailing dot, then requires either a canonical `netip` address outside an explicit reserved-prefix table or a syntactically valid DNS name (an all-digit or `0x` label is not one). The table also covers what `IsPrivate` misses, notably 100.64.0.0/10 — the pod and node CIDR on several managed Kubernetes offerings, i.e. exactly the in-cluster case. All 25 bypasses have regression tests.

`allow_private_endpoints` re-permits reserved ranges for self-hosted deployments; I split `allow_insecure_transport` out rather than have one flag do both, so needing a private endpoint doesn't also opt you into cleartext scans of public hosts.

Still a mitigation, and the code says so: dir validates the URL but `mcp-scanner` opens the connection, so a name resolving to a reserved address at dial time still gets through.

**Two things I deliberately did not fix**, because they're consequences of the merge and your call:

1. **A record with no source code now persists as a passing MCP scan.** `merge` sets `SkippedReason` only when every phase skips, so a skipped source phase plus a clean endpoint phase yields `Safe=true, Skipped=false` and writes `ScanReport{ScannerType: "MCP", IsSafe: true}`. Under the split runners no `MCP` row was written at all.
2. **A publisher who crashes the scanner on their own repo suppresses the endpoint scan.** `runSourceScan` hard-errors on any non-zero `mcp-scanner` exit, and `Run` returns before the endpoint phase, so no row is written. `RemoteRunner` used to report those findings independently.

Both are report-model questions rather than validation ones. Happy to take either here.

paralta (approved):
Thanks for addressing the feedback so promptly, @conorbronsdon !

Both points are acceptable to me. On the first, the `analyzers` field makes it clear which sub-scans actually ran, so a hosted agent with no source code getting a clean MCP result from endpoint scanning alone is fine. On the second, the retry covers the common transient failure case though, if source scanning consistently fails on a record, endpoint scanning will be permanently blocked too. That's a known limitation worth addressing later by making the two phases independent within MCPRunner, but not a blocker for this PR 👍

PR #1776, opened 2026-07-08
feat(dir): add a2a-scanner support

### LINKED ISSUE #1725: [Feature]: Add a2a-scanner support
### Problem Statement

Records that implement the Agent-to-Agent (A2A) protocol expose a different attack surface — inter-agent communication channels, delegation chains, and capability negotiation — that neither `mcp-scanner` nor `skill-scanner` covers. Without A2A-specific scanning, these records enter the directory with no assessment of A2A-specific risks.

### Proposed Solution

https://cisco-ai-defense.github.io/docs/a2a-scanner

Add an `A2ARunner` implementation to `utils/scanner` alongside `MCPRunner` and `SkillRunner`. The runner is invoked when the record signals A2A support (e.g., an A2A extension or locator type in the OASF record). Results are stored as a `ScanReport` referrer with `scanner_type = A2A` and indexed in `scan_reports` like any other scanner type.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

conorbronsdon:
Adds an `A2ARunner` to `utils/scanner` alongside `MCPRunner` and `SkillRunner`. Fixes #1725.

Records that implement the Agent-to-Agent protocol expose an attack surface (inter-agent communication, delegation chains, capability negotiation) that neither mcp-scanner nor skill-scanner assesses. This runner closes that gap using the [a2a-scanner CLI](https://cisco-ai-defense.github.io/docs/a2a-scanner).

## How it works

- The runner fires when a record carries a module whose data holds a `card_data` object (the OASF a2a_data shape, with `card_data` directly under the module's `data`) — matching on data shape rather than module name, the same way the MCP runner keys on `mcp_data`. The AgentCard is written to a temp file and scanned with `a2a-scanner scan-card`.
- Analyzers: `heuristic`, `spec`, and `yara` run by default because they need no third-party credentials or live endpoint. `llm` and `endpoint` are opt-in follow-ups, mirroring how the MCP runner treats its llm/api analyzers.
- Results go through the existing plumbing: `ScanReport` referrer with `scanner_type = A2A` plus a `scan_reports` row. The `SCANNER_TYPE_A2A` enum and DB support were already in place; this adds the missing runner.
- Config follows the existing pattern: `A2ACLIPath` on the scan task, `scan.a2a_cli_path` env binding, sample daemon config, and a `deps:a2a-scanner` Taskfile target (uv package `cisco-ai-a2a-scanner`).
- Since this added a third copy of the severity switch and the Azure env test, the shared parts are now consolidated: one `mapScannerSeverity` used by all three runners, and a `testAzureEnvDerivation` helper shared by the skill/a2a env tests (goconst and dupl were flagging the copies).

## A note on output parsing

a2a-scanner's JSON output schema is not documented upstream. `parseA2AOutput` assumes the `is_safe`/`findings` shape that skill-scanner emits, and the struct doc comment flags this as the single place to adjust if the installed binary differs. Happy to update if a maintainer can share a sample raw output.

## Testing

Unit tests cover card extraction (validated against the repo's OASF record fixtures, including records with no A2A module), output parsing, severity mapping, env derivation, and skip behavior — mirroring the mcp/skill test suites. `go build ./... && go vet ./... && go test ./...` pass in the `utils` and `reconciler` modules.

### DIFF
--- a/Taskfile.deps.yml
+++ b/Taskfile.deps.yml
@@ -420,8 +420,18 @@ tasks:
       - cmd: >-
           {{.UV_BIN}} tool install --python 3.13 cisco-ai-skill-scanner
 
+  deps:a2a-scanner:
+    desc: Install a2a-scanner CLI via uv
+    deps:
+      - task: deps:uv
+    cmds:
+      - cmd: echo "Installing a2a-scanner"
+      - cmd: >-
+          {{.UV_BIN}} tool install --python 3.13 cisco-ai-a2a-scanner
+
   deps:scanners:
-    desc: Install all security scanner CLIs (mcp-scanner and skill-scanner) via uv
+    desc: Install all security scanner CLIs (mcp-scanner, skill-scanner, and a2a-scanner) via uv
     cmds:
       - task: deps:mcp-scanner
       - task: deps:skill-scanner
+      - task: deps:a2a-scanner
--- a/cli/cmd/daemon/daemon.config.yaml
+++ b/cli/cmd/daemon/daemon.config.yaml
@@ -96,6 +96,7 @@ reconciler:
     record_timeout: 2m
     mcp_cli_path: "mcp-scanner"
     skill_cli_path: "skill-scanner"
+    a2a_cli_path: "a2a-scanner"
   local_registry:
     registry_address: "localhost:5555"
     repository_name: "dir"
--- a/reconciler/config/config.go
+++ b/reconciler/config/config.go
@@ -225,6 +225,9 @@ func LoadConfig() (*Config, error) {
 	_ = v.BindEnv("scan.skill_cli_path")
 	v.SetDefault("scan.skill_cli_path", scan.DefaultSkillCLIPath)
 
+	_ = v.BindEnv("scan.a2a_cli_path")
+	v.SetDefault("scan.a2a_cli_path", scan.DefaultA2ACLIPath)
+
 	//
 	// Providers task configuration (provider-count gauge)
 	//
--- a/reconciler/tasks/scan/config.go
+++ b/reconciler/tasks/scan/config.go
@@ -23,6 +23,9 @@ const (
 
 	// DefaultSkillCLIPath is the default binary name for the skill scanner.
 	DefaultSkillCLIPath = "skill-scanner"
+
+	// DefaultA2ACLIPath is the default binary name for the A2A scanner.
+	DefaultA2ACLIPath = "a2a-scanner"
 )
 
 // Config holds configuration for the security scan reconciliation task.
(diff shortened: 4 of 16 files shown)


conorbronsdon:
Added tests for the a2a-scanner runner exec path (the same gap the mcp/skill runners had), which brings `a2a.go` from about 60% to 95%. Built a small fake scanner in testdata rather than re-execing the test binary so it stays clear of the gosec taint checks. The earlier CodeQL red was a 504 on the Install Task step, so this push should re-run it clean.

conorbronsdon:
The red check here (`Test / E2E / Test Integration Server`) isn't from this PR — it's the repo-wide OASF fixture drift currently failing on ~10 open PRs. The `Push` test rejects the shared test record because the OASF SDK bump on `main` renamed the skill taxonomy (e.g. `natural_language_processing/natural_language_understanding` → `language_processing/language_understanding`), but the fixtures still carry the old names.

That's fixed independently in #1785, which updates exactly those skill values. Once #1785 merges I'll rebase this branch and it should go green. This PR touches only `utils/scanner/*` and config — no test fixtures — and `Test / Unit` + `Lint` pass. The other E2E jobs (Local/Network/Client) were fail-fast cancellations off the same failure.

One note for whenever you review: validation now emits a deprecation warning for the inline `modules[].data.card_data` shape this runner keys on (`please use module.artifact instead`). It's only a warning, not a failure, but a follow-up could also match `module.artifact` — happy to add that here or in a separate PR if you'd prefer.

adamtagscherer:
> The red check here (`Test / E2E / Test Integration Server`) isn't from this PR — it's the repo-wide OASF fixture drift currently failing on ~10 open PRs. The `Push` test rejects the shared test record because the OASF SDK bump on `main` renamed the skill taxonomy (e.g. `natural_language_processing/natural_language_understanding` → `language_processing/language_understanding`), but the fixtures still carry the old names.
> 
> That's fixed independently in #1785, which updates exactly those skill values. Once #1785 merges I'll rebase this branch and it should go green. This PR touches only `utils/scanner/*` and config — no test fixtures — and `Test / Unit` + `Lint` pass. The other E2E jobs (Local/Network/Client) were fail-fast cancellations off the same failure.
> 
> One note for whenever you review: validation now emits a deprecation warning for the inline `modules[].data.card_data` shape this runner keys on (`please use module.artifact instead`). It's only a warning, not a failure, but a follow-up could also match `module.artifact` — happy to add that here or in a separate PR if you'd prefer.

Hey @conorbronsdon !

We merged https://github.com/agntcy/dir/pull/1785, if you rebase on main then the CI error should be resolved.

conorbronsdon:
Rebased (picking up #1785) the failing tests should be resolved now. Thanks @adamtagscherer!

ramizpolic (approved):

PR #1783, opened 2026-07-10
feat(routing): add natural-language search to dirctl routing search

### LINKED ISSUE #1694: [Feature]: Order routing and local search results by match-score relevance
### Problem Statement

Results are not ordered by how well they match the query. Routing search computes `match_score` but only uses it as an OR filter threshold, so the strongest matches aren't surfaced first. Local index search (`GetRecordCIDs`) has no `ORDER BY` at all — order is DB-dependent and unstable under pagination. Users cannot tell which results are genuinely the best match.

### Proposed Solution

Rank each search surface by its appropriate relevance signal:
- **Routing search:** order `RoutingService.Search` results by `match_score` (descending) before applying `limit`, with unchanged OR / `min_match_score` semantics.
- **Local index search:** add a coarse relevance score derived from which query clauses matched and match specificity (exact vs. wildcard). Introduce a configurable field-weighting model (e.g. name/skill > domain > annotation) with documented defaults.
- Expose a `sort_mode` option on the search request protos where missing, with "most relevant" as the default when a query is present, documented and backward compatible.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

paralta:
This PR adds free-text query support to `dirctl routing search`, mirroring the NL search already available in `dirctl search`. When a positional query is given, the OASF extractor decomposes it into skill and domain signals (keyword signals are dropped — the DHT has no name/description index), each signal becomes a routing DHT query, and results are returned ranked by `match_score` descending. Structured routing search also gains score-sorted output, fixing the prior behaviour where results arrived in arbitrary DHT discovery order.

- The extractor uses semantic-only weights and a lower min-score threshold (0.1 vs the 0.3 default for local search): the DHT indexes exact taxonomy names, so recall matters more than precision and lexical overlap only hurts conceptual queries.
- `DecomposeWithMinScore` is introduced as a parameterised variant of `Decompose` so routing can widen the threshold without affecting the local index search path.


tkircsi:
Just one small comment. The "and results are returned ranked by match_score descending" is true only for the "limited" recordset defined by the `--limit` flag not for the entire recordset. Order happens after the limit. Maybe the default limit could be higher or the order should be happened on the server side.

cli/cmd/routing/search.go
```diff
@@ -15,44 +16,57 @@ import (
...
+	Use:   "search [query]",
 	Short: "Search for remote records from other peers",
 	Long: `Search for remote records from other peers using the routing API.
 
-This command discovers records that have been published by other peers in the network.
-It uses cached network announcements and filters out local records.
+Provide a free-text query as a positional argument to use natural-language
+search: the OASF extractor (set up by 'dirctl init') decomposes the phrase
+into skill and domain signals. Each signal is queried independently against
+the routing DHT and results are ranked by match score, best-first.
```
tkircsi: it's "the discovered window, reordered," not true global top-N.

tkircsi (approved):
LGTM! added a small finding.

cli/cmd/routing/search.go
paralta: very good point! thanks for reviewing 👍

PR #1799, opened 2026-07-13
feat(dir): add listreferrers, pullreferrer rpc

### LINKED ISSUE #1793: [Feature]: Add libp2p referrer RPC methods ListReferrers + PullReferrer (DHT Autosync #3)
## Summary

Add peer-to-peer **referrer** methods to the routing libp2p RPC service
(`/dir/rpc/1.0.0`) so a node can fetch a record's referrers (signatures, public
keys, scan reports) directly from a peer over libp2p — the transport used by DHT
autosync. Today the RPC only exposes `Lookup` and `Pull` (record body); referrers
are only reachable via the gRPC `StoreService`, which is not NAT-friendly.

Part of Epic #1790.

**Depends on: none** (Wave 1, parallelizable). This is the read-only *serving*
path (backed by `ReferrerStoreAPI`); it does not depend on the Ingestor (#2).
It is consumed by the autosync worker (#5).

## Motivation

Autosync must pull a record **and its referrers** from an allow-listed peer over
the DHT/libp2p transport. `go-libp2p-gorpc` is unary-only (no server streaming),
and scan-report referrers can be large, so a single "return everything" call
risks blowing the `MaxPullSize` (4 MB) limit. Split into:
- a cheap **metadata listing** call, then
- a **per-referrer** fetch.

### New protocol methods on `/dir/rpc/1.0.0`

- `ListReferrers(ctx, *corev1.RecordRef) -> ListReferrersResponse{ Referrers []ReferrerDescriptor }`
  where `ReferrerDescriptor{ Cid string; Type string }` — **metadata only**
  (no blob data), produced via `ReferrerStoreAPI.WalkReferrers`.
- `PullReferrer(ctx, PullReferrerRequest{ RecordRef, ReferrerType, ReferrerCid }) -> PullReferrerResponse{ Referrer *corev1.RecordReferrer }`
  — returns a single referrer blob; enforce the `MaxPullSize` cap.

Add matching method-name constants alongside `DirServiceFuncLookup` / `Pull`.

### Service access to referrer storage

- `rpc.Service` currently receives `types.StoreAPI`. In `rpc.New`, type-assert to
  `types.ReferrerStoreAPI` (the OCI store is a `types.FullStore`). If the store
  does not support referrers, the referrer RPC methods return a clear
  `codes.Unimplemented`/`FailedPrecondition` error (records-only `Pull` keeps
  working).

### Client helpers

- `Service.ListReferrers(ctx, peer.ID, *corev1.RecordRef) ([]ReferrerDescriptor, error)`
- `Service.PullReferrer(ctx, peer.ID, recordRef, referrerType, referrerCid) (*corev1.RecordReferrer, error)`

mirroring the existing `Service.Pull` (`server/routing/rpc/rpc.go:160`).

## Out of scope

- Ingesting the pulled referrers (that's the worker in #5, using the Ingestor #2).
- Wiring/DI changes (#4).

## Security / robustness

- Treat requests as untrusted: validate `RecordRef`/inputs, enforce `MaxPullSize`
  on the returned blob, and cap the number of descriptors returned by
  `ListReferrers`.
- Structured logging of served referrer requests (peer, record CID, type).

## Acceptance criteria

- [ ] `ListReferrers` returns referrer `{cid, type}` metadata for a record via
      `WalkReferrers` (no blob data).
- [ ] `PullReferrer` returns a single referrer blob and enforces `MaxPullSize`.
- [ ] Stores without referrer support return a clear error for the referrer
      methods; `Pull`/`Lookup` are unaffected.
- [ ] Client helpers (`ListReferrers` / `PullReferrer`) round-trip against a live
      in-process host (unit test with two libp2p hosts).
- [ ] Unit tests cover: multiple referrer types, empty referrer set, oversize
      blob rejection, and unsupported-store error path.

## Notes / references

- RPC service + existing `Lookup`/`Pull`: `server/routing/rpc/rpc.go`
  (protocol `Protocol = "/dir/rpc/1.0.0"`, `MaxPullSize = 4 MB`).
- Referrer serving API: `types.ReferrerStoreAPI.WalkReferrers`
  (`server/types/store.go`).
- gRPC counterpart for parity of behavior: `storeCtrl.PullReferrer`
  (`server/controller/store.go:341`).
- Design reference: `dir-tmp/dht-sync/IMPLEMENTATION_PLAN.md` §5.2.

csirmazbendeguz:


server/routing/rpc/rpc.go
```diff
@@ -45,6 +54,26 @@ type LookupResponse struct {
...
 
+// ReferrerDescriptor carries referrer metadata without blob data.
+type ReferrerDescriptor struct {
+	Cid  string
+	Type string
+}
+
+type ListReferrersResponse struct {
+	Referrers []ReferrerDescriptor
+}
```
tkircsi, line 66: Maybe it worths to ad a `Truncated` bool flag to the response, when the list is truncated it would be clear for the caller.

server/routing/rpc/rpc.go
```diff
@@ -175,5 +358,72 @@ func (s *Service) Pull(ctx context.Context, peer peer.ID, req *corev1.RecordRef)
...
+
+	return resp.Referrers, nil
+}
+
+func (s *Service) PullReferrer(
+	ctx context.Context,
+	peerID peer.ID,
+	recordRef *corev1.RecordRef,
+	referrerType string,
+	referrerCid string,
```
tkircsi: Can we use `ReferrerDescriptor` instead of 2 strings field?

tkircsi (commented):
2 small nits

tkircsi (approved):
LGTM

PR #1808, opened 2026-07-14
feat(routing): nat hardening for autosync

### LINKED ISSUE #1796: [Feature]: NAT / reachability hardening for autosync pulls (FindPeer, backoff, relay) (DHT Autosync #6)
## Summary

Make autosync work across NATs/firewalls. Real-world testing showed that when
two daemons are on a shared routable network (VPN) autosync works end-to-end,
but when a peer is behind NAT (off-VPN) the GossipSub label announcement still
propagates while **autosync never ingests** — because the DHT provider
notification and/or the `rpc.Pull` cannot cross NAT.

The decisive fix is **circuit-relay v2 + DCUtR** (working relayed connectivity so
NAT'd peers are reachable and hole punching can be coordinated), backed by a
relay service deployed on reliable public nodes. `FindPeer` re-resolution +
timeouts/backoff are secondary hardening (they help with stale addresses and
transient failures but cannot make a fundamentally unreachable peer reachable).

Part of Epic #1790. Builds on #4/#5 (FindPeer already available via `Server.DHT()`).

## Root cause (from testing)

- The libp2p host enables DCUtR (`EnableHolePunching`), UPnP (`NATPortMap`),
  AutoNAT (`EnableNATService`) and an AutoRelay **client**
  (`internal/p2p/server.go: setupAutoRelay`).
- **No node runs a circuit-relay v2 *service*** (`EnableRelayService` is not set
  anywhere), so AutoRelay clients find no relays → NAT'd peers get no
  `/p2p-circuit` address, and DCUtR has no relayed connection to coordinate
  through.
- Autosync is triggered **only** by the DHT provider-notification path
  (`handleCIDProviderNotification`); the GossipSub path does not trigger it.
  Off-VPN the GossipSub label reaches the subscriber (mesh flood) but the DHT
  provider notification often does not → autosync never fires (no autosync logs),
  and even if it did, the `Pull` back to the NAT'd peer would fail.

### 1. Circuit-relay v2 service on reliable public nodes (PRIMARY)

- Enable `libp2p.EnableRelayService()` on publicly-reachable nodes so they act as
  relays. Deploy this on the **dev and prod bootstrap nodes** (they are already
  public, stable, and known to all peers) — see `agntcy-deployment`.
- Consider relay resource limits/ACLs to protect the relay node, and whether the
  default *limited* relay caps are sufficient for record Pulls (see note below).

### 2. Client relay configuration + DCUtR

- Keep AutoRelay client enabled; additionally support **static relay addresses**
  in the daemon config (more reliable than pure DHT discovery). Point clients at
  the dev/prod relays.
- Verify DCUtR upgrades relayed connections to direct ones when possible (the
  reliable data path is "relay to coordinate → DCUtR direct"). Note: relay v2 is
  a *limited* relay by default; a multi-MB record Pull that stays on a limited
  relay (e.g. symmetric NAT where DCUtR fails) may exceed relay limits — document
  and, if needed, raise limits or use a dedicated relay.

### 3. FindPeer re-resolution + timeouts/backoff (SECONDARY)

- On pull dial failure, `FindPeer(ctx, peerID)` to refresh addresses, then retry
  with a small bounded exponential backoff + jitter; each attempt time-bounded.
  Give up gracefully after N attempts (a later re-announcement re-triggers).
  (A single FindPeer retry already exists from #5; this generalizes it.)

### 4. GossipSub-triggered autosync (ENHANCEMENT)

- Also trigger autosync from the GossipSub path (`handleRecordPublishEvent`),
  since GossipSub reaches NAT'd subscribers far more reliably than the DHT
  provider notification. GossipSub events carry only the authenticated peer ID
  (no addresses), so resolve addresses via `FindPeer` before pulling. This makes
  autosync fire even when the DHT provider notification does not arrive (the Pull
  still needs relay/DCUtR to succeed).

## Out of scope

- Core worker/trigger (#5) — this hardens its reachability.
- E2E scenarios (#7) — incl. a NAT'd topology test.

## Acceptance criteria

- [ ] A circuit-relay v2 service runs on the dev and prod bootstrap nodes;
      daemons obtain `/p2p-circuit` addresses via AutoRelay and/or static config.
- [ ] Two daemons on different networks (at least one behind NAT, no shared VPN)
      successfully autosync a record end-to-end (ingest, not just announcement).
- [ ] DCUtR upgrades to a direct connection where the NAT allows; relay limits
      for record-sized Pulls are documented (and raised/dedicated relay used if
      required for the pure-relay fallback).
- [ ] On pull dial failure, the worker re-resolves via `FindPeer` and retries
      with bounded backoff; final failure is logged and does not block others.
- [ ] Autosync is also triggered from the GossipSub path (addresses resolved via
      `FindPeer`), so a NAT'd subscriber attempts sync even without the DHT
      provider notification.
- [ ] No regression to the VPN / non-NAT happy path.

## Notes / references

- Existing NAT stack: `server/routing/internal/p2p/host.go`
  (`EnableHolePunching`, `NATPortMap`, `EnableNATService`); AutoRelay client in
  `internal/p2p/server.go` (`setupAutoRelay`).
- Autosync trigger: `handleCIDProviderNotification` (`routing_remote.go`);
  GossipSub path: `handleRecordPublishEvent`.
- Autosync manager: `server/routing/autosync`.
- Deployment: `agntcy-deployment` dev/prod bootstrap values.
- Design reference: `dir-tmp/dht-sync/IMPLEMENTATION_PLAN.md` §5.6.

tkircsi:
Closes #1796


ramizpolic (approved):

server/config/config.go
```diff
@@ -523,6 +523,17 @@ func LoadConfig(opts ...ConfigOption) (*Config, error) {
 	_ = v.BindEnv("routing.autosync.enabled")
 	v.SetDefault("routing.autosync.enabled", routing.DefaultAutosyncEnabled)
 
+	//
+	// Routing relay configuration (circuit-relay v2 for NAT traversal).
+	// Note: routing.static_relays is a list and is configured via config
+	// file/YAML only (it cannot be bound to a single environment variable).
+	//
+	_ = v.BindEnv("routing.relay_service")
+	v.SetDefault("routing.relay_service", routing.DefaultRelayServiceEnabled)
```
ramizpolic, line 532: we do not seem to be loading the `static_relays` addresses here, is this intentional?

tkircsi: We load it, but just from file by viper. Multiaddrs doesn't map cleanly to one env var so that is the reason we don't bind it.

csirmazbendeguz (approved):

server/routing/autosync/autosync.go
```diff
@@ -272,31 +290,88 @@ func (m *Manager) process(parentCtx context.Context, j job) {
...
-// DHT (FindPeer) and retries once. Fuller backoff/relay handling is a follow-up.
+// pullRecord pulls the record over libp2p RPC with bounded retries. Addresses
+// from the announcement are added to the peerstore first. Between attempts it
+// backs off (exponential + jitter) and re-resolves the peer's current addresses
+// via the DHT (FindPeer) — recovering from stale addresses and giving NAT'd
+// peers time for a hole-punch/relay handshake to settle.
 func (m *Manager) pullRecord(ctx context.Context, peerID peer.ID, ref *corev1.RecordRef, addrs []ma.Multiaddr) (*corev1.Record, error) {
 	if len(addrs) > 0 {
 		m.router.AddAddrs(peerID, addrs)
+	} else if addrInfo, err := m.router.FindPeer(ctx, peerID); err == nil && len(addrInfo.Addrs) > 0 {
```
csirmazbendeguz, line 301: NIT: refactoring idea

I feel like this block and the one in for loop are the same and the idea is this right?

```
for ... {
  sleepWithBackoff()
  FindPeer()
  Pull()
}
```

And the only reason for this block and the `if attempt > 1` block is that you don't want to sleep on the first iteration

But if you started sleepWithBackoff with 0, then this could be simplified

tkircsi: It's a readability/semantics choice, not a functional requirement — attempt is used as a human-facing 1-based counter, and the loop body/logging/backoff math all assume that.

PR #1812, opened 2026-07-15
feat(catalog): replace placeholder stats with real usage metrics

paralta:
The AI Catalog UI was showing deterministic fake values for downloads, providers, and a star rating. Usage counters (`pull_count`, `lookup_count`, `provider_count`) were already being tracked server-side but never surfaced to the catalog API. This PR injects them into each `CatalogEntry`'s metadata map — the same pattern used for scan results — and updates the UI to read and display them. The star rating is removed as no real rating signal exists.

Also fixes the metrics reconciler task being silently skipped in Docker and Helm deployments due to a missing `server_address` configuration.

image


install/charts/dir/apiserver/values.yaml
```diff
@@ -728,6 +728,17 @@ reconciler:
     # OASF schema URL for validation
     schema_url: "https://schema.oasf.outshift.com"
 
+    # Apiserver gRPC address — required by the metrics task to query provider counts
+    # from the routing layer. Typically the in-cluster apiserver service name + gRPC port.
+    server_address: ""
```
ramizpolic, line 733: how do you fetch the metrics data for a specific record (e.g. from gRPC api, if so which one)? 
my concern is that, is the reconciler using authenticated client to talk to the server, otherwise this will keep failing

ramizpolic (approved):

install/charts/dir/apiserver/values.yaml
paralta: this is not fetching metrics, it is fetching the provider count from the routing client. For authentication, we had this discussion previously here https://github.com/agntcy/dir/pull/1708

PR #1819, opened 2026-07-16
feat(install): wire up metrics reconciler task in Helm chart

paralta:
The metrics reconciler task was enabled in the Docker config but never added to the Helm chart, causing it to be silently skipped in Kubernetes deployments. This PR enables it in `values.yaml` and adds a `chart.apiserver.grpcAddress helper that auto-resolves the in-cluster gRPC address needed by the task (following the same override pattern as the postgres host and OCI registry helpers), injected into the rendered reconciler config at render time. The Docker interval changes are unrelated improvements.


tkircsi (approved):
LGTM

PR #1849, opened 2026-07-20
test(e2e): add extractor enricher coverage for dirctl import

### LINKED ISSUE #1848: [Feature]: Add e2e test for the extractor enricher in the import pipeline
### Problem Statement

The `dir-importer` enricher supports three backends — `static`, `llm`, and `extractor` — but only the `static` backend is exercised in e2e tests. The extractor enricher path (`enricher.extractor` in the import YAML config) is entirely untested end-to-end.

Specifically:
- `tests/e2e/local/10_export_test.go` forces the static enricher via `factory.Replace` in every import-based test, explicitly avoiding any real enricher.
- `tests/e2e/local/15_nl_search_test.go` does check for the extractor being provisioned (via `dirctl init`'s `~/.agntcy/oasf-sdk/manifest.json`), but only to exercise the **search** path, not the **import/enrichment** pipeline itself.

As a result, a regression in the extractor enricher integration (e.g., the `enricherconfig.ExtractorConfig` wiring in `cli/cmd/import/config.go:132`, the `oasfExtractorAdapter`, or the `internalextractor.LoadConfigured()` call) would go undetected by e2e.

### Proposed Solution

Add a new test context to `tests/e2e/local/` (e.g., a new file `16_extractor_enricher_test.go`) that:

1. **Guards on provisioning** — checks for the `manifest.json` sentinel written by `dirctl init` (same pattern as `15_nl_search_test.go:34`), and calls `ginkgo.Skip(...)` if the extractor is not provisioned. This keeps the test optional in CI environments without the assets.
2. **Writes an import YAML config** with `enricher.extractor: {}` (empty block → uses the `dirctl init` config, delegating to `internalextractor.LoadConfigured()`).
3. **Runs `dirctl import`** (e.g., importing an A2A agent card or MCP server from the shared test data) with `--config` pointing at that YAML — without any `factory.Replace` override.
4. **Asserts enrichment ran** — pulls the resulting record and verifies it has at least one non-empty skill and one non-empty domain, confirming the extractor produced output rather than returning empty results.

This mirrors how users actually invoke the extractor enricher through the CLI, and covers the code path from `loadConfig` → `loadExtractorFromConfig` → `internalextractor.LoadConfigured` → `sdk.Extractor.Extract` → record enrichment.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

Shaurya2k06:
Import-based e2e tests currently force the static enricher via `factory.Replace`, so the `enricher.extractor` CLI path (`loadExtractorFromConfig` → `LoadConfigured` → `oasfExtractorAdapter`) is never exercised end-to-end. This adds a skip-gated local e2e that mirrors how users actually invoke the extractor enricher.

- New `tests/e2e/local/16_extractor_enricher_test.go` skips unless `~/.agntcy/oasf-sdk/extractor/manifest.json` is present (after `dirctl init`)
- Writes an import YAML with `enricher.extractor: {}` and runs `dirctl import` on the shared A2A agent card fixture **without** `factory.Replace`
- Pulls the resulting record and asserts at least one non-empty skill and domain

Verified locally with OrbStack: `go test -C ./tests/e2e/local . -ginkgo.focus="Extractor enricher"` → 2 Passed.

Closes #1848

### DIFF
--- a/tests/e2e/local/16_extractor_enricher_test.go
+++ b/tests/e2e/local/16_extractor_enricher_test.go
@@ -1,0 +1,106 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package local
+
+import (
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/agntcy/dir/tests/e2e/shared/testdata"
+	"github.com/agntcy/dir/tests/e2e/shared/utils"
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
+)
+
+var _ = ginkgo.Describe("Extractor enricher import", func() {
+	ginkgo.BeforeEach(func() {
+		utils.ResetCLIState()
+	})
+
+	ginkgo.Context("import with enricher.extractor", ginkgo.Ordered, ginkgo.Serial, func() {
+		var (
+			tempDir  string
+			cfgPath  string
+			cardPath string
+			cid      string
+		)
+
+		ginkgo.BeforeAll(func() {
+			// Check for the oasf-sdk extractor manifest written by `dirctl init`.
+			// This mirrors extractor.IsProvisioned without importing the internal package.
+			home, homeErr := os.UserHomeDir()
+			gomega.Expect(homeErr).NotTo(gomega.HaveOccurred())
+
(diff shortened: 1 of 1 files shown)


paralta (approved):
Welcome to the project, and thanks for picking this up so promptly! This is exactly what was needed: a simple and effective end-to-end test for `enricher.extractor`. Approved ✅

Shaurya2k06:
Thanks! Open to trying something more difficult next 🚀

PR #1892, opened 2026-07-25
fix(cli): register cloud KMS providers

### LINKED ISSUE #1882: [Feature]: `dirctl sign` — support cloud KMS key providers (awskms/gcpkms/azurekms)
### Problem Statement

We run an independent `dir` node (apiserver + zot + reconciler) and sign every published OASF 1.0.0 record from an automated, containerized publishing pipeline with no interactive shell. We wanted the signing key held in a cloud KMS, so the raw private key is never materialized on disk or in the process. However, `dirctl sign` only accepts an on-disk key file — passing a cloud-KMS key reference is not recognized.

Environment:
- dir v1.4.0 (apiserver + reconciler, ghcr.io/agntcy/dir-*:v1.4.0)
- dirctl v1.4.0 (linux/arm64, dirctl-linux-arm64 from the v1.4.0 release)
- Deployment: containerized, non-interactive (no TTY)

Steps to reproduce:
1. `dirctl sign <cid> --key awskms:///alias/<key-name>`
2. Observe the error.

Actual: fails with `unrecognized scheme: awskms://` — there is no cloud-KMS signing provider; only a local key file works.
Expected: `--key` accepts cosign-style KMS references so signing can use a managed key without ever exposing the private key.
Impact: we had to keep a cosign private key in a secrets manager and write it to the container filesystem at runtime — weaker key custody than a KMS-backed key.

### Proposed Solution

Wire cosign's existing KMS providers into `dirctl sign` so `--key` accepts awskms://, gcpkms://, azurekms://, and hashivault:// references (cosign already supports these under the hood). Alternatively, if a KMS-backed signing path already exists, documenting it would be enough.

### Alternatives Considered

We fell back to a cosign key-pair stored in a secrets manager and written to the container filesystem at startup with an empty COSIGN_PASSWORD. It works, but the private key is materialized in the runtime environment — exactly what a KMS reference would avoid.

### Additional Context

_No response_

### Checklist

Baltsat:
## Summary

- register Sigstore’s AWS, Azure, GCP, and HashiCorp Vault KMS providers for `dirctl sign`
- add credential-free regression coverage for all four supported KMS prefixes
- keep the provider dependency graph inside the CLI module rather than the shared client library
- keep the repository tests module tidy with its local CLI replacement

## Root cause

`SignerVerifierFromKeyRef` accepts KMS references, but Sigstore providers register through package `init` functions. The `dirctl sign` process did not import those provider packages, so the provider registry was empty and KMS URIs failed as unrecognized schemes.

## Validation

Using Go 1.26.5 on Linux ARM64:

- exact regression fails on the unpatched current `main` with all four providers absent
- `go -C cli test ./cmd/sign -run "^TestKMSProvidersRegistered$" -count=1`
- `go -C cli test ./... -count=1`
- `go -C cli vet ./...`
- `go -C cli build ./...`
- golangci-lint 2.11.3 against the CLI module: `0 issues`
- `go -C cli mod tidy -diff`
- `go -C tests mod tidy -diff`
- `go -C tests list -deps ./...`
- golangci-lint 2.11.3 across the tests module: `0 issues`
- production-file revert makes the regression fail; restoring it makes the test pass
- `git diff --check`

The test verifies registration without cloud credentials or network calls. Live signing still depends on each provider’s credentials and service availability.

Implementation and validation used Codex assistance. I reviewed the final diff, dependency placement, and test results.

Fixes #1882

### DIFF
--- a/.licensei.toml
+++ b/.licensei.toml
@@ -23,6 +23,7 @@ ignored = [
   "github.com/xi2/xz", # Public Domain https://github.com/xi2/xz/blob/master/LICENSE
   "github.com/libp2p/go-libp2p-gorpc", # MIT https://github.com/libp2p/go-libp2p-gorpc/blob/master/LICENSE-MIT
   "github.com/libp2p/go-libp2p-pubsub", # MIT https://github.com/libp2p/go-libp2p-pubsub/blob/master/LICENSE-MIT
+  "github.com/hashicorp/vault/api", # MPL-2.0 https://github.com/hashicorp/vault/blob/main/api/LICENSE
   "go.mongodb.org/mongo-driver", # Apache License 2.0 https://github.com/mongodb/mongo-go-driver/blob/master/LICENSE
   "modernc.org/sqlite", # BSD-3-Clause https://gitlab.com/cznic/sqlite/-/blob/master/LICENSE
   "modernc.org/libc", # BSD-3-Clause https://gitlab.com/cznic/libc/-/blob/master/LICENSE
--- a/cli/cmd/sign/kms_providers.go
+++ b/cli/cmd/sign/kms_providers.go
@@ -1,0 +1,11 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package sign
+
+import (
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/aws"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/azure"
+	_ "github.com/sigstore/sigstore/pkg/signature/kms/gcp"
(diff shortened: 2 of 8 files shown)


csirmazbendeguz (approved):

tkircsi (approved):

tkircsi:
@Baltsat Thanks for your contributing! Could you also update the related documentation in the docs/content/dir? Thanks a lot!

Baltsat:
Thanks — I added the related KMS signing documentation in `381f0ff`:

- CLI reference with supported KMS URI formats and a Google Cloud KMS example
- trust-model coverage for KMS-backed keys
- a usage-guide walkthrough for AWS KMS, Google Cloud KMS, Azure Key Vault, and HashiCorp Vault

Local documentation validation passed: MkDocs build, codespell, pymarkdown, and `git diff --check`.

PR #1927, opened 2026-07-29
feat(dir): separate filter counts & total counts on AI catalog UI

csirmazbendeguz:
let's display the filtered counts separately

the counts in the top right corner will not change with the filters anymore

Screenshot 2026-07-29 at 16 33 40


akijakya (approved):
💯

PR #1931, opened 2026-07-29
feat(cli): add dirctl daemon config init command

### LINKED ISSUE #1779: feat(cli): add a command to dump the default daemon config to disk
## Summary

Add a way to write the daemon's built-in default configuration to disk so users can customize it, instead of authoring a `daemon.config.yaml` from scratch. Today the daemon runs entirely from an **embedded** default and there is no first-class way to obtain an editable copy.

## Motivation

`dirctl daemon` loads its config in one of two ways (`cli/cmd/daemon/config.go`, `loadConfig`):

- `--config <file>` → the file is read **as-is** (no defaults merged), or
- otherwise → the **embedded** `daemon.config.yaml` (`//go:embed`) is used as the complete default.

This means `dirctl daemon start` works out of the box with zero config files (good), but a user who wants to *tweak* one setting has to hand-write a full config that satisfies `--config`'s "read as-is, no merge" contract — there's no supported starting point. The default config filename/location (`<data-dir>/daemon.config.yaml`, data-dir default `~/.agntcy/dir`) is even logged/documented as the conventional path, yet nothing ever creates it, and `loadConfig` does not auto-read it unless `--config` points there.

(Note: this is the **daemon/server** config at `~/.agntcy/dir/daemon.config.yaml`, distinct from the **client** config at `~/.config/dirctl/config.yaml` that `dirctl init` seeds — the two should not be conflated.)

## Proposal

Add a command that writes the embedded default daemon config to disk, e.g.:

- `dirctl daemon config init` — write the embedded default to `<data-dir>/daemon.config.yaml` (respecting `--data-dir`), or to a `--output`/`-o` path.

and/or

- `dirctl daemon start --dump-config` — print the effective/default config to stdout without starting the daemon.

Behavior:

- Idempotent and non-destructive: refuse to overwrite an existing file unless `--force`.
- Writes atomically (temp + rename), consistent with the client config writer.
- The written file is a valid, complete config accepted as-is by `--config` (so the round-trip "dump → edit → `--config`" works).

## Open questions

- Should the daemon **auto-load** `<data-dir>/daemon.config.yaml` when present (falling back to the embedded default when absent), so a dumped-and-edited file is picked up without `--config`? That closes the gap between the logged "config path" and what `loadConfig` actually reads, but changes daemon startup behavior — decide separately.
- Subcommand (`daemon config init`) vs flag (`daemon start --dump-config`) vs both.
- Should `dump-config` emit the pristine embedded default, or the effective config after env/flag overrides are applied?

## Acceptance criteria

- A command writes the embedded default daemon config to a well-known/`--output` path; re-running without `--force` does not overwrite.
- The dumped file is accepted by `dirctl daemon start --config <file>` unchanged.
- `dirctl init` is **not** involved — this stays on the `daemon` command surface (the daemon self-provisions via the embedded default; `init` handles the client-side environment only).

conorbronsdon:
## What

Adds `dirctl daemon config init`, a command that writes the daemon's embedded
default configuration (`daemon.config.yaml`) to disk so it can be edited and
passed back in via `--config`.

## Why

Closes #1779.

`dirctl daemon start` loads configuration in one of two ways: an embedded
default when no `--config` is given, or a user-supplied file read as-is with
no defaults merged. There was no way to get a starting point for a custom
config short of copying the file out of the source tree, so this adds a
command that dumps the exact embedded default to disk.

## How

- New `daemon config` command group with a single `init` subcommand
  (`cli/cmd/daemon/config_init.go`).
- Target path resolution reuses the existing `Options.ConfigFilePath()`
  logic: `<data-dir>/daemon.config.yaml` by default, or the daemon's
  `--config` path if one was passed on the parent command.
- `--output` / `-o` overrides the target path explicitly (same flag/shorthand
  as `dirctl network init`).
- `--force` allows overwriting an existing file; without it the command
  refuses to touch a file that already exists.
- With `--force`, the write goes through the existing `fsutil.WriteAtomic`
  helper (temp file in the same directory + rename), so a failed or
  interrupted write never leaves a partial config file behind.
- Without `--force`, the existence check and the write are a single
  filesystem operation, not a `Stat`-then-write: the new content is staged
  in a temp file and then attached to the target path with `os.Link`, which
  itself fails with `fs.ErrExist` if the path already exists. This closes
  the TOCTOU window a plain "check, then write" would have between a
  concurrent process creating the file and this command overwriting it.
- The dumped content is the embedded `defaultConfigYAML` string verbatim
  (comments included), not a re-serialized/effective config, so it stays
  identical to what `dirctl daemon start` uses when no `--config` is given.
- Documented in `docs/content/dir/dir-cli-reference.md`: a new
  `### dirctl daemon config init` subsection under Daemon Operations, and
  `config init` added to the command-groups summary table.

## Testing

- `go build ./...` (cli module) — clean
- `go vet ./...` (cli module) — clean
- `go test ./cmd/daemon/...` — all passing (9/9), including new tests:
  - writes the default config to the default location
  - refuses to overwrite an existing file without `--force`
  - overwrites with `--force`
  - respects `--output`
  - respects an explicit `--config` set on the parent `daemon` command
  - `writeIfAbsent` rejects an existing file with no partial/leftover temp
    file, exercising the atomic no-force write path directly
  - the dumped file loads unchanged through `loadConfig()` — asserted with
    full-struct equality (`require.Equal`) between the config loaded from
    the embedded default and the config loaded from the dumped file, not
    just a single field
- `gofmt -l` / `goimports -l` — clean
- `golangci-lint run` — could not run locally: the installed 2.5.0 binary is
  built with go1.25 and refuses to lint a module targeting go1.26.5
  (`cli/go.mod`). This is a local toolchain/version mismatch, not specific to
  this change — the same error occurs on unrelated, pre-existing packages in
  the repo.

## Open questions (from the issue, left for maintainer input)

- Should the daemon auto-load `<data-dir>/daemon.config.yaml` when present,
  even without `--config`? Not implemented here — `loadConfig()` behavior is
  unchanged, so a dumped file only takes effect once passed via `--config`.
- This PR only adds the `daemon config init` subcommand from the issue's
  proposal, not a `--dump-config` flag on `daemon start`. The acceptance
  criteria only require one way to get the default config onto disk; happy
  to add the flag too if maintainers want both entry points.


cli/cmd/daemon/config_init.go
```diff
@@ -0,0 +1,163 @@
...
+// configCmd is the parent command for daemon config subcommands.
+var configCmd = &cobra.Command{
+	Use:   "config",
+	Short: "Manage the daemon configuration file",
+}
+
+var configInitOpts = &struct {
+	Output string
+	Force  bool
+}{}
```
tkircsi: The convention across cli/cmd is a named struct type paired with a package-level opts. Please, follow this pattern.

cli/cmd/daemon/config_init.go
```diff
@@ -0,0 +1,163 @@
...
+
+// writeIfAbsent atomically creates path with data, failing with
+// errConfigExists if a file is already there. It stages data in a temp file
+// in the same directory first (like fsutil.WriteAtomic), then links the temp
+// file to path instead of renaming over it: os.Link fails with fs.ErrExist
+// if path already exists, so the existence check and the create happen as a
+// single filesystem operation. A separate os.Stat followed by a write would
+// leave a window where a file created between the two steps gets silently
+// clobbered.
+func writeIfAbsent(path string, data []byte, perm os.FileMode) error {
```
tkircsi: Hm. Maybe we wshould move the cli/internal/agentcfg/fsutil into cli/internal/fsutil and it's test also. Seems to me a common fs package which can be reused by other packages.

cli/cmd/daemon/config_init.go
conorbronsdon: Done in `938b36d3`. Named type plus a package-level opts, matching `cli/cmd/pull`, `push`, `sync`, `search`, `import`, `export`, `validate`, and `network/init`.

One deviation worth flagging: those packages all use the bare names `options`/`opts`, but the `daemon` package already owns `Options`/`opts` for the daemon command itself, so this is `configInitOptions`/`configInitOpts` to avoid shadowing it. Say the word if you would rather it live in its own `cli/cmd/daemon/config/` package so it can use the bare names.

Also added `TestConfigInitCommand_ParsesFlags`, which drives the real command with argv. Every existing test called `runConfigInit` directly with a bare `cobra.Command`, so flag registration was entirely uncovered: a renamed flag or a wrong binding would not have failed anything.

cli/cmd/daemon/config_init.go
conorbronsdon: Agreed, and done in `938b36d3`: `cli/internal/agentcfg/fsutil` is now `cli/internal/fsutil`, with its test. Three import sites updated (`cmd/daemon/config_init.go`, `internal/agentcfg/mcp_engine.go`, `internal/agentcfg/skill_engine.go`), no code changes, and nothing else in the repo referenced the old path.

Your reasoning held up wrt what the package was accumulating. `writeIfAbsent` in this command was a near-duplicate of `fsutil.WriteAtomic`, identical through `MkdirAll → CreateTemp → defer Remove → Write → Close → Chmod` and diverging only at `Link` vs `Rename`. Rather than move that duplicate into the newly shared package too, I replaced it with a single `O_CREATE|O_EXCL` open in `4faf1efc`: it closes the same create-if-absent race in one syscall, and it does not need hardlink support, which the previous version did. That mattered as: `os.Link` fails outright on exFAT, on SMB shares, and in FUSE bind mounts, so `--output` onto a USB stick or a mounted share failed on the default path while `--force` succeeded on the same path.

Two other things came out of the same pass, both in `4faf1efc`:

- `configDirPerm` was `0755` while `daemon start` creates the data directory `0700`. Since this command is meant to run before the daemon has ever started, it is usually what creates `~/.agntcy/dir`, and `MkdirAll` is a no-op on an existing directory, so a later `daemon start` never repairs the mode. The daemon then writes its peer identity key, database, and object store beneath it. Now `0700` on both write paths, with a test; nothing asserted directory mode before, which is why it slipped through.
- The `gosec` suppression carried over into `internal/fsutil` still said the destination "is a known agent config location resolved by the integrate descriptors, not untrusted input". That stopped being true the moment this command started passing `--output` straight through, so the justification is reworded rather than left asserting a guarantee the code no longer has.

tkircsi: @conorbronsdon  Thank you very much for your cooperation! WDYT about give `fsutil` both writers and a mode-carrying options struct, so neither call site has to pre-create directories or re-implement chmod, for example
```go
package fsutil

// ErrExists is returned by WriteNew when the target already exists.
var ErrExists = errors.New("file already exists")

// WriteOptions controls the modes applied by the writers. A zero FileMode
// means 0600; a zero DirMode means 0755.
type WriteOptions struct {
	FileMode os.FileMode
	DirMode  os.FileMode
}

// WriteAtomic replaces path's contents atomically (temp file + rename).
func WriteAtomic(path string, data []byte, o WriteOptions) error

// WriteNew creates path, returning ErrExists if it is already there.
// O_CREATE|O_EXCL makes the existence check and the create one syscall.
func WriteNew(path string, data []byte, o WriteOptions) error
```
Both funnel through one unexported helper that does MkdirAll(dir, o.DirMode), applies FileMode through the descriptor, and removes the partial file on failure.

`config_init.go` then keeps only policy and messaging:

```go
func writeDefaultConfig(target string, force, callerChosePath bool) error {
	wo := fsutil.WriteOptions{FileMode: configFilePerm, DirMode: dirPermFor(callerChosePath)}

	write := fsutil.WriteNew
	if force {
		write = fsutil.WriteAtomic
	}

	if err := write(target, []byte(defaultConfigYAML), wo); err != nil {
		if errors.Is(err, fsutil.ErrExists) {
			return fmt.Errorf("config file already exists at %s (use --force to overwrite)", target)
		}

		return fmt.Errorf("failed to write default config to %s: %w", target, err)
	}

	return nil
}
```
The mechanism tests move with the code: mode bits under a restrictive umask, cleanup after a failed write, unwritable parent, directory target — all become table tests over both writers in `fsutil`, run once instead of once per caller.

conorbronsdon: Done in `ce0fcef` — this is a better shape than what I had, thank you.

`fsutil` now exports both writers over a shared `WriteOptions{FileMode, DirMode}` (zero → 0600 / 0755), funnelling through one unexported `write()`: `MkdirAll` at `o.DirMode`, mode applied through the descriptor, partial file removed on any failure. `WriteNew` returns `fsutil.ErrExists`, so callers match with `errors.Is` and phrase their own message — the writers don't know what the file is for.

`writeDefaultConfig` is now exactly the shape you sketched. `config_init.go` drops `writeIfAbsent`, `errConfigExists`, the pre-create `MkdirAll`, and the `path/filepath` import: **235 lines → 161**, policy and messaging only.

One thing your sketch surfaced that I hadn't noticed: `WriteAtomic` was doing `os.Chmod(tmpName, perm)` on the *path*, while `writeIfAbsent` chmod'd through the descriptor. Sharing one implementation forces them to agree, so the atomic path now goes through the descriptor too. That closes a symlink-swap window on the staging file and drops a `//nolint:gosec`.

The mechanism tests moved as you described — table tests over **both** writers, run once instead of once per caller: mode bits under a restrictive umask, `DirMode` on created parents, zero-value defaults, unwritable parent, unreachable parent, directory target, no temp files left behind. Four caller-side tests deleted as redundant (`writeIfAbsent`'s own test, dir-creation-failure, unwritable-directory, plus one that duplicated the existing both-paths directory-target case). `config_init_test.go` keeps the policy: 0700 for the data dir, 0755 for an `--output` parent, the `--force` messaging, flag parsing.

`setUmask` is build-tagged (`syscall.Umask` doesn't exist on Windows) with a no-op so `GOOS=windows` still compiles.

Verified on Linux rather than just on my Windows box, since the umask and `DirMode` cases only mean anything under POSIX:

```
fsutil            13 tests, PASS (umask + DirMode cases execute for real)
cmd/daemon        11/11 config-init tests PASS
internal/agentcfg PASS  — covers the three other WriteAtomic call sites
gofmt             clean on every file touched
```

`TestWritersLeaveNothingBehindWhenParentIsUnwritable` skips under my WSL root shell via the existing `Geteuid() == 0` guard; it'll run on CI.

The three other `WriteAtomic` callers (`mcp_engine.go`, `skill_engine.go` ×2) took the new signature as `fsutil.WriteOptions{FileMode: ...}`; a zero `DirMode` gives them the 0755 they had hard-coded before, so their behaviour is unchanged.

conorbronsdon:
Correcting myself on the `Test / Unit` failure — my earlier retrigger commit named the wrong test.

The CI task pipes stdout to `/dev/null`, so I reproduced it locally with the exact command and stdout intact:

```
go test -C ./cli -covermode=atomic -coverprofile=... ./...
```

The failure is **`TestValidateCommand_NoFileArgs`** (`cli/cmd/validate/validate_test.go:52`), not `TestEmbeddedZot`:

```
--- FAIL: TestValidateCommand_NoFileArgs (19.31s)
    Received unexpected error: validation error: failed to validate record:
    schema URL validation failed: failed to send POST request to
    https://schema.oasf.outshift.com/api/0.8.0/validate/object/record?missing_recommended=true:
    net/http: TLS handshake timeout
```

It makes an unconditional live HTTPS call to `schema.oasf.outshift.com` during the unit-test job, so it fails whenever that host is slow or unreachable from the runner.

Unrelated to this branch:

- The test was added in `c565fcce` / `bb65b75e`, long before it.
- **No commit here touches `cli/cmd/validate/`** at all.
- In the same local run `cmd/daemon` passed (`TestEmbeddedZot` included) and `internal/fsutil` passed at 78% coverage.

Happy to leave it for a retrigger. If it's worth addressing separately, the durable fix is to gate that test on an env var or an explicit build tag so a remote schema service can't make the unit job red — it's a network integration test living in the unit suite. I didn't touch it here since it's outside this PR's scope, but I can open a separate issue if useful.

Apologies for the earlier misattribution — I named `TestEmbeddedZot` from a local run that had failed for a different reason (zot readiness) and didn't confirm it was the same failure CI hit.

conorbronsdon:
The one red check here — `Test / Unit` — is fixed by #1953, now out of draft.

`cli/cmd/validate`'s unit tests call `https://schema.oasf.outshift.com` live, so the unit job fails on a TLS handshake timeout with no code change. No commit on this branch touches `cli/cmd/validate/`. #1953 points those tests at recorded responses by default, with `OASF_SCHEMA_URL` preserving the live path for contract-drift checks.

Merging #1953 first should clear this one.

tkircsi (approved):
LGTM! Nice job! Thanks!

PR #1936, opened 2026-07-30
fix(catalog): use decimal pagination tokens

### LINKED ISSUE #1935: [Fix] Don't encode AI Catalog's page_token
For some reason, we're base64-encoding the `page_token` on the AI Catalog

Image

This is unnecessary. The `page_token` is an integer, there's no reason to base64-encode this. It's confusing to both the user and the developer

We should just pass the `page_token` as an integer. I don't think this conflicts the ARD spec

lntutor:
Closes #1935

## Summary
- encode catalog pagination tokens as decimal offsets instead of base64
- send decimal page offsets directly from the catalog UI
- regenerate the checked-in static UI bundle

## Verification
- go test ./controller -count=1 (77 passed)
- go vet ./controller
- npm run check (0 errors; 2 pre-existing warnings)
- npm run build
- git diff --check
- independent code review: no findings

The commit includes a DCO sign-off matching the author identity.


csirmazbendeguz (approved):
Thank you for your contribution!

PR #1941, opened 2026-07-30
fix(tests): start the runtime discovery workload after its record exists

### LINKED ISSUE #1780: [Bug]: `Test / E2E / Daemon External` fails intermittently
### Bug Description

The `e2e:test:daemon:external` CI job is failing intermittently. The job exits with code 201 and
reports two errors: `Process completed with exit code 201` and `exit status 1`. The failure occurs
at the `Run end-to-end tests` step and appears to affect the `runtime should discover docker
workloads` spec.

### Expected Behavior

The `Test / E2E / Daemon External` job should pass consistently.

### Affected Version

main

### Steps to Reproduce

Observed on CI across multiple runs of the `e2e:test:daemon:external` task. No reliable local
reproduction steps identified yet — investigation needed.

**Reference failing run:** https://github.com/agntcy/dir/actions/runs/29013902253/job/86109843636?pr=1770#step:11:840

### Checklist

- [x] I have verified this does not duplicate an existing issue

conorbronsdon:
Closes #1780.

`docker compose up` started the labelled `service-a2a` container with the rest of the stack, ten seconds before the daemon. dir-runtime reconciles once at boot, queues every workload it finds, and gives each a bounded resolve budget: six attempts, 15s apart, 30s timeout each, so roughly 75s if attempts fail fast and 255s if they time out. The OASF record that workload's label references can only be pushed once the daemon is serving, from the suite's `BeforeAll`.

On a fast runner the record lands inside the budget and the spec passes. On a slow one the budget expires first, `resolveWorkload` persists `{"error": ...}` into the workload's OASF field, and nothing re-queues it â€” the only two writers to the work queue are boot reconciliation and new docker events. The three-minute `Eventually` cannot recover from that at any timeout, which fits the intermittency and fits why the earlier `BeforeAll` reordering reduced the flake without eliminating it.

## What changed

`service-a2a` moves behind a `runtime-discovery` compose profile, so the testenv no longer starts it. The suite starts it in `BeforeAll`, after the record push, which makes the docker start event the trigger with the record already in the store. The compose file path comes from `test-config.yaml` rather than being hardcoded, since the `default` and `external` testenvs share this Go package and only `external` has such a workload.

The spec is restored, split into registration and resolution. Registration depends only on the daemon seeing the start event; resolution depends on the record being reachable. Separating them means a lost event reads as "the workload never appeared" within a minute instead of a three-minute timeout that looks like a resolver fault. The resolution half asserts the absence of the `"error"` key directly, so an exhausted budget reports itself rather than surfacing as a missing `"cid"`.

It also did not compile as commented â€” `:= discoverResp[0]` had lost its variable name.

## What this does not fix, and what it costs

Registration previously had two independent triggers: boot reconciliation and the start event. The flake was that both fired too early. Deferring the container removes the reconciliation trigger outright, so registration now rests solely on the event stream, and that path is lossy â€” dir-runtime subscribes to docker events without a `since` and only starts watching after `reconcile` returns, so an event landing in that gap is gone. This trades a recoverable-looking flake for a rarer unrecoverable one. **A periodic re-reconcile in dir-runtime would make the ordering irrelevant, and is the real fix.** I'm happy to open that upstream if you agree it belongs there.

The mechanism above is derived from reading dir-runtime v1.3.4, not from logs of a failing run. Every step is in the source, but no failing run is cited.

## On re-enabling the CI job

The matrix entry is re-enabled here. Two caveats:

Its TODO gave a different reason from the flake â€” "possible schema drift on the runtime side." That is a real and separate risk this change does not touch: the resolution assertion compares the resolved record against the fixture with `MatchJSON`, and a serialization drift fails that regardless of ordering. I re-enabled it because a disabled job cannot tell us whether either problem is real, and dir-runtime has moved 1.3.3 â†’ 1.3.4 since it was turned off.

Also worth knowing: the job has been dark longer than the spec has been commented out, so `push`, `pull`, and the cosign sign/verify specs in this suite have not run in this environment either.

**If you would rather not have it back in the matrix yet, say so and I will drop that hunk â€” the ordering fix stands on its own.**

## Verification

Run locally, not just in CI. The bug is timing-dependent, so a single green run proves nothing; both orderings were given the same handicap, a 90s delay before the record push, standing in for a slow runner. That is longer than the resolver's ~75s minimum budget.

| | old ordering | with this change |
|---|---|---|
| boot reconcile | `workloads_registered=1` | `workloads_registered=0` |
| registration | at boot, before the record can exist | `event_type=added`, after the push |
| resolver | attempts 1&rarr;6, every one `record not found`, exhausted | resolves first try |
| suite | **FAIL** &mdash; 3 passed, 1 failed | **PASS** &mdash; 4 passed, 0 failed |
| exit code | **201** | 0 |

The old-ordering run exits **201**, the code #1780 reports. The resolver gave up about ten seconds before the record landed, and nothing re-queued it:

```
attempt=6 max_attempts=6 error="failed to pull OASF record baeareiabbog2...:
  rpc error: code = NotFound desc = failed to pull record: record not found"
```

The failure also lands exactly where the split puts it: under the old ordering the registration spec still passes, and only resolution fails. The workload was never undiscovered, it was discovered too early to resolve.

Unhandicapped, the suite passes locally as it does in CI: 4 of 6 specs, 0 failed. Two skip for want of cosign.

**No schema drift.** The `MatchJSON` comparison against the fixture passes at dir-runtime v1.3.4, locally and in CI, which answers the question the disabled job's TODO raised.

Environment: Docker 29.1.3, Compose 2.40.3, Go 1.26.5, Ubuntu 24.04. `go vet ./e2e/daemon/...` clean. The ordering claims were checked against the dir-runtime v1.3.4 source, then confirmed against its logs.

### DIFF
--- a/.github/workflows/reusable-test-e2e.yaml
+++ b/.github/workflows/reusable-test-e2e.yaml
@@ -28,10 +28,8 @@ jobs:
             label: "Local"
           - task: e2e:test:network:local
             label: "Network"
-          # TODO: runtime tests disabled for now due to possible schema drift on the runtime side.
-          # TODO: investigate for details to reenable the test.
-          # - task: e2e:test:daemon:external
-          #   label: "Daemon External"
+          - task: e2e:test:daemon:external
+            label: "Daemon External"
           - task: e2e:test:integration-server:default
             label: "Integration Server"
     steps:
--- a/tests/e2e/daemon/01_daemon_test.go
+++ b/tests/e2e/daemon/01_daemon_test.go
@@ -5,9 +5,12 @@ package daemon
 
 import (
 	"context"
+	"encoding/json"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
+	"time"
 
 	corev1 "github.com/agntcy/dir/api/core/v1"
 	signv1 "github.com/agntcy/dir/api/sign/v1"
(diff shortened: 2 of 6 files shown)


conorbronsdon:
One more result worth adding, since it validates a design choice rather than just the fix.

Running the old ordering under the 90s handicap, the failure reports itself like this:

```
Summarizing 1 Failure:
  [FAIL] Daemon e2e runtime workflow [It] runtime should resolve the workload's OASF record

  [FAILED] Timed out after 180.004s.
  resolver gave up before the record was reachable: failed to pull OASF record
  baeareiabbog2umgduqhlcb64fzt6adn34kblzvru3fdzkl75hjhwt6h3da: rpc error:
  code = NotFound desc = failed to pull record: record not found
```

Two things that were guesses when I opened this are now observed.

The `NotTo(HaveKey("error"))` assertion is what fires, so the diagnostic names the cause outright. Without it the same run fails on a missing `cid` key, which reads as "the resolver produced the wrong record" when what actually happened is that it gave up minutes earlier. That was the argument for asserting the error key directly, and this is it working.

The registration spec **passes** in that same run; only resolution fails. That is the whole reason for splitting one spec into two. The workload was never undiscovered — it was discovered too early to resolve — and the two specs now say which. A single combined spec would have failed at the three-minute mark with no way to tell those apart, which is roughly the position #1780 started from.

The suite exits 201 here, matching the code in the issue report.

paralta (approved):
Thanks for the investigation on this bug and for the explanation provided 🚀 Keeping in mind the periodic re-reconcile in `dir-runtime` as the long term fix for this issue, I believe your PR provides a solid pragmatic solution for now.

PR #1942, opened 2026-07-30
fix(config): preserve unknown top-level config sections across writes

### LINKED ISSUE #1777: fix(config): preserve unknown top-level config sections across writes (lossless round-trip)
## Summary

The `dirctl` client config writers (`SaveFile` and its callers `SaveContext`, `SaveExtractor`, `SetCurrentContext`) decode the file into the `config.File` struct and re-marshal it. Any **top-level** YAML key not represented on `File` is dropped on write. This is lossy for forward/extended config: an older `dirctl` binary that writes the config would silently discard a top-level section added by a newer version (or another command).

Surfaced in review of #1762 / PR #1764.

## Current behavior (not a bug today, but a latent risk)

- The only top-level sections today are `current_context`, `contexts`, and `extractor` — all present on `File`, so nothing is lost in practice.
- `LoadFile` uses strict `KnownFields(true)` and **rejects** unknown top-level keys (see `TestLoadFile` "rejects unknown fields"), so such a config can't even be loaded by the strict path.
- The `loadOptionalFile(..., allowUnknownFields=true)` path tolerates unknown keys but does not preserve them across a save.

## Why it's not a trivial fix

A naive fix — an `,inline` catch-all map on `File` — would **absorb** unknown fields and thereby break the strict-rejection contract that `LoadFile`/`TestLoadFile` rely on. The real work is to separate two concerns that are currently coupled:

- **strict validation** (reject typos / unknown keys when a user explicitly loads their config), and
- **lossless round-tripping** (a writer must not clobber sections it doesn't own).

## Possible approaches

- Decode into a `yaml.Node`, merge the known sections back into it on write, and re-encode — preserving unknown nodes while keeping a strict *validation* pass separate.
- Or add an inline catch-all for the round-trip path while keeping a separate strict decode for the explicit-load/validate path.

## Acceptance criteria

- A config file containing an unknown top-level section survives a `SaveContext` / `SaveExtractor` / `SetCurrentContext` write with that section intact.
- The existing strict "rejects unknown fields" behavior of `LoadFile` is preserved (a round-trip test plus the existing rejection test both pass).

## Notes

- Atomic writes were already addressed in PR #1764 (`SaveFile` now writes temp + rename).

conorbronsdon:
Closes #1777.

`SaveFile` and its callers (`SaveContext`, `SaveExtractor`, `SetCurrentContext`, `ClearExtractor`) decode the config into the typed `File` struct and re-marshal it, so any top-level YAML key not modeled on `File` is dropped on write. An older `dirctl` silently discards a section a newer one wrote.

The issue notes why the obvious fix is wrong: an `,inline` catch-all on `File` would absorb unknown keys and break the strict-rejection contract `LoadFile` and `TestLoadFile` rely on. So this separates the two concerns instead, taking the `yaml.Node` approach the issue suggests.

## How it works

The tolerant load path (`loadFile` with `allowUnknownFields=true`) parses the raw bytes into a `yaml.Node` tree alongside the typed decode, and records every top-level key not represented on `File` as its original nodes. `SaveFile` re-appends them after encoding the known fields. Strict `LoadFile` is untouched and still rejects unknown keys — capture is gated on the tolerant flag and is unreachable from the strict path, since `Decode` has already errored.

Scope is top-level only, as the issue specifies. All four writers go through the tolerant path, so the preservation applies wherever it matters.

## Two hazards this had to handle

An adversarial review pass caught both, and each would have produced a config **no version of dirctl could load** — strictly worse than the dropped section this fixes. They're the reason for the second commit.

**Aliases.** Recognized sections are re-encoded from the struct and emit no anchors. An unknown section containing an alias to an anchor defined in a recognized section — `contexts: dev: &devctx` plus `future: {copied_from: *devctx}` — was written out dangling, and every later load failed with `yaml: unknown anchor 'devctx' referenced`. Captured nodes now have aliases inlined and anchors stripped. That's value-preserving: an alias means exactly the node it references, and only the indirection is lost. A `visiting` set guards the self-referential case.

**Key-set derivation.** `knownTopLevelKeys` read yaml struct tags but skipped untagged fields, while the encoder falls back to the lowercased field name. A future untagged field would therefore be captured as unknown *and* emitted by the struct marshal — a duplicate top-level key, which is a load error. Latent today since all three fields are tagged, but it's precisely the trap the next contributor would hit. The derivation now uses the encoder's own fallback.

## Verification

`go test ./config/...` passes. The alias fix has a regression test that reproduces the exact original failure when the inlining is removed, and the four preservation tests were mutation-checked — removing the re-append in `marshalFile`, or no-oping the capture, turns them red. `golangci-lint run ./config/...` is at the same baseline as `upstream/main` (10 goconst, plus 2 gci that are a local CRLF artifact): no new issues.

## Known limits, stated rather than implied

- **Single document.** A second YAML document after `---` was dropped before this change and still is; both the decoder and the capture pass see only the first.
- **A leading file comment** does not survive, since it attaches to the first known key, which is re-encoded. Comments attached to *unknown* sections do survive, including inline ones.
- **`SaveFile` is still lossy for a hand-built `File`.** Only a `File` from the tolerant load carries the sections. Every in-repo writer does; the doc comment now says so, since it's exported client API.

Happy to squash the two commits if you'd prefer a single one — I kept them separate because the second is a distinct set of fixes with its own reasoning.

### DIFF
--- a/client/config/config.go
+++ b/client/config/config.go
@@ -5,10 +5,12 @@
 package config
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"sort"
 	"strconv"
 	"strings"
@@ -35,6 +37,22 @@ type File struct {
 	CurrentContext string             `yaml:"current_context"`
 	Contexts       map[string]Context `yaml:"contexts"`
 	Extractor      *Extractor         `yaml:"extractor,omitempty"`
+
+	// unknown holds top-level YAML sections this binary does not recognize,
+	// captured verbatim during a tolerant load so they survive a save. It keeps
+	// the file forward-compatible: an older dirctl must not silently drop
+	// sections written by a newer one. It is populated only by the tolerant load
+	// path (see loadFile); the strict LoadFile still rejects unknown keys. The
+	// field is unexported so yaml marshal/unmarshal ignore it, and SaveFile
+	// re-injects the sections explicitly.
+	unknown []unknownSection
+}
+
+// unknownSection is a top-level key/value pair captured verbatim as YAML nodes so
+// its value, nesting, and any comments are preserved across a write.
+type unknownSection struct {
+	key   *yaml.Node
+	value *yaml.Node
 }
 
 // Extractor is the machine-wide OASF taxonomy extractor provisioning record,
@@ -142,13 +160,12 @@ func loadFile(path string, allowUnknownFields bool) (*File, error) {
 		path = defaultPath
 	}
 
-	f, err := os.Open(path)
+	data, err := os.ReadFile(path)
 	if err != nil {
 		return nil, fmt.Errorf("failed to open client config file %s: %w", path, err)
 	}
-	defer f.Close()
 
-	decoder := yaml.NewDecoder(f)
+	decoder := yaml.NewDecoder(bytes.NewReader(data))
 	decoder.KnownFields(!allowUnknownFields)
 
 	file := &File{}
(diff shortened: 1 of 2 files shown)


tkircsi (approved):
LGTM! Thank you for contributing!

PR #1953, opened 2026-07-31
test(cli): stop the validate unit tests calling a live schema service

conorbronsdon:
Draft — happy to drop it if you'd rather keep the live call.

## The problem

`cli/cmd/validate`'s unit tests point `--url` at `https://schema.oasf.outshift.com`, so the **unit** job depends on a remote host being reachable and fast. It goes red with no code change:

```
--- FAIL: TestValidateCommand_NoFileArgs (19.31s)
    schema URL validation failed: POST https://schema.oasf.outshift.com/api/0.8.0/validate/object/record...
    net/http: TLS handshake timeout
```

I hit this on an unrelated PR (#1931) and initially misdiagnosed it, because `task test:unit:coverage` pipes stdout to `/dev/null` — the CI log names the failing *task*, never the failing *test*. Worth flagging on its own: any failure in `./cli` currently surfaces as `exit status 1` with no test name.

## The change

Tests point at a per-test `httptest` server by default. The responses it serves are **recorded from the real service** (2026-07-31), one per request fixture, committed under `testdata/schema_response_*.json`. Recorded rather than invented, so the fake can't drift into asserting a protocol that doesn't exist.

The live path stays available and is the one to use for contract drift:

```
OASF_SCHEMA_URL=https://schema.oasf.outshift.com go test ./cmd/validate/...
```

Dispatch is on the record's semantic content, not a byte compare, so reformatting a fixture can't silently turn every response into the valid one. An unrecognised body returns 501 rather than inheriting another fixture's answer.

## Verification

```
offline           ok  0.108s   (was 19.3s and failing)
OASF_SCHEMA_URL   ok  8.123s   <- the fake matches the live service
```

Mutation-tested rather than assumed: making the fake rubber-stamp everything as valid fails `TestValidateCommand_RealFiles`.

**Known gap, stated rather than papered over:** mutating the 501 branch to return the valid response leaves the suite green. That guard protects a *future* fixture added without a recorded response; nothing today exercises it. Pinning it would need a fixture whose only purpose is to be unrecognised, which seemed worse than the gap.

## Judgement calls for you

- **Is losing the live check by default acceptable?** My read is that schema-contract coverage belongs in e2e, and the unit job shouldn't be the thing that notices the schema service is down. But that's your call, and the env override exists so the live check is one variable away.
- **Recorded responses go stale.** They'd need re-recording when the OASF schema version bumps. Alternative: keep the live call but gate it behind a build tag, so it runs deliberately rather than on every unit run.

Separate from #1931 on purpose — no overlap, different files.


akijakya (commented):
LGTM, the OASF API can be considered stable now, so the response format is not expected to change soon, and there is still live validation on record pushing in the e2e test suite.

One note: `api/core/v1/record_test.go:138` also calls the live OASF server for validation, maybe that also worth to be adjusted.

conorbronsdon:
Split out as #1981: while checking I noticed `cli/cmd/root_test.go:61` also runs validate against the live server in the unit suite (it can't fail offline since the assertion only checks for the OIDC error, but it's still a network call on every run). Can do small follow-up for that if helpful.

akijakya:
> Split out as #1981: while checking I noticed `cli/cmd/root_test.go:61` also runs validate against the live server in the unit suite (it can't fail offline since the assertion only checks for the OIDC error, but it's still a network call on every run). Can do small follow-up for that if helpful.

There is also another place I mentioned in my previous comment, I think the ones that might fail in case of a network/server error could be handled in this PR.

conorbronsdon:
> > Split out as #1981: while checking I noticed `cli/cmd/root_test.go:61` also runs validate against the live server in the unit suite (it can't fail offline since the assertion only checks for the OIDC error, but it's still a network call on every run). Can do small follow-up for that if helpful.
> 
> There is also another place I mentioned in my previous comment, I think the ones that might fail in case of a network/server error could be handled in this PR.

@akijakya I’ve folded the api/core/v1/record_test.go change from #1981 into this PR. TestRecord_Validate now contacts a live server only when OASF_SCHEMA_URL is explicitly set, so the default unit run stays offline. The existing fake-validator tests and live record-push e2e coverage remain in place.

I believe that makes #1981 redundant, so I’ll close it as superseded by this PR. The non-failing root-command cleanup remains separate in #1985.

akijakya (approved):
LGTM, thanks for the contribution!

paralta:
@conorbronsdon can you sign this commit https://github.com/agntcy/dir/pull/1953/commits/371ea86204534fbe18560934027247ddd430ae7e? DCO check is currently failing, let's merge this once that is fixed 😄

conorbronsdon:
Done - rebased with sign-offs on both commits, DCO should be green now. Thanks @paralta!

Issue #1939, opened 2026-07-30
SearchService: return total match count alongside paginated results


akijakya:
## Summary
`SearchCIDsRequest` and `SearchRecordsRequest` already support `limit` and `offset` (`proto/agntcy/dir/search/v1/search_service.proto`), but the responses carry no total match count. Any client that wants to paginate has no way to show "page 1 of N" or a total-results figure without streaming the entire result set just to count it — which defeats the purpose of pagination.

## Motivation
The AI Catalog UI gets this right via `AIFinderService.ListAgents`, which returns `total_count` from `CountCatalogEntries`. But that endpoint is a catalog-specific projection (`CatalogEntry`) and excludes any record without an MCP/A2A/AgentSkills module, so it's not usable by general clients (CLI, TUIs such as lazydir, scripts) that list *all* records via `SearchService`.

The counting machinery already exists on the catalog side — it just isn't exposed on the general search API.

## Proposal
Add a total count to the search responses. Two shapes to consider:
- A `uint32 total_count` on a response header/first message of the stream, or
- A dedicated lightweight `CountRecords(SearchRecordsRequest) returns (CountResponse)` RPC.

The count should honor the same `queries` as the request but ignore `limit`/`offset` (i.e. total matches, not page size), mirroring `CountCatalogEntries`.

## References
- `proto/agntcy/dir/search/v1/search_service.proto` — `limit`/`offset` already present
- `server/database/gorm/catalog.go` — `CountCatalogEntries` (existing count precedent)

akijakya:
Related: #1940 (filter negation). Together these enable full server-side tri-state filtering + counted pagination for non-UI clients.

Coooder-Crypto:
Hi, I'd like to work on this issue.

I propose adding a dedicated unary RPC: rpc CountRecords(CountRecordsRequest) returns (CountRecordsResponse); CountRecordsRequest would contain only the repeated RecordQuery queries, and CountRecordsResponse would expose uint32 total_count.

This avoids changing the semantics of the existing streaming responses, supports empty/out-of-range pages, and keeps the API compatible with Buf's request/response naming and uniqueness rules. The database implementation would reuse the existing record filters, count distinct record CIDs, and ignore pagination and sorting.

Does this API shape sound good before I start implementing it?

akijakya:
[...]

Issue #1909, opened 2026-07-28
[Test]: e2e for POST /v1/search parity and POST /v1/extract


akijakya:
### Problem Statement

The extractor-backed endpoints need end-to-end coverage against a real node, in both deployment shapes: a locally provisioned extractor (in-process assets) and a remote OASF-SDK server. Handler unit tests use a fake extractor and cannot catch resolution, wiring, or deployment regressions.

Rescoped: covers `POST /v1/search` (parity) and `POST /v1/extract` instead of the original extract-then-filter endpoint.

### Proposed Solution

Extend the e2e suite to cover both endpoints in both extractor modes.

**`POST /v1/extract`**
- Returns skills and domains above threshold for a descriptive phrase
- Returned class names exist in the OASF taxonomy for the node's schema version
- 503 when no extractor is configured

**`POST /v1/search`**
- Returns relevance-ranked entries for a free-text query over a seeded record set
- **Parity assertion:** the same query through `dirctl search "q"` and `POST /v1/search` returns the same CIDs in the same order — this is the endpoint's contract, so it belongs in e2e rather than unit tests
- Pagination: page 1 and page 2 are disjoint and together cover the ranked set
- Ranking is stable across repeated identical requests (guards the tie-determinism fix)
- Empty-extraction fallback returns name-substring matches
- 503 when no extractor is configured

**Modes**
- **Local:** extractor assets provisioned on the node (`dirctl daemon start` path)
- **Remote:** `oasf-sdk` deployed via the chart, gateway auto-wired to it (kind testenv already enables the subchart by default)

### Notes

- The remote path needs the OASF-SDK server to actually serve `ExtractorService`, which requires `OASF_SDK_EXTRACTOR_OASF_URL` to be set — the chart does this, but the model is downloaded from HuggingFace at pod start, so the remote suite depends on egress being available in CI.
- Seed records must carry skills and domains the extractor plausibly returns, otherwise assertions become tautological. Prefer asserting on ranking and parity over exact class names.

### Acceptance criteria

- [ ] e2e coverage for `POST /v1/extract` in local and remote extractor modes
- [ ] e2e coverage for `POST /v1/search` in local and remote extractor modes
- [ ] Parity assertion between `dirctl search` and `POST /v1/search`
- [ ] Pagination and ranking-stability assertions
- [ ] 503 assertions when no extractor is configured

### Depends on

#1905 and #2001.

Issue #1908, opened 2026-07-28
[Feature]: AI Catalog UI: recommend tags from a description (extractor-backed)


akijakya:
### Problem Statement

The AI Catalog's tag filter bar matches tag names as substrings. That works only if the user already knows the OASF taxonomy. Someone who types what they actually want — "analyzing customer support tickets" — gets no matches, because no tag is named that. They are left guessing at hundreds of hierarchical labels.

Replaces the original scope of this issue (wiring the search bar to `/v1/search`). The UI does not consume `/v1/search`; it recommends tags and filters with them.

### Proposed Solution

Keep one search bar. Add a **suggest** action that reinterprets the same text through the extractor (`POST /v1/extract`, #2001) and offers matching skills and domains as selectable tags.

```
[ analyzing customer support tickets            ] [ Suggest tags ]

  Matching tags
    (none)
  ────────────────────────────────────────────
  Suggested from your description
    natural_language_processing/text_classification   12 records
    customer_support/ticket_triage                     4 records
```

Selecting a suggestion behaves exactly like selecting any other tag — it becomes a chip, and filtering still goes through `GET /v1/agents?filter=`. Nothing about the filtering path changes; only tag *discovery* does.

### Design decisions

**One search bar, not two.** Both modes serve one intent — find a tag. Splitting the input by matching strategy exposes an implementation detail and forces an unsure user to choose an algorithm before knowing which works. The natural flow is escalation: type, scan, find nothing, ask for help.

**Suggestions are added below, not swapped in.** Replacing the substring matches creates a hidden mode: results no longer contain the typed text and there is no obvious way back. A labeled section avoids the mode entirely.

**Empty state drives discovery.** When substring matching returns nothing (exactly what happens when a user types a phrase), surface the action inline — "No tags match that — try suggesting from your description." This is the main path users will find the feature through.

**Explicit trigger, not per-keystroke.** Extraction is model inference and half-typed words are semantically meaningless. Use a button; if it should feel automatic, debounce ~500ms and require a space or ~15 characters.

**Drop suggestions with no records.** A suggested tag matching zero records is worse than no suggestion — the user takes our advice and lands on an empty catalog. The UI already loads `GET /v1/tags` for the typeahead, so intersect against that list client-side and drop the rest. Showing record counts alongside makes the choice informed (needs a count on `CatalogTag`, tracked separately).

**Skills and domains only.** Keywords come back from `/v1/extract` but have no tag to map to; do not render them as selectable.

[...]

Issue #1904, opened 2026-07-28
[Refactor]: Shared pluggable OASF extractor (local assets or remote gRPC) + Decompose in client


akijakya:
### Problem Statement

The OASF extractor is wired only into the CLI (`cli/internal/extractor`), as an in-process library over locally-provisioned assets, and the natural-language decomposition lives in `cli/internal/nlsearch` + `cli/cmd/search/nlsearch.go`. The gateway (which will serve UI search) is in the `server` module and cannot reach any of it. We also need a second way to reach the extractor — a **remote gRPC OASF-SDK server** — without the caller knowing which backend it got.

### Proposed Solution

Introduce a shared, backend-agnostic extractor in the `client` module (imported by both `cli` and `server`):

- `client/extractor`:
  - `Extractor` interface: `Extract(ctx, text, opts) (extractor.Result, error)`.
  - `localExtractor` — wraps the in-process `oasf-sdk/pkg/extractor` over provisioned assets (moved from `cli/internal/extractor`).
  - `remoteExtractor` — gRPC client for `agntcy.oasfsdk.extractor.v1.ExtractorService`, mapping `ExtractResponse` → `extractor.Result` (near 1:1: `skills/domains/modules/keywords`, `ScoredClass` fields).
  - `ResolveExtractor(cfg)` — **remote-first when an endpoint is configured** (one health probe, cached), else local assets, else a typed error with an actionable message.
- Extend the shared config record `client/config.Extractor` with a `RemoteAddr` field, so the remote-vs-local choice is a **shared** concern read by every caller (see #1906), not gateway-specific:
  ```go
  type Extractor struct {
      OASFURL  string `yaml:"oasf_url"`
      AssetDir string `yaml:"asset_dir"`
      RemoteAddr string `yaml:"remote_addr"` // NEW: gRPC OASF-SDK server; empty => local assets
  }
  ```
  `ResolveExtractor` reads `RemoteAddr` to decide the backend.
- `client/nlsearch`: move `Decompose` (text → skill/domain/keyword signals) here so both the CLI and the gateway share it. `Decompose` takes the resolved `Extractor` interface and stays **config-agnostic** — config is consumed only by `ResolveExtractor`, one layer up.
- Refactor `dirctl search` and `dirctl routing` natural-language paths to consume the shared package. Output behavior is unchanged for the local-assets case; the new `RemoteAddr` path is wired in #1906.

[...]

PR #1969, opened 2026-08-04
feat(api): post search endpoint

### LINKED ISSUE #1905: [Feature]: Backend POST /v1/search — natural-language search with dirctl parity
### Problem Statement

There is no natural-language search endpoint on the API. `dirctl search "some phrase"` already does this well — it decomposes the phrase with the OASF extractor, fans out one query per signal, and ranks results by how many signals matched — but that capability is CLI-only.

This issue brings that exact capability to the API, with **feature parity as the acceptance bar**: the same query against the same node must return the same records in the same order through either path.

Supersedes the original extract-then-filter design in this issue. See the epic (#1903) for why the UI no longer consumes this endpoint.

### Proposed Solution

**1. Share the fan-out.** Today `fanOutAndScore` / `collectCIDs` / `collectKeywordCIDs` live in `cli/cmd/search/nlsearch.go`. Move them into `utils/nlsearch` (where `Decompose` already lives), abstracted over how CIDs are fetched:

```go
// Searcher runs one RecordQuery and returns matching CIDs, bounded by limit.
type Searcher interface {
    SearchCIDs(ctx context.Context, q *searchv1.RecordQuery, limit uint32) ([]string, error)
}

func FanOutAndScore(ctx context.Context, signals []Signal, s Searcher, opts FanOutOptions) Ranked
```

The CLI adapter wraps `client.SearchCIDs` (gRPC); the gateway adapter wraps `db.GetRecordCIDs` (in-process). Both then run identical matching and ranking code — parity by construction rather than by convention.

**2. Fix tie determinism (required).** The current scorer builds its ranked list in goroutine-completion order, then stable-sorts by hit count, so equal-scoring records order differently between runs. Invisible in a one-shot CLI print; a correctness bug once results are paginated (page 2 can repeat or skip page 1 rows). Impose a total order: hit count desc → summed signal score desc → CID asc.

**3. Endpoint.**

```protobuf
rpc SearchAgents(SearchAgentsRequest) returns (SearchAgentsResponse) {
  option (google.api.http) = { post: "/v1/search"; body: "*" };
}
message SearchAgentsRequest {
  string query = 1;       // free text, required
  uint32 page_size = 2;
  string page_token = 3;
}
message SearchAgentsResponse {
  repeated CatalogEntry results = 1;
  string next_page_token = 2;
}
```

Handler: decompose → `FanOutAndScore` → slice the page from the ranked list → hydrate that page's CIDs into `CatalogEntry` preserving rank order.

**4. Behavior inherited from `dirctl search`:** signals are OR-united (not ANDed), keyword signals fan out to `NAME` ∪ `DESCRIPTION` and are deduped, each signal is capped at 500 candidates, per-signal query sort is recency. No extractor configured → `codes.Unavailable` (HTTP 503).

### Non-goals

- **No `displayName` substring fallback.** Originally specified, now dropped: it answers a different question than the one asked and would return hits where `dirctl search` returns none. An unusable query is rejected instead, with the advice the CLI gives.
- **No `filter` parameter.** The original design had one; `dirctl search` ignores its facet flags in NL mode, so a filter param would break parity and reintroduce the facet-after-pagination problem. Structured filtering stays on `GET /v1/agents?filter=`.
- **Not consumed by the AI Catalog UI.** The UI filters via tags (see #1903). This endpoint serves API and CLI-parity consumers.

### Acceptance criteria

- [ ] Fan-out and scoring live in `utils/nlsearch`, used by both `cli/cmd/search` and the gateway controller
- [ ] Ranking is deterministic across runs (tie order is total and stable)
- [ ] `POST /v1/search` returns relevance-ranked `CatalogEntry` for a free-text query, paginated
- [ ] Same query through `dirctl search "q"` and `POST /v1/search` yields the same CIDs in the same order
- [ ] A query the extractor derives no signals from is rejected with `INVALID_ARGUMENT` and rephrasing advice, not an empty page
- [ ] Extractor unavailable → HTTP 503 with an actionable message
- [ ] Unit tests for fan-out (union, hit-count ranking, keyword dedup, per-signal errors, determinism) and the handler (rank order preserved, pagination disjoint and complete)

### Depends on

Extractor foundation (#1906, #1907) — the gateway needs a resolved extractor before this endpoint can work.

### LINKED ISSUE #1906: [Feature]: Extractor config (RemoteAddr) + resolver wiring for dirctl and gateway
### Problem Statement

The remote-vs-local extractor choice must be usable by **every** caller of the shared resolver, not just the gateway. Today:

- `dirctl` already reads an extractor config section (`client/config.Extractor{OASFURL, AssetDir}`) via `LoadConfigured()`, but it can only build a **local** extractor — there is no way to point `dirctl search` at a running OASF-SDK server.
- `dirctl daemon start` embeds only the OASF *validator*, not the extractor at all.

We need one config switch (`RemoteAddr`, added to the shared record in #1904) that selects remote vs local, read symmetrically by `dirctl` and the gateway, without the code caring where it runs.

### Proposed Solution

Wire `ResolveExtractor(cfg)` into all three callers, each reading its own config and handing the resolved `Extractor` to the config-agnostic `nlsearch`:

1. **`dirctl` (CLI search / routing / import enrichment):** `LoadConfigured()` passes the persisted `RemoteAddr` through to `ResolveExtractor`. `RemoteAddr` set ⇒ `dirctl` uses a remote OASF-SDK server; empty ⇒ local provisioned assets (today's behavior). Extend `dirctl init` so a user can configure a server address instead of downloading the ~89 MB assets.
2. **Standalone gateway / apiserver (cloud):** add the extractor config surface and resolve from it:
   ```yaml
   extractor:
     remoteAddr: ""   # e.g. "oasf-sdk:<grpc-port>" in cloud; empty => local assets
     assetDir:  ""    # local fallback (defaults to ~/.agntcy/oasf-sdk/extractor)
   ```
3. **`dirctl daemon start` embedded gateway (local):** resolve from the same config — `remoteAddr` empty ⇒ reuse the `dirctl init` assets; a local OASF-SDK server address ⇒ use that.

Surface a clear startup/first-use log line stating which backend was resolved.

### Acceptance criteria

- [ ] `dirctl search` respects `RemoteAddr`: set ⇒ remote OASF-SDK server, empty ⇒ local assets (no behavior change).
- [ ] `dirctl init` can configure a remote extractor address as an alternative to downloading assets.
- [ ] Gateway/apiserver: `remoteAddr` set ⇒ remote extractor; empty ⇒ local assets.
- [ ] `dirctl daemon start` + provisioned assets serves `/v1/search` end-to-end locally.
- [ ] Resolved backend is logged for each caller.

Depends on the shared pluggable extractor issue (#1904, which adds `RemoteAddr` to `client/config.Extractor`). Blocks the Helm deployment issue.

### LINKED ISSUE #1907: [Feature]: Helm: optional OASF-SDK extractor server deployment + config
### Problem Statement

The directory Helm chart has no extractor. In the cloud we want an OASF-SDK gRPC server deployed alongside the directory, with the gateway pointed at it. The OASF-SDK repo already publishes a Helm chart + image, so this is wiring, not building a service.

### Proposed Solution

- Add `oasf-sdk` as an **optional** subchart dependency of the apiserver chart (`oci://ghcr.io/agntcy/oasf-sdk/helm-charts`), gated `condition: oasfSdk.enabled`, mirroring the existing optional `oasf` (schema-server) subchart.
- Default `oasfSdk.enabled: false` (local uses provisioned assets); the prod gitops `values.yaml` flips it `true`.
- Set the apiserver Deployment's `extractor.remoteAddr` to the OASF-SDK service DNS when enabled.
- Document enabling it in the gitops prod values (the deployment repo).

### Acceptance criteria

- [ ] `helm template` with `oasfSdk.enabled=true` renders the OASF-SDK Deployment/Service and points `extractor.remoteAddr` at it.
- [ ] Default render (`false`) is unchanged from today.
- [ ] README/values documented.

Depends on the extractor config + daemon/gateway wiring issue. Needed by the remote-mode e2e.

### Service exposure

The subchart must deploy the OASF-SDK server as **`ClusterIP`**, not `NodePort`. The gateway reaches it through cluster DNS, so a node port buys nothing, and the server serves unauthenticated plaintext gRPC — exposing it on every node IP would put free-text user queries outside the cluster.

The upstream chart cannot currently do this: it defaults to `NodePort` and always renders `nodePort`, which Kubernetes rejects on a `ClusterIP` service. Tracked upstream in agntcy/oasf-sdk#179 — this issue depends on that fix (or on pinning a chart version that includes it).

- [ ] Rendered OASF-SDK Service is `ClusterIP` with no `nodePort`

akijakya:
Closes #1905
Closes #1906
Closes #1907

## Summary

Adds a natural-language search endpoint to the AI Catalog — **`POST /v1/search`** — so the UI (and other clients) can send free text like *"a skill for reviewing code"* and get back relevance-oriented `CatalogEntry` results, without exposing the OASF extractor or the streaming `SearchService` to the browser.

This is the **backend → config → deploy** spine of epic #1903 (natural-language search in the AI Catalog UI). It builds on the shared pluggable extractor from #1904 (#1925). UI wiring (#1908) and e2e (#1909) are separate follow-ups.

## What it does (extract-then-filter, v1)

`POST /v1/search { query, page_size, page_token, filter }` on the AI Finder gateway:

1. Extract the top skills/domains from `query` via the OASF extractor.
2. Turn them into `RecordQuery`s and run one in-process content search (`GetRecordCIDs`).
3. Hydrate the CIDs into `CatalogEntry` (same shape as `ListAgents`, so the UI only swaps data source).
4. Apply the optional `filter` facets (`type` / `verified` / `trusted` / `safe` / `tags`) via the existing AND-grammar parser.
5. Empty extraction → fall back to a `displayName` substring match so the user still gets results.
6. No extractor configured → `codes.Unavailable` (HTTP **503**) with an actionable message.

The extractor is pluggable — **local** in-process assets (`dirctl daemon start`) or a **remote** OASF-SDK gRPC server (cloud) — resolved once at gateway startup and injected into the controller.

## Why the extractor package was refactored *again*

#1904 (#1925) introduced the shared pluggable extractor, but it landed in the **`client`** module (the dir gRPC SDK). This endpoint needs the extractor in the **server**, and importing `client` into `server` was the wrong direction:

- the server would depend on its own SDK client, and
- because `client/extractor` → `client/config` → the root `client` package, it dragged the entire SDK dependency closure (sigstore/cosign, k8s client-go, go-tuf, zitadel/oidc, … ~100 modules) into the server binary.

So this PR moves the **SDK-agnostic core** — `Extractor`, `ResolveExtractor`, the local/remote backends, `Config`, `Provision` — into the genuinely-shared **`utils`** module as `utils/extractor` + `utils/nlsearch`. The two convenience functions that read the persisted `dirctl` config (`ResolveConfigured`, `LoadConfigured`) stay in **`client/config`**, since they legitimately depend on the CLI config layer.

Result: **`server` imports `utils/extractor` only** — `go list -deps ./server/...` no longer references `agntcy/dir/client`, and the heavy SDK deps are gone from the server. (Visible in the diff: `cli/cmd/import/config.go` now calls `clientconfig.LoadConfigured()` instead of `extractor.LoadConfigured()` — same behavior, correct home. `utils/extractor` never imports `client`, which is what preserves the layering.)

## Config & deployment

| Scenario | How the extractor is resolved |
|---|---|
| **Local node** (`dirctl daemon start`) | Gateway on by default; empty extractor config → in-process **local** assets at `~/.agntcy/oasf-sdk/extractor`. Works out of the box. |
| **Local + remote server** | `DIRECTORY_DAEMON_SERVER_EXTRACTOR_REMOTE_ADDR=host:port` → remote backend. |
| **Cloud (Helm)** | Optional `oasf-sdk` subchart (off by default). `oasf-sdk.enabled=true` deploys it; the gateway auto-wires `extractor.remote_addr` to the in-cluster service, and `OASF_SDK_EXTRACTOR_OASF_URL` enables the extractor gRPC service on the server. |

New server config block: `extractor: { remote_addr, asset_dir, oasf_url }` (env `DIRECTORY_SERVER_EXTRACTOR_*`), consulted only when the HTTP gateway is enabled. Resolution failure is **non-fatal** — the server still starts; `/v1/search` returns 503 until an extractor is configured.

## Commits

- `feat(api)` — `SearchAgents` RPC (`POST /v1/search`) + regenerated bindings
- `refactor(extractor)` — move the extractor + nlsearch from `client` to `utils` (see above)
- `feat(server)` — extract-then-filter `POST /v1/search` handler + tests
- `feat(server)` — resolve the OASF extractor at gateway startup (#1906)
- `feat(cli)` — configure a remote extractor in `dirctl init` (#1906)
- `feat(helm)` — optional OASF-SDK extractor subchart + gateway auto-wiring (#1907)
- `feat(extractor)` — **default to two score tiers everywhere the extractor is used** (`extractor.DefaultTiers`): gateway `/v1/search`, `dirctl search`/routing, and import enrichment now widen recall to the two closest score groups by default; callers can still narrow (`Tiers: 1`) or widen per query
- `refactor(extractor)` — **unify on one resolution path** (`ResolveExtractor` / `ResolveConfigured`): drop the local-only `LoadConfigured`, so every caller resolves *remote if an address is configured, else local (configured or default dir), else error*. Import enrichment now works against a **remote** OASF-SDK server too (with `remote_addr` in the import file config, a per-`Extract` timeout, and `Close()` on completion)
- `fix` — review feedback: shutdown ordering (close extractor after `GracefulStop`), resolver-leak on gateway-init failure, remote smoke-check timeout, log query length instead of text, chart-doc corrections

## Testing

- **Unit/integration** — handler tests with a fake extractor + in-memory store (relevance order, facet composition, empty-extraction fallback, `503` when unavailable, pagination round-trip); `nlsearch` tier tests; server config-binding tests. `golangci-lint` clean across touched modules.
- **Manual** — `dirctl daemon start` → `POST /v1/search` returns 200 with in-process extraction; separately confirmed the published OASF-SDK server serves `ExtractorService` and extracts correctly once it has the model. Helm rendering verified (`helm template` + `helm lint --with-subcharts`).

## Design notes & follow-ups

- **Skills AND domains — intentional for v1.** A record must match an extracted skill **and** an extracted domain (fields are AND-ed; multiple values *within* a field are OR-ed). This is a deliberate precision-first choice for the first cut, not a defect. A natural future enhancement is a user-selectable **breadth** control that ORs the signals (fan-out + union) to widen recall on demand.
- **Relevance ranking.** The catalog already has usage-metrics ranking — `popularity_score` (pull + lookup counts) and `provider_count`, with working `POPULARITY` / `PROVIDER_COUNT` / `RECENCY` sort modes. What is not yet implemented is *semantic / query-match* relevance: `SORT_MODE_RELEVANCE` is a scaffold that currently falls back to recency order, so `/v1/search` results are recency-ordered for now. Follow-up: rank by match quality (how many extracted signals a record hit) with `popularity_score` as a tiebreaker — or, as an interim step, sort the endpoint by `POPULARITY`. (`total_count` for search responses is tracked separately in #1939.)
- **Cloud model provisioning.** The published `oasf-sdk` image downloads the embedding model from HuggingFace at pod start (the chart can't yet mount pre-provisioned assets). Fine where the cluster has HF egress; a model-baked image is the robust path.
- **UI (#1908) / e2e (#1909)** remain separate; the search bar still calls `GET /v1/agents` until #1908 wires it to `/v1/search`.


akijakya:
Closing this in favour of a reworked plan for the epic (#1903). The design changed after review, so the search half of this branch no longer reflects where we are going. The extractor half does, and will be re-opened as its own PR.

### What changed

This PR implemented **extract-then-filter**: free text goes to the backend, the extractor decides which records match, and results come back ranked. Building and testing it surfaced three problems with putting the extractor in the decision seat:

- **Not explainable.** "cloud computing" maps to `devops_cloud_infra/*` and `technology/cloud_computing`, never to a skill literally named `cloud_computing`. A node holding four matching records returned nothing, and neither the user nor the developer could see why.
- **Ranking is unproven.** `SORT_MODE_RELEVANCE` is a scaffold that falls back to recency. An endpoint that silently ranks makes a promise the backend cannot currently keep.
- **Filters and inferred search compose badly.** Facets applied after the ranked page is cut produce under-filled pages with a next-page token still attached (raised in review); applied before, they constrain a ranking the user never asked for.

The new plan lets the extractor **suggest** instead of **decide**:

- **UI** — the tag filter bar keeps substring matching, plus a suggest action that reinterprets the typed text through the extractor and offers skills and domains as selectable tags (#1908, backed by the new `POST /v1/extract`, #2001). Filtering itself stays on `GET /v1/agents?filter=`, unchanged. The user confirms the taxonomy, so intent is explicit and results are explainable.
- **API** — `POST /v1/search` stays, rescoped to **feature parity with `dirctl search "query"`** (#1905): the CLI's fan-out and hit-count ranking move into `utils/nlsearch` and are shared by both callers, rather than the API having its own matching algorithm. It takes no `filter` parameter, which also removes the facet-after-pagination problem entirely.

### What is being kept

Most of this branch is unaffected by the redesign and will land as a separate **extractor foundation** PR:

| Commit | |
|---|---|
| `01dbcfad` | move the OASF extractor and `nlsearch` from `client` to `utils` |
| `678e49ec` | resolve the extractor for the gateway (#1906) |
| `5b2b6c3c` | configure a remote extractor in `dirctl init` (#1906) |
| `ed17426f` | optional OASF-SDK subchart (#1907) |
| `eaf84c6a` | default to two score tiers |
| `f9b2d844` | unify extractor resolution, remote import enrichment |

#1906 and #1907 are fully implemented and independent of the search design, so there is no reason to hold them behind this.

Dropped: the `SearchAgents` proto (needs reworking without `filter`) and the extract-then-filter handler.

### Review feedback

Thanks @coderabbitai — the review items were addressed in `300e6ef9`, and the fixes to the extractor, `dirctl init`, and the chart carry over to the foundation PR. Two items are resolved by the redesign rather than by a patch: the facet-after-pagination issue disappears with the `filter` parameter, and the insecure gRPC dial to the OASF-SDK server is being handled by making that service cluster-internal (`ClusterIP`) rather than `NodePort`, tracked separately.

PR #1978, opened 2026-08-05
feat(cli): support batch delete via variadic CIDs and --stdin

### LINKED ISSUE #1974: [Feature]: Support batch delete via variadic CIDs and `--stdin`
### Problem Statement

`dirctl delete` accepts exactly one CID — `if len(args) != 1` in `cli/cmd/delete/delete.go`.
Nothing below the CLI imposes that limit:

- The RPC is already client-streaming (`server/controller/store.go:175` loops on `Recv`).
- The SDK already exposes `DeleteBatch` and `DeleteStream` (`client/store.go:250-288`).

It is also inconsistent with the rest of the CLI: `push`, `sync create`,
`routing publish`, and `routing unpublish` all take `--stdin`, and `routing unpublish`
additionally takes variadic CIDs. `routing unpublish`'s own help text tells the user to
"use `dirctl delete` to remove the record entirely" — pointing at a command that cannot
accept its output.

The cost lands on cleanup and policy workflows: deleting a result set requires a shell
`while read` loop, spawning one process and one gRPC connection per record.

### Proposed Solution

Mirror `cli/cmd/routing/unpublish.go`, which already implements this pattern:

- `Use: "delete <cid> [cid...]"`, with `Args` allowing variadic CIDs, or
  `MaximumNArgs(0)` when `--stdin` is set.
- `--stdin` reading both a JSON array from `dirctl search --output json` and
  line-delimited CIDs.
- Route through `client.DeleteBatch` / `DeleteStream` so a batch uses one stream.
- Update the help examples, and drop the caveat in `routing unpublish`'s long help once
  the two commands compose.

Enables `dirctl search --format cid --output json | dirctl delete --stdin`, matching the
existing unpublish example.

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

paralta:
`dirctl delete` accepted exactly one CID even though the RPC is client-streaming and the SDK already exposes `DeleteBatch`, so cleanup workflows needed a shell loop spawning a process per record. Delete now takes variadic CIDs and a `--stdin` flag reading either a JSON array from `dirctl search --output json` or line-delimited CIDs, sending the batch over a single stream.

- Single-CID output is unchanged — `--output raw` still emits just the CID, since delete's help advertises that for scripting. Only batches get the new array shape.
- Stdin parsing, deduplication, and argument validation shared with `routing publish`/`unpublish` moved into a shared package instead of being copied a third time.
- `sync create --stdin` is left alone: same flag name, but it reads routing search responses for the peer multiaddrs a CID list can't provide.

### DIFF
--- a/cli/cmd/delete/delete.go
+++ b/cli/cmd/delete/delete.go
@@ -10,6 +10,7 @@ import (
 
 	corev1 "github.com/agntcy/dir/api/core/v1"
 	"github.com/agntcy/dir/cli/presenter"
+	cidsUtils "github.com/agntcy/dir/cli/util/cids"
 	ctxUtils "github.com/agntcy/dir/cli/util/context"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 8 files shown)


akijakya (approved):
LGTM! 🚀

PR #1981, opened 2026-08-05
test(api): gate the live-validation record test behind OASF_SCHEMA_URL

conorbronsdon:
Follow-up to #1953, from @akijakya's review note there: `TestRecord_Validate` in `api/core/v1/record_test.go` also calls the live OASF schema server, so the default unit run for that module still depends on the host being reachable. It now runs only when `OASF_SCHEMA_URL` is set (the variable #1953 introduces for the cli validate tests) and skips otherwise. The hermetic `fakeValidator` tests in the same file keep covering the validation plumbing by default; live-schema coverage for record pushes stays in the e2e suite.

Verified both ways: the default `go test ./core/v1/...` (from `api/`) skips the live test and passes offline in 0.008s; with `OASF_SCHEMA_URL=https://schema.oasf.outshift.com` the integration test runs its five cases and passes.

Note on checks: the `Test / E2E / Test Client` failure is a pre-existing intermittent race in `tests/e2e/client/02_events_test.go` (a case-sensitive assertion loses to `RST_STREAM ... CANCEL`), in a module this PR can't reach — the same job passes on the base commit on main.


conorbronsdon:
Now redundant with #1953 per comment here https://github.com/agntcy/dir/pull/1953#issuecomment-5243782369

PR #1982, opened 2026-08-05
fix(ui): remove all media type filter

### LINKED ISSUE #1928: [Fix] Remove "All" Media Type filter
The AI Catalog includes an "All" filter in Media Type:

Image

I don't think this is neccessary. The other categories (Status, Tags) don't have this either.

Would it be better UX to remove it?

dlanov:
## Summary

Closes #1928.

- remove the visible "All" option from the Media Type filters
- preserve the existing unfiltered behavior when no media types are selected
- simplify individual media type selection now that "All" is no longer user-selectable
- rebuild the checked-in catalog UI assets

## Verification

- `npm run check` — 0 errors; two pre-existing warnings
- `npm run build`
- `git diff --check`

`npm run lint` remains blocked by eight pre-existing Prettier failures in unrelated files.


akijakya (approved):
LGTM, thanks for the contribution!

PR #1983, opened 2026-08-06
feat(search): support negating a RecordQuery for exclude filters

### LINKED ISSUE #1940: SearchService: support negating a RecordQuery (exclude filters)
## Summary
`RecordQuery` can only express inclusion — "records that match value X". There is no way to express exclusion — "records that do NOT match value X". This blocks tri-state filtering (include / exclude / off) in clients built on `SearchService`.

## Proposal
Add a `bool negate = 3;` field to `RecordQuery` (`proto/agntcy/dir/search/v1/record_query.proto`):

```proto
message RecordQuery {
  RecordQueryType type = 1;
  string value = 2;
  // When true, exclude records matching this query instead of including them.
  bool negate = 3;
}
```

Default `false` preserves current behavior (backwards-compatible). This is the minimal addition that enables tri-state filters and maps 1:1 to include/exclude/off. (A general `MatchOperator` enum was considered but is more than tri-state needs.)

## Implementation notes (important — correctness)
Negation is NOT a simple `LIKE` → `NOT LIKE` flip for the JOIN-based fields.

Today a skill include compiles to (`server/database/gorm/record.go:426`):

    JOIN skills ON skills.record_cid = records.record_cid WHERE skills.name LIKE 'nlp'

Negating that predicate is wrong: a record with skills `[nlp, python]` still matches via the `python` row, giving "has some other skill" rather than "has no nlp skill". Correct exclusion for multi-valued/joined fields must use a subquery:

    NOT EXISTS (SELECT 1 FROM skills
                WHERE skills.record_cid = records.record_cid AND skills.name LIKE 'nlp')

The codebase already uses exactly this pattern for Trusted (`record.go:575-577`). Negation should reuse the `EXISTS`/`NOT EXISTS` shape for all joined/multi-valued fields (skills, domains, modules, locators, annotations); scalar columns on `records` (name, version, schema_version, description) can use plain `NOT (...)`.

Also: the current builder aggregates all values of a field into one OR'd condition (`BuildWildcardCondition("skills.name", cfg.SkillNames)`), so include and exclude of the same field can't share one slice. `QueryToFilters` (`server/database/utils/utils.go:99`) and `RecordFilters` (`server/types/search.go:6`) would need to separate them (e.g. `SkillNames` vs `ExcludedSkillNames`) or compile each negated query as its own `NOT EXISTS` clause.

## Complements
This pairs with the request to expose a total match count on `SearchService` (paginated listing), enabling full server-side tri-state filtering + pagination for non-UI clients.

## References
- `proto/agntcy/dir/search/v1/record_query.proto`
- `server/database/gorm/record.go` (filter compilation; `EXISTS`/`NOT EXISTS` precedent at L575)
- `server/database/utils/utils.go` — `QueryToFilters`
- `server/types/search.go` — `RecordFilters`

akijakya:
## Summary

- add `bool negate` to `RecordQuery` so `SearchService` clients can express "does NOT match X" alongside the existing "matches X"
- for joined/multi-valued fields (skills, domains, modules, locators, annotations), compile exclusions to a correlated `NOT EXISTS` subquery rather than negating the include-path `JOIN` — negating the `JOIN` would wrongly resurrect a record via an unrelated matching row (e.g. excluding skill "nlp" on a record with skills `[nlp, python]` must still exclude it, not match it via the "python" row)
- for nullable scalar columns (`description`, `schema_version`, `oasf_created_at`, `authors`), use a NULL-safe `NOT(...)` that retains NULL rows instead of silently dropping them
- for boolean query types (`VERIFIED`, `TRUSTED`, `SCAN_SAFE`), flip the parsed bool rather than adding new storage
- regenerate the Go/gRPC API stubs and add database, utils, and controller test coverage

## Motivation

`RecordQuery` could only express inclusion — "records that match value X". There was no way to express exclusion — "records that do NOT match value X" — which blocked tri-state filtering (include / exclude / off) in clients built on `SearchService`.

## Compatibility

Default `negate: false` preserves current behavior exactly (wire- and behavior-compatible) — this is an additive, backward-compatible change.

## Validation

- `task lint`
- `task test`

Closes #1940


muscariello (approved):
LGTM

PR #1985, opened 2026-08-06
test(cli): stop the root OIDC test calling the live schema server

conorbronsdon:
Follow-up to #1953, from the review thread there: the validate case in `TestLocalCommandsDoNotRequireOIDCToken` still handed `--url` the live schema service, so the default `cli/cmd` unit run made a real network call on every run. It couldn't go red offline — the assertion only checks the error isn't the OIDC one — but the request was real.

The fix reuses the `OASF_SCHEMA_URL` opt-in convention #1953 introduces: set it to run against a live service, otherwise a local httptest server answers. Unlike the validate package's server this one is deliberately body-blind — the test asserts on the auth path, not validation semantics, so recorded-response dispatch would be machinery with no assertion behind it. The response bytes are byte-identical to the `schema_response_valid.json` fixture #1953 adds; they're inlined here because that fixture isn't on main until #1953 merges and `go:embed` can't reference a parent directory, which keeps this PR green in either merge order.

Verified both ways: the default run passes offline with the request landing on the local server (`POST /api/0.8.0/validate/object/record?missing_recommended=true`, confirmed via temporary instrumentation); with `OASF_SCHEMA_URL=https://schema.oasf.outshift.com` the case passes against the live service.


muscariello:
Hey @conorbronsdon are you aware that we have a discord channel ?
It would be useful if you could join there.

conorbronsdon:
> Hey @conorbronsdon are you aware that we have a discord channel ?
> It would be useful if you could join there.

Happy to join

lgecse (approved):
LGTM!

PR #1988, opened 2026-08-06
feat(routing): add publish all support

### LINKED ISSUE #1965: [Feature]: Add `dirctl routing publish --all` flag
### Problem Statement

A Directory may contain records that are not published to the DHT network. This is expected. `dirctl routing publish` is a separate step that may or may not be run after `dirctl push`, `dirctl import`, etc.

For public Directory nodes, like Outhsift's Directory (http://ads.outshift.io/), this is not desired. Since it's a public node, we expect all records to be published to the network. Outshift's Directory hosts ~2000 records, most of which are not published

If there are a lot of existing records not published, manually publishing them may be infeasible.

### Proposed Solution

Extend the `dirctl routing publish` command with the `--all` flag.

By running `dirctl routing publish --all`, users can publish all unpublished records.

Note: the command should ask the user for confirmation:

```
This operation will publish every unpublished record. Would you like to proceed?
```

The command should also accept a `--yes` flag to skip the confirmation

### Alternatives Considered

Use a Bash script that fetches all CIDs and calls `dirctl routing publish` for each.
Functionally, it's the same.
The downside is, the Bash script doesn't work with the hardened `dirctl` image.

### Checklist

dlanov:
## Summary

- add an `all_records` publication request to the routing API
- add `dirctl routing publish --all` with an interactive confirmation
- add `--yes` to support non-interactive bulk publication
- publish every currently stored record; publication is idempotent
- refuse interactive `--all` prompting on non-terminal stdin unless `--yes` is provided
- reject conflicting CID, `--stdin`, `--all`, and `--yes` input modes
- add focused CLI, controller, and publication-worker tests

## Testing

- `(cd proto && go run github.com/bufbuild/buf/cmd/buf@v1.66.1 dep update)`
- `(cd proto && go run github.com/bufbuild/buf/cmd/buf@v1.66.1 lint)`
- `(cd proto && go run github.com/bufbuild/buf/cmd/buf@v1.66.1 generate --template buf.gen.yaml)`
- `golangci-lint run --config .golangci.yml ./cmd/routing`
- `golangci-lint run --config .golangci.yml ./publication`
- `go test ./cmd/routing -count=1`
- `go test ./publication -count=1`
- regenerated the routing API bindings and verified regeneration produces no changes

Fixes #1965


cli/cmd/routing/publish.go
```diff
@@ -115,5 +155,76 @@ func runPublishCommand(cmd *cobra.Command, args []string) error {
...
+func confirmPublishAllIfNeeded(cmd *cobra.Command) (bool, error) {
+	if publishOpts.Yes {
+		return true, nil
+	}
+
+	return confirmPublishAll(cmd)
+}
+
+func confirmPublishAll(cmd *cobra.Command) (bool, error) {
+	presenter.Printf(cmd, "%s [y/N]: ", publishAllPrompt)
```
paralta:
```suggestion
	presenter.PrintSmartf(cmd, "%s [y/N]: ", publishAllPrompt)
```

cli/cmd/routing/publish.go
```diff
@@ -115,5 +155,76 @@ func runPublishCommand(cmd *cobra.Command, args []string) error {
...
+}
+
+func runPublishAllCommand(cmd *cobra.Command, c *client.Client) error {
+	confirmed, err := confirmPublishAllIfNeeded(cmd)
+	if err != nil {
+		return err
+	}
+
+	if !confirmed {
+		presenter.Printf(cmd, "Aborted. No records were submitted for publication.\n")
```
paralta:
```suggestion
		presenter.PrintSmartf(cmd, "Aborted. No records were submitted for publication.\n")
```

paralta (approved):
LGTM, thanks for the contribution! Just one small fix before we merge: use `presenter.PrintSmartf` here instead of `Printf` — the former routes to stderr for structured output formats, so the prompt and abort message don't end up inside the JSON stream on `--all --output json`.

cli/cmd/routing/publish.go
```diff
@@ -14,25 +15,29 @@ import (
...
 Key Features:
-- Network announcement: Makes record discoverable by other peers
-- Local storage: Stores record in local routing index
-- DHT announcement: Announces record and labels to distributed network
+- Network announcement: Makes records discoverable by other peers
+- Local storage: Stores records in the local routing index
+- DHT announcement: Announces records and labels to the distributed network
 - Background retry: Failed announcements are retried automatically
 - Batch publication: Submit multiple CIDs in one request
+- Bulk publication: Publish every currently unpublished stored record
```
tkircsi:
```suggestion
- Bulk publication: Publish every currently stored record
```

cli/cmd/routing/publish.go
```diff
@@ -45,25 +50,60 @@ Usage examples:
 3. Publish records from stdin (JSON array or line-delimited CIDs):
    dirctl search --format cid --limit 100 --output json | dirctl routing publish --stdin
 
-4. Output formats:
+4. Publish every currently unpublished stored record:
```
tkircsi:
```suggestion
4. Publish every currently stored record:
```

cli/cmd/routing/publish.go
```diff
@@ -115,5 +155,76 @@ func runPublishCommand(cmd *cobra.Command, args []string) error {
...
+	reader := bufio.NewReader(cmd.InOrStdin())
+
+	line, err := reader.ReadString('\n')
+	if err != nil && line == "" {
+		return false, fmt.Errorf("read confirmation: %w", err)
+	}
+
+	answer := strings.ToLower(strings.TrimSpace(line))
+
+	return answer == "y" || answer == "yes", nil
```
tkircsi: `confirmPublishAll` reads from `cmd.InOrStdin()` and errors with "read confirmation" on EOF, so in CI or a piped context `--all` without `--yes` fails on a read error rather than a clear "refusing to prompt, pass --yes" message. A TTY check would read better.

tkircsi (changes requested):
Thanks a lot for contributing! Please, check the routing Publish process. I think, the unpublished filtering is not needed, because Publish is idempotent.

cli/cmd/routing/publish.go
```diff
@@ -115,5 +157,83 @@ func runPublishCommand(cmd *cobra.Command, args []string) error {
...
+	if !ok || !term.IsTerminal(int(in.Fd())) { //nolint:gosec // G115: a file descriptor fits in an int.
+		return false, errors.New(
+			"refusing to prompt for --all on non-terminal stdin; pass --yes to continue",
+		)
+	}
+
+	return confirmPublishAll(cmd)
+}
+
+func confirmPublishAll(cmd *cobra.Command) (bool, error) {
```
tkircsi: `confirmPublishAll` is a duplicate of the unexported `confirm` in `cli/cmd/install/install.go` Canwe export it to a shared helper in something like `cli/util/prompt.`

dlanov:
@tkircsi Thanks for the review. I addressed the requested changes:

- `--all` now publishes every stored record directly without filtering against the routing layer, since publication is idempotent.
- Updated the CLI and API wording accordingly.
- Added a TTY check so non-interactive `--all` requires `--yes`.
- Moved the duplicated yes/no confirmation logic into the shared `cli/util/prompt` helper and reused it from install and routing.
- Added and updated focused tests for the shared prompt and publish behavior.

All checks are passing. Could you please take another look when you have a chance?

tkircsi (approved):
LGTM. @dlanov Thank You very much!

PR #2004, opened 2026-08-11
feat(skill): fix and update dir skills

tkircsi:
## Summary

Fixes incorrect guidance in the `agntcy-dir` agent skill and fills gaps that
caused agents to fail or report success wrongly. All behaviour below was
verified against dirctl v1.6.3, not read off the structs.

**`authoring.md` — import enrichment was documented wrong.** The no-LLM path
told users to set `enricher.skip_enricher: true` with `skills`/`domains`
directly under `enricher:`. `skip_enricher` does not exist in the codebase, and
the real schema is `enricher.llm` / `enricher.static` / `enricher.extractor`.
Because the import config parser is not strict, the bogus keys were silently
discarded and import fell back to the LLM enricher — so following the
documentation verbatim fails with `AZURE_OPENAI_API_KEY is required`, the exact
opposite of the promised no-credentials path. Replaced with the three real
blocks, added the previously undocumented `extractor` (LLM-free, needs
`dirctl init`), and added a full `import.config.yaml` reference with the
flags-override-file precedence and the `dry_run: true` gotcha.

**`setup.md` — daemon configuration.** `dirctl daemon config init` was missing
entirely. Documented it, plus three silent failure modes: the config is not
auto-discovered (`daemon start` ignores `<data-dir>/daemon.config.yaml` unless
`--config` is passed), `--config` replaces the defaults rather than merging,
and `--data-dir`/`--config` are parent flags so `status`/`stop` need the same
`--data-dir` as `start`. Added a throwaway-test-setup workflow for agents
validating configs without touching the user's directory.

**`setup.md` — contexts and auth.** Documented all 15 context keys (only 4 were
listed) and that the parser rejects unknown keys, that there is no
`context add`/`create` subcommand, and `doctor.bootstrap_peers`. For auth:
bare `auth login` when a context supplies the issuer, the `--force` requirement
after changing `oidc_scopes` (login is a no-op with a cached token, so scope
changes silently do not apply), and how to read the `auth status` token line.

**`SKILL.md` — authentication is not authorization.** A valid token says
nothing about what the caller may do. Gateways enforce per-method access
control, so `PermissionDenied` on `push`/`sign`/`routing publish` is an
entitlement problem that re-logging-in cannot fix. Added as a CLI convention
plus a dispatch row for auth intents.

### DIFF
--- a/.skill/SKILL.md
+++ b/.skill/SKILL.md
@@ -23,6 +23,7 @@ and publish" → authoring + publishing).
 | User intent (examples)                                                                                      | Read first                                               |
 | ----------------------------------------------------------------------------------------------------------- | -------------------------------------------------------- |
 | "install dirctl", "set up a local directory", "start the daemon", "configure contexts", "connection issues" | [references/setup.md](references/setup.md)               |
+| "log in", "am I authenticated?", "switch to another directory node", "permission denied / forbidden"        | [references/setup.md](references/setup.md)               |
 | "create a record", "validate my record", "import MCP servers / A2A cards / skills into DIR"                 | [references/authoring.md](references/authoring.md)       |
 | "push", "publish to the network", "sign my record", "prove name ownership"                                  | [references/publishing.md](references/publishing.md)     |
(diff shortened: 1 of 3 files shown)


akijakya (approved):
LGTM 🚀

PR #2015, opened 2026-08-12
feat(dir): add extractor to import-records workflow

### LINKED ISSUE #1735: feat(dirctl): Extractor-based enrichment for 'dirctl import' + adopt in import-records workflow
Builds on dir#1743 (`dirctl init` pre-provisioning).

## Summary

Wire the **Extractor** into `dirctl import` as an alternative to the MCP-server enrichment path. When selected, `dirctl import` builds an Extractor from the locally-provisioned assets (`pkg/extractor`, loaded **in-process**) and constructs the importer's Enricher with it, instead of the MCP + LLM pipeline. Then adopt this in `.github/workflows/import-records.yaml`.

## Motivation

- Deterministic, model-based enrichment for skills/domains without the LLM/MCP round-trip. The model is loaded once per `dirctl import` process from local files (warm ~150 ms) and reused across records — no separate service to run.

## Scope

- Add a `dirctl import` configuration path (flag/config) that:
  - loads `pkg/extractor` in-process from the local asset dir (provisioned by `dirctl init`, dir#1743), erroring clearly with "run `dirctl init`" if the assets are missing, and
  - constructs the importer's Enricher with that Extractor (depends on the dir-importer Enricher change, agntcy/dir-importer#58).
- Keep the MCP/LLM enrichment as the alternative; make the choice **explicit** (flag/config), defaulting to preserve current behavior until validated.
- Update `.github/workflows/import-records.yaml` to run enrichment via the Extractor: run `dirctl init` to pre-provision the assets, and cache `~/.agntcy/oasf-sdk/extractor` across CI runs (key on model + taxonomy version).

## Depends on

- agntcy/oasf-sdk#170 (extractor library)
- agntcy/dir-importer#58 (Enricher accepts an Extractor)
- dir#1743 (`dirctl init` pre-provisioning)

## Acceptance criteria

- [ ] `dirctl import` can enrich via the in-process Extractor as an explicit alternative to the MCP path.
- [ ] Missing assets produce a clear "run `dirctl init`" error.
- [ ] `import-records.yaml` provisions via `dirctl init` and caches the asset dir across CI runs.
- [ ] The MCP-based enrichment path continues to work unchanged when selected.

csirmazbendeguz:
NB: this will make extractor the default in the import workflow

### DIFF
--- a/.github/workflows/import-records.yaml
+++ b/.github/workflows/import-records.yaml
@@ -39,6 +39,13 @@ on:
         description: "Enable dirctl import debug output (dedup and validation details)"
         type: boolean
         default: false
+      enrichment:
+        description: "Record enrichment backend: extractor (local model) or llm (Azure OpenAI + MCP)"
+        type: choice
+        options:
+          - extractor
+          - llm
+        default: extractor
       sign_max_retries:
         description: "Per-CID signature attempts before deleting unsigned records"
         type: number
@@ -69,6 +76,7 @@ jobs:
       sign: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.sign || 'true' }}
       force: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.force || 'false' }}
       debug: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.debug || 'false' }}
+      enrichment: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.enrichment || 'extractor' }}
       sign_max_retries: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.sign_max_retries || '3' }}
       sign_cleanup_on_failure: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.sign_cleanup_on_failure || 'true' }}
     env:
@@ -162,6 +170,7 @@ jobs:
       SIGN: ${{ needs.setup-matrix.outputs.sign }}
       FORCE: ${{ needs.setup-matrix.outputs.force }}
       DEBUG: ${{ needs.setup-matrix.outputs.debug }}
+      ENRICHMENT: ${{ needs.setup-matrix.outputs.enrichment }}
       SIGN_MAX_RETRIES: ${{ needs.setup-matrix.outputs.sign_max_retries }}
       SIGN_CLEANUP_ON_FAILURE: ${{ needs.setup-matrix.outputs.sign_cleanup_on_failure }}
     steps:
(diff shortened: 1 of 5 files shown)


akijakya:
The caching is worth it for the model weights (~89 MB, never change), but the **static key is the problem** when a new OASF version ships.

`sdk.Provision` is documented as idempotent — it "skips the model load and re-embed when the on-disk assets **already match** the endpoint and taxonomy," implying it contacts the schema server to check. So:

1. Cache hit with old assets → `dirctl init` runs → SDK detects new OASF version → re-provisions (re-embeds taxonomy) → ✅ correct output this run.
2. BUT `actions/cache` v5 **does not save on a cache hit** by default. The key `oasf-extractor-linux-v1` already exists → updated assets are never written back to cache.
3. Next run: same stale cache hit → same re-provision from scratch → same no-save. Every run after a schema change pays the full re-embed cost, defeating the cache.

Secondary concern: `IsProvisioned` only checks for `manifest.json` existence — it doesn't validate the schema version embedded in assets. So if the SDK's staleness check isn't actually contacting the server, stale taxonomy could silently persist for the lifetime of the cache entry.

**Suggested fix**: use a run-unique primary key with a prefix-based `restore-keys` fallback so you always restore the latest warm cache and always save fresh assets:

```yaml
- name: Cache OASF extractor assets
  uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25
  with:
    path: ~/.agntcy/oasf-sdk/extractor
    key: oasf-extractor-${{ runner.os }}-${{ github.run_id }}
    restore-keys: |
      oasf-extractor-${{ runner.os }}-
```

Cache miss on the exact key (always), but `restore-keys` gives a warm start from the previous run's save, and the post-step saves whatever `dirctl init` left on disk — including any re-provision triggered by a new schema. Each run's entry expires via GitHub's 7-day LRU. Cost: one cache entry per run instead of one forever, but you stay current automatically.

.github/workflows/import-records.yaml
```diff
@@ -368,6 +405,23 @@ jobs:
...
+      - name: Provision OASF extractor
+        if: ${{ env.ENRICHMENT == 'extractor' }}
+        run: |
+          set -euo pipefail
+          EXTRACTOR_ASSET_DIR="${HOME}/.agntcy/oasf-sdk/extractor"
+          echo "Provisioning OASF extractor assets to ${EXTRACTOR_ASSET_DIR}"
+          dirctl init --yes \
+            --oasf-url https://schema.oasf.outshift.com \
+            --asset-dir "${EXTRACTOR_ASSET_DIR}"
+
```
tkircsi: CAn we make an GH action for this and re-use?

csirmazbendeguz:
Thanks @akijakya ! I think, let's just not worry about caching then. We don't run this job often anyway.

.github/workflows/import-records.yaml
csirmazbendeguz: good idea

tkircsi (approved):

akijakya (approved):

PR #2016, opened 2026-08-12
fix(cli): keep dirctl logs off stdout

### LINKED ISSUE #2009: [Bug]: dirctl writes structured logs to stdout, corrupting machine-readable command output
### Bug Description

`dirctl` writes its structured `slog` output to **stdout**, the same file descriptor it uses for command results. Any log record emitted while a command runs is interleaved with the data that command prints, so `--output json`, `--output jsonl`, and raw/binary payloads can all be silently corrupted.

Every documented pipeline in the CLI reference is exposed — `dirctl search ... | jq`, `dirctl pull > record.json`, `dirctl search --format cid --output json | dirctl routing publish --stdin`.

This is not limited to verbose configurations. The default log level is `INFO`, and under x509/SPIFFE authentication the client logs at `INFO` on **every TLS handshake**, so a stock `dirctl` pollutes stdout on every command in any SPIFFE-enabled environment.

## Root cause

Three defects compound. Each is independently worth fixing.

**1. The log sink is stdout, and there is no way to select stderr.**

`getLogOutput()` in `utils/logging/logging.go`:

```go
func getLogOutput(logFilePath string) *os.File {
	if logFilePath != "" {
		// ... open the file ...
	}

	return os.Stdout
}
```

The only alternative is `DIRECTORY_LOGGER_LOG_FILE`. There is no `os.Stderr` reference anywhere under `utils/logging/`, so a user cannot keep stdout clean without diverting logs to a file. Meanwhile `cli/presenter` writes command data to `cmd.OutOrStdout()`, which is `os.Stdout` in any real invocation. Both land on fd 1.

**2. The logger is installed by a package `init()` and cannot be re-configured.**

```go
var once sync.Once

func InitLogger(cfg *Config) {
	once.Do(func() { /* ... */ })
}

func init() {
	cfg, err := LoadConfig()
	if err != nil {
		slog.Error("Failed to load config", "error", err)
		os.Exit(1)
	}

	InitLogger(cfg)
}
```

`dirctl` picks this up transitively (`cli/cmd/root.go` → `client` → `utils/logging`), so the sink is bound to stdout before `main()` runs. Because `init()` consumes the `sync.Once`, a later `InitLogger` call from the CLI entrypoint is a no-op — the CLI **cannot** currently redirect its own logs. This is the constraint that makes the fix a small restructure rather than a one-line change.

**3. The presenter already implements the correct policy, and `slog` bypasses it.**

`cli/presenter/presenter.go` routes metadata to stderr whenever the output format is structured:

```go
// PrintSmartf writes formatted output to stdout for human format,
// or to stderr for structured formats (json, jsonl, raw).
// Use this for metadata messages that should not pollute structured output.
func PrintSmartf(cmd *cobra.Command, format string, args ...any) {
	opts := GetOutputOptions(cmd)
	if opts.IsStructuredOutput() {
		Errorf(cmd, format, args...)
	} else {
		Printf(cmd, format, args...)
	}
}
```

The intent is explicit and documented in `docs/content/dir/dir-cli-reference.md` ("Structured formats send data to **stdout** and messages to **stderr**, so piping to `jq` works cleanly"). But `slog` writes to `os.Stdout` directly, never touching Cobra, so it punches straight through the policy.

The codebase has already worked around this three times rather than fixing the sink, which suggests it is recognised but unaddressed:

- `cli/presenter/output.go` — `WriteMessageToFile` exists "to avoid stdout pollution from logs"
- `cli/cmd/verify/verify.go` — `--output-file` takes priority "to avoid stdout pollution"
- `samples/README.md` — the documented way to run the sample is `DIRECTORY_LOGGER_LOG_LEVEL=error go run main.go`

## Blast radius

In rough order of severity:

1. **`dirctl export` without `--output-file`** writes the formatter payload to stdout (`presenter.Print(cmd, string(output))` in `cli/cmd/export/export.go`). For `--format=agent-skill-bundle` that payload is a **gzip tar archive** — interleaved log lines corrupt binary content, and the failure surfaces as a broken archive rather than a legible error.
2. **`dirctl events listen -o jsonl`** streams NDJSON in a loop. One log line mid-stream breaks any line-oriented consumer, and the stream is long-lived so the odds accumulate.
3. **`dirctl pull`** defaults to JSON on stdout — `cli/cmd/pull/options.go` sets `AddOutputFlagsWithDefault(Command, string(presenter.FormatJSON))` with the comment "Records are primarily machine-consumed, so default to JSON on stdout".
4. **Every command with `--output json|jsonl|raw`**: `search`, `push`, `info`, `verify`, `sign`, `delete`, `doctor`, `sync create|list|status|delete`, `routing publish|unpublish|list|search|info`, `naming verify`.
5. **`dirctl context current --json`**, which encodes directly to `cmd.OutOrStdout()`.

Log records that actually fire during a normal run:

- `utils/spiffe/retry.go` logs at `INFO` on entry to and success of `GetX509SVID()`, which the gRPC credentials call on **every TLS handshake**. This is the default-configuration case.
- `client/options.go` logs at `INFO` on successful SVID acquisition.
- `client/routing.go` and similar stream paths log at `ERROR`, which always passes the default filter.
- At `DEBUG`, `client/options.go` adds a block of auth setup lines during `client.New()` in `PersistentPreRunE`, before any command body runs.

## Why CI does not catch this

The e2e harness captures Cobra's writer, not the process's stdout — `tests/e2e/shared/utils/cli.go`:

```go
cmd := clicmd.RootCmd
cmd.SetOut(&outputBuffer)
```

`slog` holds a direct reference to `os.Stdout`, so its output never reaches `outputBuffer`. Every pipe-cleanliness assertion in the suite inspects a buffer the logs bypass, including the JSON parse checks in `tests/e2e/local/11_pull_test.go`. `cli/presenter/presenter_test.go` has the same blind spot: it verifies `PrintSmartf` keeps stdout clean, but only for presenter output.

`utils/logging/logging_test.go` currently *asserts* the stdout behaviour (`TestGetLogOutputStdout`), so it needs updating alongside any fix.

## Proposed fix

1. **Default the CLI's log sink to stderr.** Logs are diagnostics; stdout belongs to the command's result. This is standard CLI behaviour and makes every documented pipeline correct by default.
2. **Make the sink selectable rather than hardcoded.** Add an output-stream option to `logging.Config` (e.g. `DIRECTORY_LOGGER_LOG_STREAM=stdout|stderr`, defaulting per binary) and have `getLogOutput` honour it, keeping `LogFile` as the highest-priority override.
3. **Drop the auto-`init()` in favour of explicit initialization per binary**, or add a `Reconfigure`-style entrypoint that bypasses the `sync.Once`. Without this the CLI cannot set its own sink no matter what defaults are chosen. Explicit initialization in each `main()` is cleaner and makes the CLI/server split obvious at the call site.
4. **Keep server behaviour unchanged.** Logging to stdout is correct 12-factor behaviour for `apiserver` (`server/cmd/main.go`), the standalone `reconciler` (`reconciler/main.go`), and `dirctl daemon start` (`cli/cmd/daemon/start.go`), all of which share this package. The change must be scoped to the CLI process rather than applied to `utils/logging` globally, or it will break existing log collection.
5. **Consider retiring the point workarounds** once the sink is fixed. `WriteMessageToFile` and the `verify --output-file` precedence are useful features in their own right, but their stated justification disappears.

### Suggested regression tests

- Run an `--output json` command with the real `os.Stdout` swapped for a pipe (not `cmd.SetOut`) at log level `DEBUG`, and assert the captured stdout parses as a single JSON value. This is the assertion the current e2e harness structurally cannot make.
- The same check for `events listen -o jsonl`: every line on stdout parses as JSON.
- A byte-for-byte comparison for `export --format=agent-skill-bundle` with and without logging enabled, asserting the gzip stream is identical and decompresses.
- Update `TestGetLogOutputStdout` for the new default, and add coverage for the stream selection option.
- Assert the server/daemon path still logs to stdout, so the CLI fix cannot regress it.

## Related: chart-side hardening

The generated policy script in `install/charts/dirctl/templates/_policies.tpl` takes `count="$(printf '%s' "$out" | jq 'length')"` without checking that `$out` is a single JSON array. When it is a concatenated stream, `jq` emits one length per value, `[ "$count" -eq 0 ]` fails as a non-numeric comparison, and because that is an `if` condition `set -e` does not stop the script — it falls through and pipes the garbage into a mutating command. The template comment already anticipates exactly this hazard for empty output. A `jq -e 'type == "array"'` guard before counting would turn silent corruption into a clear failure.

### Expected Behavior

Diagnostic logs should never share a stream with machine-readable command output.

Concretely:

- `dirctl` should write logs to stderr by default, at every log level, so stdout carries only the command's result.
- `dirctl <cmd> --output json 2>/dev/null | jq .` should succeed regardless of `DIRECTORY_LOGGER_LOG_LEVEL`.
- Binary payloads (`export --format=agent-skill-bundle`) written to stdout should be byte-identical whether or not logging is enabled.
- Server components (`apiserver`, `reconciler`, `dirctl daemon start`) should keep logging to stdout, unchanged.

### Affected Version

v1.6.3 (e75565a4), and `main`.

### Steps to Reproduce

**Minimal reproduction** — no server and no SPIFFE required. Discard stderr and observe that stdout, which should contain only the command's JSON result, carries a log record instead:

```bash
DIRECTORY_LOGGER_LOG_LEVEL=DEBUG \
DIRECTORY_LOGGER_LOG_FORMAT=json \
DIRECTORY_CLIENT_AUTH_MODE=insecure \
DIRECTORY_CLIENT_SERVER_ADDRESS=127.0.0.1:1 \
dirctl search --format cid --output json 2>/dev/null
```

Observed on stdout:

```json
{"time":"2026-08-12T11:54:57.351186+02:00","level":"DEBUG","msg":"Using insecure connection (explicit mode)","component":"client.auth","auth_mode":"insecure"}
```

The connection error correctly goes to stderr; the log record incorrectly goes to stdout. Against a reachable server the same log line is prepended to the real JSON array, so `| jq .` receives two concatenated JSON values instead of one.

**Real-world instance.** The `dir-admin` nightly `publish-all` cronjob fails on every run in two environments. Those set `DIRECTORY_LOGGER_LOG_LEVEL=DEBUG`, and the generated policy script feeds `dirctl search` output straight into `dirctl routing publish --stdin`:

```
sh: 5
4
4
7
...
[publish-all] matched 5
4
...
86 record(s) at offset 0
Error: failed to lookup: failed to receive: rpc error: code = NotFound
desc = failed to lookup record: record not found: {"time":"2026-08-12T07:57:39.317630321Z",
"level":"DEBUG","msg":"Setting up X509 authentication","component":"client.auth",
"spiffe_socket_path":"unix:/run/spire/agent-sockets/api.sock"}
```

`jq 'length'` returned one length per JSON value in the stream rather than a single count, and the log lines were then parsed as CIDs. The apiserver confirms the shape:

```
"msg":"Lookup request received","component":"controller/store",
"cid":"{\"time\":\"2026-08-12T07:57:39.317630321Z\",\"level\":\"DEBUG\",
\"msg\":\"Setting up X509 authentication\",...}"
```

Re-running the identical job with only `DIRECTORY_LOGGER_LOG_LEVEL=ERROR` changed succeeds (`[publish-all] processed 86 record(s)`), isolating the cause to log output on stdout rather than anything in the publish path.

### Workaround

Divert logs off stdout, either by raising the threshold above anything the happy path emits:

```bash
DIRECTORY_LOGGER_LOG_LEVEL=error dirctl search --format cid --output json | jq .
```

or by sending them to a file, which works at any level:

```bash
DIRECTORY_LOGGER_LOG_FILE=/tmp/dirctl.log dirctl pull <cid> > record.json
```

Note that `INFO` is **not** sufficient under x509 authentication — the SVID messages are INFO-level and still land on stdout.

dlanov:
Fixes #2009.

## Summary

Keep `dirctl` diagnostic logs off stdout so machine-readable command output can be safely piped or redirected.

- Default normal `dirctl` logging to stderr.
- Preserve stdout logging for `dirctl daemon start`, apiserver, and reconciler.
- Add `DIRECTORY_LOGGER_LOG_STREAM=stdout|stderr` for explicit stream selection.
- Keep `DIRECTORY_LOGGER_LOG_FILE` as the highest-priority logging destination.
- Preserve already-created component loggers by routing the binary default through a switchable writer.

## Behavior

With no explicit log stream:

- normal `dirctl` commands log to stderr
- `dirctl daemon start` continues logging to stdout
- apiserver and reconciler remain unchanged on stdout

Explicit `DIRECTORY_LOGGER_LOG_STREAM` overrides the binary default, while a valid `DIRECTORY_LOGGER_LOG_FILE` overrides both.

This keeps JSON, JSONL, raw, and binary command results clean on stdout.

## Testing

- `go test -race -count=10 ./logging/...`
- `go test -race -count=1 ./cmd/... ./presenter/...`
- `go build ./...` for cli, server, and reconciler
- `go vet ./...` for cli, server, reconciler, and utils/logging
- `git diff --check`

Manual verification confirmed:
- default `dirctl`: logs on stderr, stdout clean
- `LOG_STREAM=stdout`: logs on stdout
- `LOG_STREAM=stderr`: logs on stderr
- `LOG_FILE` takes precedence over `LOG_STREAM`
- `dirctl daemon start` retains stdout logging


tkircsi (approved):
LGTM! Thank you for your contribution!

PR #2041, opened 2026-08-18
fix(scanner): make MCP source and endpoint scans produce results

### LINKED ISSUE #2040: [Bug]: MCP endpoint scan always fails: missing --analyzers and cmd.Env in runMCPScannerEndpoint
## Summary

The MCP **endpoint** scan phase (`remote`, `prompts`, `resources`, `instructions`) fails for every record on every deployment. `runMCPScannerEndpoint` in `utils/scanner/mcp_endpoint.go` omits both the `--analyzers` flag and `cmd.Env`, so `mcp-scanner` falls back to its default analyzer set (`api,yara,llm`) and never receives the `MCP_SCANNER_LLM_*` credentials that the source-scan path derives.

Because the endpoint phase is skip-with-warning by design, this fails silently: records get a scan result that looks complete but contains no endpoint coverage at all.

Observed on the dev cluster (reconciler `v1.6.3`):

```
time=2026-08-18T11:52:26.974Z level=WARN msg="mcp-scanner endpoint subcommand failed, skipping"
  component=utils/scanner/mcp subcommand=remote url=https://api.thousandeyes.com/mcp-public
  error="mcp-scanner remote exited with error: Error during scanning: Cannot proceed with scan - missing required configuration:
   • API analyzer requested but MCP_SCANNER_API_KEY not configured
   • LLM analyzer requested but MCP_SCANNER_LLM_API_KEY not configured (or AWS credentials for Bedrock models): exit status 1"
```

## Root cause

`utils/scanner/mcp_endpoint.go`:

```go
func runMCPScannerEndpoint(ctx context.Context, cliPath, subcommand, serverURL string) ([]byte, error) {
	var stdout, stderr bytes.Buffer

	cmd := exec.CommandContext(ctx, cliPath, "--raw", subcommand, "--server-url", serverURL)
	cmd.Stdout = &stdout
	cmd.Stderr = &stderr

	if err := cmd.Run(); err != nil {
```

Two defects, both absent from the equivalent source-scan path (`runMCPScanner` in `utils/scanner/mcp.go`):

**1. No `--analyzers` flag.** Per the CLI, the default is `api,yara,llm`:

```
--analyzers ANALYZERS
      Comma-separated list of analyzers to run. Options:
      api, yara, llm, behavioral, virustotal, readiness,
      vulnerable_package, meta (default: api,yara,llm)
```

That default requests the Cisco AI Defense `api` analyzer — a credential the Directory does not provision anywhere — plus the `llm` analyzer. Those are exactly the two analyzers named in the error. `runMCPScanner` avoids this by pinning `--analyzers yara,readiness`.

**2. `cmd.Env` is never set.** `buildMCPScannerEnv()`, which maps `AZURE_OPENAI_*` onto the `MCP_SCANNER_LLM_*` names that `mcp-scanner` actually reads, is only called by `runMCPScanner`. Leaving `cmd.Env` nil inherits the parent environment, so `AZURE_OPENAI_API_KEY` reaches the subprocess but under a name the scanner ignores. Even with the `api` analyzer dropped, the `llm` analyzer would still fail.

Deployment configuration is **not** at fault. The Vault sync (`externalSecrets.azureOpenAI` → `azure-openai-api-key`) and the chart wiring in `install/charts/dir/apiserver/templates/reconciler-deployment.yaml` correctly inject `AZURE_OPENAI_API_KEY` into the reconciler container.

## Proposed fix

In `runMCPScannerEndpoint`, pass an explicit analyzer list and set `cmd.Env = buildMCPScannerEnv()`. The analyzer set is a policy decision worth settling in review:

- `yara` only — zero credentials, always works, no LLM analysis.
- `yara,llm` — uses the Azure key already provisioned; drops the unprovisionable `api` analyzer.
- `yara` always, `llm` appended only when `AZURE_OPENAI_API_KEY` (or `MCP_SCANNER_LLM_API_KEY`) is set — degrades gracefully on deployments without LLM credentials, which is the current behaviour for `reconciler.azureOpenAI.enabled: false`.

The third option seems preferable, since the chart defaults `azureOpenAI.enabled` to `false` and a hard `llm` requirement would reintroduce this same silent failure for those users.

## Related cleanup

`runEndpointSubcommand` sets `result.Analyzers = []string{subcommand}`, recording the subcommand name (`remote`, `prompts`, …) where the source path records real analyzer names (`yara`, `readiness`). Scan results therefore cannot show which analyzers ran against an endpoint. Worth aligning as part of this change.

## Reproduction

Any record declaring an `sse` or `streamable-http` connection URL, with `reconciler.config.scan.enabled: true` and no `MCP_SCANNER_API_KEY` in the environment.

paralta:
MCP scanning returned almost nothing: endpoint scans aborted before dialing, source scans emitted output the parser could not read, and the instructions sub-scan was discarded every run. A local daemon run showed all three (`missing required configuration`, `invalid character '='`, `produced unparsable output`). A final local run reported none of them — only genuine environmental failures: unreachable published endpoints, one with an unexpanded `{tenant_id}` placeholder, and repositories that do not resolve.

- Endpoint scans now request analyzers explicitly. A requested analyzer with missing credentials raises rather than warns on that path, so the `api,yara,llm` default aborted every scan without a Cisco AI Defense key. yara and readiness always run; llm and api are added once their key is configured, keeping them opt-in rather than excluded.
- The endpoint subprocess now gets the mapped environment: credentials arrive as `AZURE_OPENAI_*` but the scanner reads `MCP_SCANNER_LLM_*`. The source path already did this.
- `--raw` moved after the subcommand for source scans only. The behavioral subparser's own `--raw` default overwrote the global one, so the CLI printed a human-readable summary. The live-server subparsers reject a trailing `--raw`, so the paths deliberately differ; both were verified against the real binary.
- Findings decode from either shape the scanner emits — an object keyed by analyzer, or the array instructions returns. An unrecognized shape now errors instead of being silently dropped.

Not addressed: GitHub `/tree/<ref>/<path>` locators still fail to clone, which is how monorepo-hosted MCP servers are usually published. Separate change, and it would give a live path to exercise the source fix.

### DIFF
--- a/utils/scanner/mcp.go
+++ b/utils/scanner/mcp.go
@@ -7,6 +7,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -238,13 +239,25 @@ func gitClone(ctx context.Context, repoURL, dest string) error {
 }
 
 func runMCPScanner(ctx context.Context, cliPath, scanDir string) ([]byte, error) {
-	var stdout, stderr bytes.Buffer
+	// Results go to a file because stdout carries the scanner's own logs and
+	// LiteLLM's banners. --output is redeclared by the behavioral subparser,
+	// whose default clobbers a global one, so it must follow the subcommand.
+	outDir, err := os.MkdirTemp("", "mcp-scan-out-*")
+	if err != nil {
+		return nil, fmt.Errorf("create mcp-scanner output dir: %w", err)
+	}
 
-	// The behavioral subparser declares its own --raw, whose default
-	// overwrites a global one, so --raw must follow the subcommand to get
-	// JSON rather than a human-readable summary.
-	cmd := exec.CommandContext(ctx, cliPath, "--analyzers", "yara,readiness", "behavioral", scanDir, "--raw")
-	cmd.Stdout = &stdout
+	defer os.RemoveAll(outDir)
+
+	outPath := filepath.Join(outDir, "scan-result.json")
+
+	var stderr bytes.Buffer
+
+	cmd := exec.CommandContext(ctx, cliPath,
+		"--analyzers", "yara,readiness",
+		"behavioral", scanDir,
+		"--output", outPath,
+	)
 	cmd.Stderr = &stderr
 	cmd.Env = buildMCPScannerEnv()
 
@@ -254,7 +267,12 @@ func runMCPScanner(ctx context.Context, cliPath, scanDir string) ([]byte, error)
 		return nil, fmt.Errorf("mcp-scanner exited with error: %w", err)
 	}
 
-	return stdout.Bytes(), nil
+	out, err := os.ReadFile(outPath)
+	if err != nil {
+		return nil, fmt.Errorf("read mcp-scanner output: %w", err)
+	}
+
+	return out, nil
 }
 
 // buildMCPScannerEnv returns the parent env with MCP_SCANNER_LLM_* vars derived from
@@ -363,10 +381,8 @@ func (r mcpScannerResult) subject() string {
 }
 
 func parseMCPOutput(raw []byte) (*ScanResult, error) {
-	raw = trimToJSON(raw)
-
-	var results []mcpScannerResult
-	if err := json.Unmarshal(raw, &results); err != nil {
+	results, err := decodeMCPResults(raw)
+	if err != nil {
 		return nil, fmt.Errorf("parse mcp-scanner output: %w", err)
 	}
 
@@ -400,12 +416,45 @@ func parseMCPOutput(raw []byte) (*ScanResult, error) {
 	return &ScanResult{Safe: len(findings) == 0, Findings: findings}, nil
 }
 
-// trimToJSON strips any leading non-JSON content by finding the first '['.
-func trimToJSON(raw []byte) []byte {
-	idx := bytes.IndexByte(raw, '[')
-	if idx > 0 {
-		return raw[idx:]
+// decodeMCPResults extracts the results array from output that may carry log
+// lines either side of it: the live-server subcommands ignore --output, and an
+// ANSI colour escape (\x1b[1;31m) contains a '[' of its own. A non-empty array
+// wins outright so a bracket in a log line cannot pass an empty result off as
+// a clean scan.
+func decodeMCPResults(raw []byte) ([]mcpScannerResult, error) {
+	var (
+		empty      []mcpScannerResult
+		foundEmpty bool
+	)
+
+	for off := 0; off < len(raw); {
+		idx := bytes.IndexByte(raw[off:], '[')
+		if idx < 0 {
+			break
+		}
+
+		start := off + idx
+		off = start + 1
+
+		var results []mcpScannerResult
+
+		// Decoder, not Unmarshal: log lines may trail the array too.
+		if err := json.NewDecoder(bytes.NewReader(raw[start:])).Decode(&results); err != nil {
+			continue
+		}
+
+		if len(results) > 0 {
+			return results, nil
+		}
+
+		if !foundEmpty {
+			empty, foundEmpty = results, true
+		}
 	}
 
-	return raw
+	if foundEmpty {
+		return empty, nil
+	}
+
+	return nil, errors.New("no JSON results array found in output")
 }
--- a/utils/scanner/mcp.go
+++ b/utils/scanner/mcp.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	typesv1 "buf.build/gen/go/agntcy/oasf/protocolbuffers/go/agntcy/oasf/types/v1"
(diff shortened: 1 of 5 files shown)


tkircsi (commented):
Reviewed against #2040. I built the branch, ran the tests, and probed the real `mcp-scanner` binary plus its installed Python source to check the CLI claims.

**Both defects from #2040 are correctly fixed. However the third claim — that source scans now parse — does not hold: I reproduced the `invalid character` parse failure on the new code path.** There is also a hard failure mode that the PR's credential-aware design does not extend to the source scan.

## What checks out

`cmd.Env = buildMCPScannerEnv()` on the endpoint path and the credential-aware `endpointAnalyzers()` both address #2040 directly. The `yara,readiness` baseline is the right choice: in `mcpscanner/core/scanner.py`, `ReadinessAnalyzer()` is constructed unconditionally ("no API keys needed") and `_validate_analyzer_requirements` only rejects `readiness`/`yara` if initialization outright failed, so that pair genuinely cannot fail on credentials.

The flag-placement asymmetry is justified, which I did not expect until I checked. The `behavioral` subparser really does redeclare `--raw` (and `--format`, defaulting to `summary`), so its default clobbers a global `--raw`. The `remote` and `instructions` subparsers declare only `--server-url`, `--bearer-token`, and `--header`, so globals before the subcommand survive there. Both halves of the claim hold.

Severity handling on the new array shape is fine — `mapScannerSeverity` normalizes via `strings.ToUpper`, so uppercase `HIGH`/`YARA` map correctly. Branch builds clean, `go test ./scanner/...` passes.

## Blocking

### 1. The `invalid character '='` failure is not fixed

The clobbered `--raw` is only one of two triggers. LiteLLM writes ANSI-colored banners to **stdout**, and `trimToJSON` takes the *first* `[` — which is the `[` inside the escape sequence `\x1b[1;31m`, at byte offset 2. The real JSON sits at the end of the stream.

Captured actual stdout from a run whose LLM call failed, then fed it to this branch's `parseMCPOutput`:

```
err = parse mcp-scanner output: invalid character ';' after array element
```

Same family as the reported `'='`. The process exited **0**, so this bypasses the exec-error branch and lands in `runSourceScan`'s `parseMCPOutput` error path — a hard failure that also discards the endpoint findings.

This will resurface as intermittent scan failures on any transient LLM condition (rate limits, timeouts, connection blips). The clean local run just did not hit a retry. Scanning from the last `[`, stripping ANSI, or using `--output <file>` so JSON never shares a stream with logs would all close it; `--output` is the only one robust by construction.

### 2. The source scan still hard-fails without LLM credentials

The `behavioral` subcommand requires LLM credentials unconditionally — `--analyzers` does not gate it:

```
$ mcp-scanner --analyzers yara behavioral /tmp/mcpsrc --raw
Error during scanning: LLM provider API key is required for alignment verification
```

Identical with `--analyzers readiness`. So the endpoint phase now degrades gracefully while the source phase cannot, and because `runSourceScan` returns an error rather than a skip, the whole record scan fails. Since `reconciler.azureOpenAI.enabled` defaults to `false`, every MCP record with a source locator fails to scan on a default deployment — so "source scans produce results" is not yet true there. Either the source phase needs the same credential-aware treatment (e.g. `static` when no LLM key is present), or a missing credential should be a skip rather than a hard error.

## Worth fixing while here

**`buildMCPScannerEnv` exports a malformed model.** `"azure/"+os.Getenv("AZURE_OPENAI_DEPLOYMENT")` is never empty, so `appendEnvIfMissing`'s `fallback == ""` guard never fires and `MCP_SCANNER_LLM_MODEL=azure/` is exported when the deployment name is unset. Pre-existing, but this PR routes four extra invocations per endpoint through it, and it feeds directly into finding 1.

**`endpointAnalyzers()` gates on the key alone.** It adds `llm` when `AZURE_OPENAI_API_KEY` is set, but the call also needs base URL and deployment. Key-only config passes the scanner's validation and then fails at request time — the same path into finding 1. Gating on the full triple would make the degradation honest.

**Entry-level shapes still degrade silently.** The container shape now errors, but entries do not: an object whose values lack `severity`/`threat_summary` unmarshals into zero values, yielding a finding with `SeverityInfo` and a message like `[yara] toolname: `. `DisallowUnknownFields`, or rejecting entries with empty severity and summary, would match the stated intent.

**Finding order is non-deterministic** in the object branch (ranges over a map). Pre-existing, but it now sits beside a deterministic array branch, and these results are stored — sorting by analyzer name would keep output stable.

**The `Analyzers` cleanup from #2040 is still open.** `runEndpointSubcommand` reports the subcommand name (`remote`, `prompts`) where `endpointAnalyzers()` now knows the real list. Relatedly `runSourceScan` hardcodes `[]string{"yara", "readiness"}` even though the behavioral alignment pass (LLM) also runs, so it understates.

**No test pins flag ordering.** The `--raw` placement is the most fragile part of this change — it depends on argparse subparser-default behavior that is not evident from the code — yet the `fakecli` tests do not assert argv. A test asserting argument order for both paths would stop a future cleanup from silently reintroducing the summary-output bug.

tkircsi (approved):
LGTM! Thanks!

PR #2067, opened 2026-08-25
fix(dir/server): handle missing manifest gracefully to prevent orphaned search index rows

### LINKED ISSUE #2066: [Bug]: Deleting a record can orphan its search-index row, permanently and unrecoverably
### Problem Statement

`StoreService.Delete` removes a record from the OCI store and from the search index as two
separate, non-atomic steps. If the OCI step reports an error, the method returns before the
index cleanup runs, leaving a search-index row with no backing record — an *orphan*.

The orphan is permanent. Every subsequent delete for that CID fails at the same point, because
the manifest is now genuinely absent, so the index row can never be cleaned up through the API
that is supposed to clean it up. `dirctl delete` reports `not found` forever while the record
keeps appearing in `dirctl search`.

### Proposed Solution

Keep the current ordering — OCI first, then the database — but make "record not found in OCI"
a **non-fatal** outcome that allows the delete to continue, so the DB row is still removed.

**Change 1 — recognise a missing manifest by error value.** In `server/store/oci/internal.go`:

```go
import (
    "errors"
    "oras.land/oras-go/v2/errdef"
)

if err != nil {
    // Manifest already absent: deletion is idempotent, treat as success so the
    // caller still reconciles the search index.
    if errors.Is(err, errdef.ErrNotFound) {
        internalLogger.Info("Manifest not found (never existed or deleted already)", "cid", cid)

        return nil
    }

    return status.Errorf(codes.Internal, "failed to resolve manifest for deletion: %v", err)
}
```

`errors.Is` against `errdef.ErrNotFound` is the correct check for `oras-go/v2` (the server
already depends on `oras.land/oras-go/v2 v2.6.2`). The existing `404` / `NOT_FOUND` string
checks can be retained as a fallback for registries that surface a bare HTTP error, but they
must not be the primary mechanism.

**Change 2 — defend the invariant in the controller.** Even with change 1, the controller should
not let a store-side `NotFound` block index cleanup, whatever layer produces it:

```go
err = s.store.Delete(stream.Context(), recordRef)
if err != nil && status.Code(err) != codes.NotFound {
    st := status.Convert(err)

    return status.Errorf(st.Code(), "failed to delete record: %s", st.Message())
}

// Reached when the record was deleted, or was already absent. Either way the
// search index must no longer list it.
if err := s.db.RemoveRecord(recordRef.GetCid()); err != nil {
    storeLogger.Error("Failed to remove record from search index", "error", err, "cid", recordRef.GetCid())
}
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

### Checklist

vijit-vishnoi:
**Fixes #2066**

### Problem Statement
Previously, `StoreService.Delete` would fail and return an error if a record's manifest was already missing from the OCI store. Because the OCI deletion step failed, the subsequent database search-index cleanup step was never reached, leaving behind an orphaned, unrecoverable row in the search index.

### Proposed Changes
This PR implements the maintainer's proposed solution to make "record not found in OCI" a non-fatal outcome, allowing the deletion process to continue and clean up the database index:

1. **OCI Store (`server/store/oci/internal.go`):** 
   - Updated `deleteFromOCIStore` and `deleteFromRemoteRepository` to check for `errdef.ErrNotFound` using `errors.Is`.
   - If the manifest is missing, it logs an informational message and returns `nil` (success), treating the deletion as idempotent.
2. **Controller (`server/controller/store.go`):** 
   - Updated the error evaluation block immediately following `s.store.Delete`.
   - The deletion process now proceeds to `s.db.RemoveRecord` even if the store returns a `codes.NotFound` gRPC status.
3. **Unit Tests Added:**
   - `TestDeleteFromOCIStore_NotFound`: Verifies that attempting to delete a non-existent CID in the OCI store correctly returns `nil`.
   - `TestStoreDelete_NotFound`: Verifies that the controller's `Delete` method correctly invokes the database `RemoveRecord` mock even when the OCI store returns `codes.NotFound`.

### Checklist
- [x] I have read the contributing guidelines
- [x] I have signed off my commits (DCO)
- [x] I have added tests that prove my fix is effective
- [x] All local tests pass successfully (`go test`)

### DIFF
--- a/server/controller/store.go
+++ b/server/controller/store.go
@@ -201,7 +201,7 @@ func (s storeCtrl) Delete(stream storev1.StoreService_DeleteServer) error {
 
 		// Delete record from store
 		err = s.store.Delete(stream.Context(), recordRef)
-		if err != nil {
+		if err != nil && status.Code(err) != codes.NotFound {
 			st := status.Convert(err)
 
 			return status.Errorf(st.Code(), "failed to delete record: %s", st.Message())
--- a/server/controller/store_test.go
+++ b/server/controller/store_test.go
@@ -1,0 +1,81 @@
+// Copyright AGNTCY Contributors (https://github.com/agntcy)
+// SPDX-License-Identifier: Apache-2.0
+
+package controller
+
+import (
+	"context"
+	"io"
+	"testing"
+
+	corev1 "github.com/agntcy/dir/api/core/v1"
+	"github.com/agntcy/dir/server/types"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/emptypb"
+)
+
+type mockStore struct {
+	types.StoreAPI
+	deleteErr error
+}
+
+func (m *mockStore) Delete(ctx context.Context, ref *corev1.RecordRef) error {
+	return m.deleteErr
+}
+
+type mockDB struct {
+	types.DatabaseAPI
+	removeRecordCalled bool
+}
+
+func (m *mockDB) RemoveRecord(cid string) error {
+	m.removeRecordCalled = true
+
+	return nil
+}
+
+type mockDeleteServer struct {
+	grpc.ServerStream
+	reqs []*corev1.RecordRef
+	idx  int
+}
+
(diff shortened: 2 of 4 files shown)


paralta:
@vijit-vishnoi Please run `FIX=true task lint` to fix the linter failures

paralta (approved):
LGTM 🚀 Thank you so much for picking this issue up and for how quickly you turned it around — neat fix!

PR #2075, opened 2026-08-28
feat(ci): raise import matrix max-parallel to 3 (fixes #2047)

### LINKED ISSUE #2047: [Feature]: import-records: raise matrix max-parallel from 1 to 3 now that the dedup pre-scan is gone
## Summary

Both matrices in `.github/workflows/import-records.yaml` are pinned to `max-parallel: 1`, so every MCP entry and every agent-skill repo is imported strictly one after another. The nightly run's wall clock is the sum of all jobs, and recent runs have taken hours mostly because of queuing rather than per-job work.

The constraint that made serialization prudent is gone. Raising `max-parallel` to 3 is a two-line change with no known correctness risk.

## Why it was reasonable before

The importer used to build a deduplication cache eagerly at startup, pulling the entire existing catalog for the record type before the pipeline began. With one job per vendor prefix or per skill repo, running several in parallel meant several catalog-wide scans hitting prod simultaneously. Serializing was cheap insurance.

That pre-scan was removed in agntcy/dir-importer#92 (lazy, per-item lookups), and the internal/external OCI address split (#2019 / #2021) removed the hairpinning that made those reads expensive. Import steps in the latest run complete in tens of seconds to ~12 minutes each.

## Why parallel is safe now

- Each matrix entry runs on its own runner with its own downloaded `dirctl` and its own generated `import_ci.yaml`.
- Matrix entries cover disjoint inputs — distinct vendor prefixes for MCP, distinct repos for agent skills — so two jobs cannot race on the same record.
- Pushes are content-addressed and idempotent by CID; concurrent pushes of unrelated records do not conflict.
- Artifact names are already unique per entry (`failed-to-sign-mcp-${{ matrix.entry.index }}`, `failed-to-sign-agent-skills-${{ matrix.entry.index }}`).
- Cache keys are already unique per repo (`agent-skill-repo-${{ matrix.entry.repo_slug }}`).
- The workflow-level `concurrency: group: import-records` still prevents two *runs* from overlapping, which is the guard that actually protects prod from compounding load.
- The enrichment path in CI is `extractor`, which runs in-process with no shared external dependency and no rate limiter.

## Proposal

Change `max-parallel: 1` to `max-parallel: 3` in both places:

- `.github/workflows/import-records.yaml:163` (`import-mcp`)
- `.github/workflows/import-records.yaml:358` (`import-agent-skills`)

Optionally expose it as a `workflow_dispatch` input so it can be tuned or dropped back to 1 without a code change.

## Risk and what to watch

The binding constraint is not the Directory or the store — the store ingress allows 50 r/s with a burst of 250, and each job pushes records serially. It is **Sigstore**: parallel jobs multiply concurrent Fulcio and Rekor calls, and Rekor is already the slowest leg of signing (90s timeout, one retry in sigstore-go).

Start at 3 and watch for:

- Rekor `409 an equivalent entry already exists` annotations (these come from sigstore-go retrying the same bundle after a lost or timed-out response, so they are a slowness signal).
- Per-CID signing time creeping up in the sign step logs.
- Any new `Unauthenticated` or stream errors against the gateway.

Rollback is reverting the two values to 1.

## Verification

Run the workflow manually with `sign: false` and `dry_run: false` against a small subset first, confirm no cross-job interference, then compare total run duration and per-job import/sign durations against the previous nightly run.

## Not in scope

Two related tracks are deliberately excluded and will be filed separately:

- Signing throughput: batching CIDs into one `dirctl` process to amortize process start, gRPC dial, auth, and the Fulcio certificate; and parallelizing within a job.
- Why signing volume is high at all: agent-skill records appear to be re-imported and re-signed on every run, likely the moduleless-OASF gap in the dedup module filter discussed in agntcy/dir-importer#92.

vijit-vishnoi:
**Fixes #2047**

### Problem Statement
The importer pipelines were historically bottlenecked at `max-parallel: 1` to protect production from compounding load during eager dedup cache building. Since the importer now uses lazy per-item lookups, this serialization is no longer necessary.

### Proposed Changes
This PR increases `max-parallel` from `1` to `3` for both the `import-mcp` and `import-agent-skills` matrix jobs.

### ⚠️ Verification Required
As noted in the original issue, **I am unable to run the Verification step locally** since this workflow requires access to repository secrets and Sigstore authentication. 

Could a maintainer please trigger this workflow manually via `workflow_dispatch` with `sign: false` and `dry_run: false` to verify no cross-job interference before merging?

### Checklist
- [x] I have read the contributing guidelines
- [x] I have signed off my commits (DCO)

### DIFF
--- a/.github/workflows/import-records.yaml
+++ b/.github/workflows/import-records.yaml
@@ -160,7 +160,7 @@ jobs:
       id-token: write # Required for OIDC signing with Sigstore
     strategy:
       fail-fast: false
-      max-parallel: 1
+      max-parallel: 3
       matrix:
         entry: ${{ fromJson(needs.setup-matrix.outputs.mcp_matrix) }}
     env:
@@ -355,7 +355,7 @@ jobs:
       id-token: write # Required for OIDC signing with Sigstore
     strategy:
       fail-fast: false
-      max-parallel: 1
+      max-parallel: 3
       matrix:
         entry: ${{ fromJson(needs.setup-matrix.outputs.agent_skills_matrix) }}
     env:


vijit-vishnoi:
Hi! It looks like CI pipelines failed due to a race condition. Since this in completely unrelated tom yaml changes could a maintainer hit re-run on all the jobs?

ramizpolic (approved):

lgecse:
@vijit-vishnoi thank you for the contribution!

PR #2095, opened 2026-09-02
feat(scan): persist scan failures and expose them to search

### LINKED ISSUE #2086: [Feature]: Persist and expose security scan failures
### Problem Statement

The security scan reconciler discards every scan it could not complete. When a
runner skips or fails, `reconciler/tasks/scan/task.go` logs and moves on without
writing a `scan_reports` row or an OCI referrer. Four consequences follow:

1. **Unscannable records are unobservable.** `dirctl search` cannot express
   "could not be scanned". Because such a record has no `scan_reports` row at
   all, it falls out of *both* `--safe` (which requires an `EXISTS` row) and
   `--safe=false` (which requires a row with `is_safe = false`). It is
   indistinguishable from a record that has simply not been scanned yet.

2. **Retries are unbounded.** `GetRecordsNeedingScan` selects records with no
   scan row newer than the TTL. A failed scan writes no row, so the same record
   is re-selected and re-attempted on *every* scan interval — one hour in the
   Helm values — indefinitely. Repositories that will never be cloneable are
   cloned forever.

3. **Failure causes cannot be told apart.** `ScanResult.SkippedReason` is a
   free-form `fmt.Sprintf` string. Nothing distinguishes failures caused by
   published record data (unreachable source repository, malformed endpoint URL)
   from failures caused by our own scanning infrastructure (subprocess timeout,
   OOM kill, missing scanner credentials). Any consumer of this state would have
   to string-match log prose.

4. **Partial scans report as complete.** The MCP runner scans in two phases and
   merges them; `merge` clears `Skipped` when *any* phase succeeded. A record
   whose source clone failed but whose endpoints scanned cleanly persists an
   ordinary passing report, and the failed phase disappears. The existing
   `ScanResult.Notices` field documents this gap but is only ever logged.

### Proposed Solution

Make scan outcomes durable and queryable. No automated remediation.

**1. Typed failure reasons** — add a `FailureReason` string type to
`utils/scanner` with stable lowercase-hyphen codes (`source-unreachable`,
`endpoint-url-invalid`, `scanner-timeout`, `scanner-crashed`,
`llm-not-configured`, …) and an `IsPublisherFault()` classifier, so a record is
never penalised for our own infrastructure failing. Set a reason at each
existing skip site.

**2. Persist every attempt** — extend `scan_reports` with `status`
(`completed` / `partial` / `failed`), `failure_reason`, `failure_detail`,
`consecutive_failures`, `next_attempt_at`, and `last_success_at`. `partial` is
the durable form of `Notices`, covering problem 4. Failure rows store
`is_safe = false` so that any query which forgets the status gate fails closed.

**3. Bound the retry** — schedule the next attempt in Go via `next_attempt_at`:
`now + TTL` on success, `now + min(base * 2^(n-1), max)` on failure. Eligibility
becomes a single timestamp comparison, with no date arithmetic in SQL.

**4. Expose it in search** — add `RECORD_QUERY_TYPE_SCAN_STATUS` and
`RECORD_QUERY_TYPE_SCAN_FAILURE_REASON` (requires `task gen`), plumb them
through `RecordFilters` and the GORM filters, and register `--scan-status` /
`--scan-failure-reason` in the `valueFilters()` table so the `--exclude-*` twins
come for free.

**5. Fix `--safe` in the same change** — gate both branches on
`status IN ('completed', 'partial')`. Without this, every unscannable record
starts reporting as **safe** the moment failure rows exist. This needs a
regression test.

**6. Migration** — `004_scan_report_status.go`, backfilling existing rows to
`status = 'completed'` with their current schedule preserved. Custom migrations
run before `AutoMigrate`, so it must add its own columns (see
`002_backfill_artifact_media_type.go`).

### Out of scope

- Automated deletion, pruning, or unpublishing. Deletion is irreversible, does
  not stick under content addressing, does not withdraw from routing, and
  authorization has no per-record ownership model. Once the filters above exist,
  the `dirctl` chart's `policies:` block can already express such a policy with
  no chart change: an admin wanting to drop records whose source repository is
  private would add a `policies:` entry with `action: prune`, `schedule`, and
  `match: {scan-failure-reason: source-unreachable}`, which renders a CronJob
  running `dirctl search ... | dirctl delete --stdin` (start with
  `dryRun: true`).
- Propagating failures between nodes. A clone failure describes this node's
  network position, not the record, and scan referrers sync to peers — so
  `security/v1/scan_report.proto` is deliberately left unchanged.
- Publish-time URL validation. Worthwhile follow-up (running the existing
  `validateEndpointURL` during ingest would reject malformed endpoints at the
  door) but a separate change to the ingest path.

### Checklist

### LINKED ISSUE #2103: [Bug]: Superseded scan reports accumulate on live records — one new referrer per record per scanner per week
**Decision needed before implementation:** whether a scan trail is wanted at all — see "Open question" below.

## Summary

Every time the scan task re-scans a record it pushes a **new** ScanReport referrer and leaves the previous one in place. The reports are not identical between runs — each embeds `ScannedAt: time.Now()` — so each run produces different bytes, a different CID, and a new manifest with a new tag.

This is a source of registry garbage that garbage collection cannot reach. #2101 deals with referrers whose subject has been deleted; these subjects are alive, so no garbage collector will ever touch them, and the growth continues for as long as the record exists.

It is also not purely a storage problem. `dirctl pull --scan-report` returns **every** ScanReport referrer on the record, so accumulated reports are visible to users — `cli/cmd/pull/pull.go:183-205` appends all of them, with no dedup and no ordering.

Related: #2101 (enable Zot's `deleteReferrers` — cannot help here, see below), #2102 (stop CID-tagging referrers — **shares its step-1 prerequisite**), [Discussion #2072](https://github.com/agntcy/dir/discussions/2072).

## Problem

`scanRecord` builds a report and pushes it as a referrer, with no attempt to remove the report it supersedes:

https://github.com/agntcy/dir/blob/ecc1cd1c56204295ac0c11a685ee659baa64b6ea/reconciler/tasks/scan/task.go#L140-L156

```go
report := buildScanReport(r.Name(), result)

// Push as OCI referrer - failure is logged but does not block the gate.
if pushErr := t.pushReferrer(ctx, recordCID, report); pushErr != nil {
	logger.Warn("Failed to push scan referrer", "runner", r.Name(), "cid", recordCID, "error", pushErr)
}

// Upsert DB row - failure is also non-fatal.
row := &gormdb.ScanReport{ ... }
if upsertErr := t.db.UpsertScanReport(row); upsertErr != nil {
```

Note the asymmetry between those two calls. The **database** row is an upsert keyed on `(record_cid, scanner_type)`, so the DB holds exactly one row per record per scanner and stays correct. The **registry** gets an unconditional push. The two diverge immediately and permanently.

If the report content were stable the push would be harmless — same bytes, same CID, same tag, a no-op. It is not stable, because of one field:

https://github.com/agntcy/dir/blob/ecc1cd1c56204295ac0c11a685ee659baa64b6ea/reconciler/tasks/scan/task.go#L194-L202

```go
return &scanv1.ScanReport{
	ScannerType:    toProtoScannerType(runnerName),
	ScannerVersion: result.Version,
	IsSafe:         result.Safe,
	ScannedAt:      time.Now().UTC().Format(time.RFC3339),
	MaxSeverity:    maxSev,
	Findings:       findings,
	Analyzers:      result.Analyzers,
}
```

The referrer CID is derived from the canonical JSON of that payload, so a fresh `ScannedAt` guarantees a fresh CID on every run, even when the scan result is otherwise byte-for-byte identical.

### Why GC cannot help

Zot's `deleteReferrers` policy — the one #2101 turns on — removes a referrer only when its **subject is absent from `index.json`** (Zot `pkg/storage/gc/gc.go:524-537`). A superseded scan report points at a record that is alive and correctly tagged. It is a perfectly valid referrer by every rule the registry knows. Only Directory knows it has been superseded, so only Directory can remove it.

### How fast is this producing?

The rate follows from the config. Selection is driven by the **TTL**, not by the task interval: `GetRecordsNeedingScan` returns records with no `scan_reports` row newer than the TTL (`server/database/gorm/scan.go:58-71`). Prod sets `interval: "30m"` and no TTL, so it inherits `DefaultTTL = 7 * 24h` (`reconciler/tasks/scan/config.go:16`); the 30-minute interval only controls how often the query runs.

So each record is re-scanned weekly and gains **one referrer per matching scanner per week** — roughly 52 per record per scanner per year. Across 2,889 records that is thousands of new referrers per week in steady state.

What the production data shows, from the 2026-09-01 clone:

| referrer type | live | orphaned |
|---|---|---|
| Signature | — | 3,330 |
| PublicKey | — | 3,330 |
| ScanReport | — | 8,246 |
| **all types** | 10,984 across 2,889 records (3.80 per record) | 14,906 |

Signature and PublicKey orphans are **exactly equal** at 3,330, which is the signature of one-per-record objects pushed once and never re-pushed. ScanReport is 2.48× that figure.

Measured against a weekly rate, though, 2.48 is *small* — months of accumulation would have produced far more. The live side says the same: 3.80 referrers per record, of which roughly 1.8 are scan reports once a Signature and a PublicKey are accounted for. So the excess is more likely records matching more than one of the three scanners (MCP, skill, A2A) than accumulated history, and today's backlog is modest.

**The argument for fixing this is the forward rate, not the current mess.** Confirm before sizing any cleanup sweep: pick ten records older than a few TTL periods, list their ScanReport referrers, and group by the `scanner_type` inside each payload. More than one report per scanner type proves accumulation, and the count dates when the scan task went live.

## Open question: is a scan trail intended?

Multiple reports per record is intentional in one sense and apparently accidental in the other, and this issue proposes to change only the second.

**Across scanner types it is by design.** MCP, skill and A2A each produce their own report for the same record, and all three must coexist. The fix below keys on `scanner_type` precisely to preserve that.

**Repeated reports from the same scanner over time show no sign of being intended:**

- The `scan_reports` table keeps exactly one row per `(record_cid, scanner_type)` and overwrites `is_safe` / `max_severity` on each scan, so the system of record already discards the history.
- `applyReferrerDBEffects` states the model outright — the upsert exists "so the SCANNED and SCAN_SEVERITY search filters reflect **the latest result**" (`server/ingest/ingest.go:134-137`).
- Nothing reads `scanned_at` except `dirctl pull`, which prints it. Nothing orders by it or selects the newest.
- The `ScanReport` proto comment describes a report attached to the record CID it describes, and says nothing about versions or history.
- No "history", "latest scan" or "supersede" concept appears anywhere in the codebase.

So what exists today is neither a usable audit trail nor a clean current-state model: unordered, unbounded, unqueryable, and the summary row discards the history regardless.

**The question still deserves an explicit answer before this ships.** Scan results are security attestations, and "safe in March, critical in June" has real audit value. If we want that, it should be designed rather than inherited: bounded retention (keep N, or keep 90 days), ordered reads, a `scan_reports` row per scan rather than per scanner, and some way to query the trail.

Replace-on-push is the right change *if* the answer is "current state only", which is what every consumer assumes today. If the answer is "we want a trail", this stops being a bug and becomes a retention-policy issue, and the fix is a cap rather than a replace. Either way, the present behaviour is not the intent.

### Replace instead of append

Before pushing a new report, remove the record's existing reports **for the same scanner type**:

```go
func (t *Task) pushReferrer(ctx context.Context, recordCID string, report *scanv1.ScanReport) error {
	referrer, err := report.MarshalReferrer()
	if err != nil {
		return fmt.Errorf("marshal scan report: %w", err)
	}

	// A re-scan supersedes the previous report rather than adding to it.
	// Nothing else can remove these: the subject is alive, so GC never sees them.
	if err := t.deleteSupersededReports(ctx, recordCID, report.GetScannerType()); err != nil {
		return fmt.Errorf("delete superseded reports: %w", err)
	}

	if _, err := t.refStore.PushReferrer(ctx, recordCID, referrer); err != nil {
		return fmt.Errorf("push referrer: %w", err)
	}

	return nil
}
```

**The implementation detail that matters:** `DeleteReferrer`'s type filter operates on the *referrer* type, and all three scanners share the single `ScanReportReferrerType`. Filtering on it would delete the MCP and A2A reports when re-pushing the skill one. `deleteSupersededReports` therefore has to `WalkReferrers` with `ScanReportReferrerType`, unmarshal each payload, and delete only those whose `scanner_type` matches.

`DeleteReferrer` already takes an exact `referrerCID` alongside the type filter (`server/store/oci/referrers.go:317-341`), so `deleteSupersededReports` can reuse it per matching CID rather than opening a new deletion path.

**Custom scanners have no distinct type, so `scanner_type` alone is not a safe key.** `ScannerType` is a closed enum — `UNSPECIFIED`, `MCP`, `SKILL`, `A2A`:

https://github.com/agntcy/dir/blob/ecc1cd1c56204295ac0c11a685ee659baa64b6ea/proto/agntcy/dir/security/v1/scan_report.proto#L9-L17

Third parties can push their own reports through `client.PushScanReport`, so a custom scanner arrives as `SCANNER_TYPE_UNSPECIFIED`, and keying the replace on `scanner_type` would have two independent custom scanners delete each other's reports. Either exclude `UNSPECIFIED` from replacement, or widen the key to include `scanner_version` or a scanner-identity annotation. The `scan_reports` upsert already collides the same way today, so whichever key is chosen should be applied to both.

Order the delete before the push, and treat a delete failure as fatal to the push. The alternative — push first, then clean up — leaves the accumulation in place whenever the cleanup fails, which is the behaviour we are trying to remove.

**Prerequisite shared with #2102.** `DeleteReferrer` resolves each referrer by its CID *tag* (`server/store/oci/internal.go:114-120`, `internal.go:183-190`) and returns `nil` when the tag is absent. That works today, because referrers are tagged. If #2102 step 2 (stop tagging) lands before its step 1 (delete by digest), this fix silently becomes a no-op — worse than not shipping it, because it would look like it worked. Land #2102 step 1 first, or fold it into this change.

### Also worth doing: make unchanged scans idempotent

Even with replace-on-push, every scan cycle rewrites `index.json` twice (one delete, one push) per record, whether or not anything changed. Dropping `ScannedAt` from the referrer payload would make an unchanged scan produce an identical CID, so `PushReferrer` becomes a genuine no-op and the delete can be skipped entirely when the new CID matches the existing one.

The cost is losing the scan timestamp from the registry, and there is only one place it can go: the `scan_reports` DB row, which already carries `CreatedAt`/`UpdatedAt`.

**It cannot be moved onto the referrer instead.** Two things rule that out. The `ReferrerCreatedAtAnnotationKey` annotation is written only when the referrer sets `CreatedAt` (`server/store/oci/referrers.go:111-113`), and `ScanReport.MarshalReferrer` sets only `Type` and `Data` (`api/security/v1/scan_report.go:42-45`) — only `client/sign.go` populates `CreatedAt`, for signatures and public keys. So the annotation is absent on scan reports today.

More fundamentally, populating it would defeat the purpose. `created_at` (field 4) and `annotations` (field 3) are both fields of `RecordReferrer`, and `Marshal()` canonicalises the whole message into the blob whose digest becomes the CID (`referrers.go:84-98`). Any per-scan timestamp carried anywhere on the referrer object changes the CID. Stable CIDs and a registry-side scan timestamp are mutually exclusive under the current design.

This is a proto-visible change to `ScanReport`, so it needs its own discussion. Replace-on-push is the change that stops the growth; this one makes the steady state cheap.

## What this does not fix

- **The reports already accumulated.** Once replace-on-push ships, the next scan of each record collapses its history to one report per scanner — but only for records that get re-scanned. Records whose scan TTL has not expired, or that no longer match any scanner, keep their backlog. A one-time sweep may be worth it once the inventory shows the size.
- **Signature and PublicKey referrers.** These appear to be pushed once per record and show no sign of accumulating. Re-signing a record legitimately adds a signer, so replace-on-push would be wrong there.
- **Records every runner skips.** They never get a `scan_reports` row, so `GetRecordsNeedingScan` re-selects them on every cycle — every 30 minutes in prod — forever. That produces no referrers and no garbage, but it is permanent repeated work, and the candidate set never shrinks. Worth a cheap negative-result row if the scan task ever becomes expensive per candidate rather than per report.

## Verification

1. **Prove the accumulation first** (see above): group a sample record's ScanReport referrers by `scanner_type` and count.
2. On a local daemon, push a record that matches a scanner. Force two scan cycles.
   - Before the fix: 2 ScanReport referrers with the same `scanner_type` and different `ScannedAt`.
   - After: 1.
3. Push a record matching two scanners; force a cycle; confirm both reports exist and neither deleted the other.
   - Then push two client-side reports with `SCANNER_TYPE_UNSPECIFIED` and confirm the chosen key does not have them delete each other.
4. `dirctl pull --scan-report` returns **exactly one report per scanner type** — it returns all of them today, so this is the user-visible assertion — and the `SCANNED` / `SCAN_SEVERITY` search filters still behave.
5. In production, track total referrer count per record weekly. It should become flat rather than rising.

## Rollback

Revert the commit. Accumulation resumes; nothing that was deleted is needed, since the DB row and the current report both survive.

paralta:
The reconciler only recorded a scan when it reached a verdict, so a record that could not be scanned — a private source repository, an unreachable endpoint, a missing scanner binary — left no trace at all. It matched the "needs scanning" query on every pass and was re-cloned indefinitely, and an operator had no way to list those records or tell a fault of the published record apart from a fault of our own infrastructure. Every attempt is now persisted with a status and a machine-readable failure reason, retries are bounded by exponential backoff, and `--scan-status` / `--scan-failure-reason` make the failures searchable so an admin can act on them — for example, a scheduled job that deletes records whose source repository is permanently unreachable.

**Failure Examples from imported records**
| Cause | Status | Reason | Records |
|---|---|---|---|
| `github.com/sravanism/esrp-oss-mcp-server` — private or deleted | `failed` | `source-unreachable` | 3 |
| `github.com/bap-microsoft/MCP-Platform/` — private or deleted | `failed` | `source-unreachable` | 2 |
| `example.com/mcp-server.git` — placeholder never substituted | `failed` | `source-unreachable` | 2 |
| `.../servers/tree/main/src/git` — subdirectory URL, not a repo | `failed` | `source-unreachable` | 1 |
| `mcp.circleci.com`, `mcp.newrelic.com` — gated behind OAuth | `failed` | `endpoint-unreachable` | 2 |

> Note: the fourth example can be handled by changing the `git clone`  code


tkircsi (approved):
LGTM.

Issue #2093, opened 2026-09-02
[Bug]: Extractor enricher sometimes assigns far too many OASF skills on import


paralta:
### Bug Description

Most records imported with the OASF extractor enricher look fine (typically a handful of skills). A smaller set of records published to prod (`ads.outshift.io` / [AI Catalog](https://ai-catalog.outshift.io)) end up with a very large skill list.

This is not a catalog projection bug. Catalog tags match the record’s OASF `skills` 1:1 (`oasf:1.1.0:skills:<name>`). The extra classes are already on the record from extractor enrichment at import.

Examples from prod (checked 2026-09-01). Typical recent imports have **2–6 skills**; these are the outliers:

| Record | Skills |
|---|---|
| [`prospect`](https://ai-catalog.outshift.io/v1/agents/baeareicog4yhsvhua4iilvirccemo3n4vtf742thk5z47k44i65rhouzay) | 404 (~80% of the OASF 1.1.0 skill taxonomy, ~496 classes) |
| `developing-genkit-js` | 387 |
| `hf-cli` | 366 |
| `stripe-best-practices` | 283 |

The import extractor adapter (`cli/cmd/import/extractor_adapter.go`) does not set `ExtractOptions.Tiers`, so it uses `DefaultTiers = 2` (two score groups; see `utils/extractor/extractor.go`). That default is shared with natural-language search and is tuned for recall. There is also no cap on how many skills/domains get written onto the record. This has been the import-records default since [#2015](https://github.com/agntcy/dir/pull/2015) (v1.7.0); nothing obvious changed recently on the extractor itself.

When it does happen, the catalog detail view, `GET /v1/agents/{cid}`, tag filters, and skill-based search are much less useful for that record. Already-imported copies stay as they are until they are re-enriched or replaced.

### Expected Behavior

Extractor enrichment should attach a small, relevant set of OASF skills and domains — similar to the 2–6 classes most imports already get — even for longer or more generic inputs.

Import does not need the same recall behaviour as NL search. Either import should use a tighter setting (e.g. `Tiers: 1`), or assigned classes should be bounded (score threshold and/or max skill/domain count). Catalog projection can stay as-is: one tag per OASF skill/domain.

### Affected Version

`v1.7.0` / `main`

### Steps to Reproduce

1. Open [https://ai-catalog.outshift.io](https://ai-catalog.outshift.io) (or `GET https://ai-catalog.outshift.io/v1/agents/{cid}`).
[...]

PR #2106, opened 2026-09-03
fix(dir): restrict extractor results

csirmazbendeguz:
https://github.com/agntcy/dir/issues/2093

this PR restricts extractor:
- adds a cap of 10 to the skills and domains. so it will return the top 10 skills and top 10 domains for a record

### DIFF
--- a/cli/cmd/import/extractor_adapter.go
+++ b/cli/cmd/import/extractor_adapter.go
@@ -10,11 +10,17 @@ import (
 
 	enricherconfig "github.com/agntcy/dir-importer/enricher/config"
 	extractor "github.com/agntcy/dir/utils/extractor"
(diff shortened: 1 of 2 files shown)


akijakya:
Has the reason for the occasional high  number of skills from the extractor been found? Is it possible, that the extractor results are so low and close to each other in relevance in these cases, that half of the taxonomy gets sorted into 2 tiers?

csirmazbendeguz:
> Has the reason for the occasional high number of skills from the extractor been found? Is it possible, that the extractor results are so low and close to each other in relevance in these cases, that half of the taxonomy gets sorted into 2 tiers?

yes, that's pretty much it

muscariello (approved):
LGTM

PR #2108, opened 2026-09-03
fix(store): stop tagging referrers with their CID

### LINKED ISSUE #2102: [Refactor]: Stop CID-tagging referrers (90% of tags) — requires digest-based DeleteReferrer first
## Summary

Directory tags every referrer with its own CID. Referrers do not need tags — they are discovered through the OCI subject relationship — so 25,890 of production's 28,779 tags exist for no reason. They make `tags/list` return ten times more data than there are records, and they force the indexer to fetch and inspect a manifest for every one of them just to discard it.

Removing the tag is one line. Doing it safely is not, because `DeleteReferrer` finds referrers *by resolving that tag*, and it fails silently when the tag is absent. That prerequisite is the whole risk in this issue.

Related: #2101 (enable Zot's `deleteReferrers`; run step 3 after it), #2043 (indexer treats referrer CID tags as records — step 2 removes that root cause), #2070 (readiness probe times out on the full tag list — step 3 cuts that list by 90%), [Discussion #2072](https://github.com/agntcy/dir/discussions/2072).

## Problem

Every referrer push ends by tagging the manifest with the referrer's own CID:

https://github.com/agntcy/dir/blob/ecc1cd1c56204295ac0c11a685ee659baa64b6ea/server/store/oci/referrers.go#L119-L134

```go
manifestDesc, err := oras.PackManifest(ctx, s.repo, oras.PackManifestVersion1_1, ocispec.MediaTypeImageManifest,
	oras.PackManifestOptions{
		Subject:             &recordManifestDesc,
		ManifestAnnotations: annotations,
		Layers:              []ocispec.Descriptor{blobDesc},
	},
)
// ...
// Create CID tag for content-addressable storage
err = s.tagWithRetry(ctx, manifestDesc.Digest.String(), referrerCID)
```

The `Subject` field above is what actually makes the referrer discoverable: `WalkReferrers` uses the Referrers API, never the tag (`referrers.go:173-186`). The tag is redundant for reads.

What it costs:

| | prod today |
|---|---|
| records (legitimately tagged) | 2,889 |
| referrer CID tags | 25,890 |
| **total tags** | **28,779** |

The indexer pays for this every cycle. It lists all tags, diffs against its previous snapshot, and then — because referrers share the repository with records and are tagged the same way — must fetch each new tag's manifest to work out whether it is a referrer and skip it:

https://github.com/agntcy/dir/blob/ecc1cd1c56204295ac0c11a685ee659baa64b6ea/reconciler/tasks/indexer/task.go#L120-L123

```go
for _, tag := range newTags {
	// Referrers share the repository with records and are tagged with their
	// own CID, so they must be filtered out before attempting to index them.
	isReferrer, err := t.isReferrerTag(ctx, tag)
```

Untag the referrers and that filtering problem disappears along with 90% of the tag list. It is also the root cause of #2043 — no referrer tags, nothing for the indexer to mistake for a record.

## Prerequisite: `DeleteReferrer` must delete by digest first

This is the part that makes the ordering non-negotiable.

`DeleteReferrer` collects referrer CIDs from `WalkReferrers`, then switches on the repository type and deletes each CID through one of two helpers:

https://github.com/agntcy/dir/blob/ecc1cd1c56204295ac0c11a685ee659baa64b6ea/server/store/oci/referrers.go#L348-L356

```go
for _, cid := range cids {
	switch s.repo.(type) {
	case *oci.Store:
		err = s.deleteFromOCIStore(ctx, cid)
	case *remote.Repository:
		err = s.deleteFromRemoteRepository(ctx, cid)
```

**Both** helpers resolve the CID as a **reference**, and both return `nil` when resolution fails — `deleteFromOCIStore` at `internal.go:114-120` and `deleteFromRemoteRepository` at `internal.go:183-190`:

```go
if err != nil {
	// If manifest is completely missing (errdef.ErrNotFound), treat as successful deletion
	if stdErrors.Is(err, errdef.ErrNotFound) {
		internalLogger.Info("Manifest not found (never existed or deleted already)", "cid", cid)

		return nil
	}
```

A referrer CID is not a digest, so the only thing that can resolve it is its tag. Remove the tag and `Resolve` returns `ErrNotFound`, and the function **reports success while deleting nothing**. Referrer deletion — and therefore the cascade delete on record removal — would become a silent no-op. Note that this affects the embedded/local-directory daemon as much as the remote-registry deployment, so a fix must cover both helpers, not just the remote one.

The fix is available at no cost: `WalkReferrers` already holds the referrer's manifest descriptor when it iterates the referrers listing, and that descriptor carries the digest. Thread it through so deletion targets `manifestDesc.Digest` instead of the CID string.

**Ship the digest-based delete on its own, verify it against today's still-tagged referrers, and only then remove the tagging.** It is a strict improvement even if the rest of this issue never lands.

### Step 1 — delete by digest (prerequisite, separately shippable)

Rework `DeleteReferrer` to carry the manifest descriptor from `WalkReferrers` through to deletion, in both the `*oci.Store` and `*remote.Repository` paths. Verify against the current registry, where referrers are still tagged and both addressing modes work — so a regression is visible immediately rather than after step 2.

### Step 2 — stop tagging new referrers

Delete the `tagWithRetry` call at `referrers.go:134`. Nothing else changes: the manifest is still pushed, still carries its `subject`, still discoverable.

### Step 3 — strip the existing referrer tags

A one-time job that issues `DELETE /v2/dir/manifests/<referrer-cid-tag>` for each referrer tag. Run it **after** #2101, so it only has the ~10,984 live referrers to process rather than all 25,890.

Deleting by tag is the correct operation here, and Zot's semantics are exactly what we want (`pkg/storage/common/common.go:384`):

> In case of delete by digest we remove the manifest right away from storage but in case of delete by tag we want to only remove the tag and handle the manifest based on retention rules later. If there are more than one tags with the same digest we want to keep the others. If and only if there are no other tags except the one we want to remove we need to add a new descriptor without a tag name to keep track of the manifest in `index.json` so it remains accessible by digest.

The manifest survives as an untagged descriptor and stays reachable by digest and through the Referrers API.

## Safety analysis

| Concern | Verdict | Evidence |
|---|---|---|
| Will untagged referrers be garbage collected? | **No.** Any manifest with a non-nil subject is marked referenced before the untagged sweep. | Zot `gc.go:835`; dev's 152 untagged referrers have survived four months of hourly GC |
| Will untagged referrers still be discoverable? | **Yes.** `WalkReferrers` uses the Referrers API and the subject relationship, not tags. | `referrers.go:173-186` |
| Will `DeleteReferrer` still work? | **Only after step 1.** Otherwise it silently no-ops, in both repo paths. | `internal.go:114-120`, `internal.go:183-190` |
| Does this shrink `index.json` entry count? | **No.** Descriptors remain; only the `ref.name` annotation goes. | Zot `common.go:384-391` |
| Concurrent referrer writes? | **Safe on Zot**, which has a real Referrers API and no shared index document. This would *not* be safe on Distribution. | registry evaluation in Discussion #2072 |

**One thing this analysis does not cover.** `DeleteReferrer` is confirmed to resolve referrers by CID tag. Other code paths that might resolve a referrer by CID have **not** been exhaustively audited — the gRPC `PullReferrer` handler, the libp2p RPC referrer handlers, and `dirctl pull --signature/--public-key/--scan-report` all appear to go through `WalkReferrers` and match on CID in memory, which is safe, but that is a reading of the code rather than a test. Note also that `fetchAndParseManifest` (`internal.go:44`) has the same resolve-CID-as-reference shape; it is called with record CIDs, which stay tagged, but any future referrer caller would break. **Audit and test these before step 2**, because the failure mode is a silent 404 rather than an error.

## Expected effect

| | today | after |
|---|---|---|
| tags in `dir` | 28,779 | 2,889 |
| `index.json` entries | 13,873 (post #2101) | 13,873 — unchanged |
| `index.json` size | ~3.7 MB (post #2101) | ~2.8 MB |
| indexer manifest inspections per new tag | 1 per referrer | 0 |

Be precise about what moves and what does not. #2101 removes entries and roughly halves the file. This issue removes *tags*: a modest further size reduction from dropping the annotation, but a large reduction in indexer work and in the cost of `tags/list`, which is server-paginated and currently a hundreds-of-round-trips crawl — the same crawl that makes the readiness probe in #2070 time out.

There is also a forward-looking reason. Every non-Zot registry evaluated in Discussion #2072 expects referrers to be untagged; CID-tagging them is the single largest source of migration friction. Doing this now is useful whether or not we stay on Zot.

## Verification

Step 1, against today's tagged referrers:

1. Attach 3 referrers, `DeleteReferrer` one by CID, confirm it is gone and the other two remain.
2. Delete a record with referrers and confirm all referrers are removed.
3. Exercise both repo paths: a local-directory daemon (`*oci.Store`) and one against a remote registry (`*remote.Repository`).

Step 2, on a local daemon:

4. Push a record and attach a signature. `GET /v2/dir/tags/list` contains the record CID and **not** the referrer CID.
5. `dirctl pull --signature`, `--public-key`, `--scan-report` all return payloads; `dirctl verify` reports `trusted`.
6. Delete the record; confirm the referrer manifest 404s — this is the regression that step 1 exists to prevent.
7. Restart Zot, wait two GC intervals, confirm the untagged referrer is still fetchable by digest.

Step 3, in production:

8. Tag count falls to ~2,889; `index.json` entry count is unchanged.
9. Sample 50 records: `dirctl verify` still `trusted`, referrer payloads still retrievable.
10. Indexer log shows no referrer-tag skips on the next cycle.

## Rollback

Steps 1 and 2 are a commit revert. Step 3 is not reversible — the tags are gone — but it is also not damaging, because nothing reads them once step 2 has shipped. If you revert step 2 after step 3, newly pushed referrers get tags again and the historical ones stay untagged; both are discoverable, so the mixed state is harmless. Dev has been running in exactly that mixed state since February.

tkircsi:
Closes #2102

Referrers are no longer tagged with their CID. Deletion now works from the manifest descriptor, which is the only thing those tags were needed for.

Also fixes two findings from the #2095 review: superseded scan reports are removed in one batched call instead of one full referrer walk per report, and the prune checks the referrer type itself, since the store's type filter passes every custom type.

Verified against a local Zot with `deleteUntagged` and `deleteReferrers` on: signing adds 0 tags and 2 untagged referrers, `pull --signature/--public-key/--scan-report` and `verify` all work, and an untagged referrer whose subject is alive survives a GC pass.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/reconciler/tasks/indexer/manifest.go
+++ b/reconciler/tasks/indexer/manifest.go
@@ -29,10 +29,13 @@ type manifestReader interface {
 // isReferrerTag reports whether the manifest behind tag is a referrer artifact
 // (signature, public key, scan report, ...) instead of a record.
 //
-// Referrers are stored in the same repository as their subject record and are
-// tagged with their own CID, so a tag parsing as a CID does not make it a
-// record. Referrer manifests are identified by the OCI subject field that links
+// Referrers are stored in the same repository as their subject record. New ones are not tagged,
+// but those written earlier still carry a tag of their own CID, so a tag parsing as a CID does
+// not make it a record. Referrer manifests are identified by the OCI subject field that links
 // them to the record they describe, and by the Dir referrer type annotation.
+//
+// This check is therefore still needed, and stays needed until those legacy tags are stripped
+// from the registries.
 //
 // If the tag cannot be inspected, the error is returned and the caller decides
 // how to proceed; an unreadable manifest is not assumed to be a referrer.
--- a/reconciler/tasks/scan/referrer_test.go
+++ b/reconciler/tasks/scan/referrer_test.go
@@ -21,11 +21,13 @@ type fakeReferrerStore struct {
 	pushedCID string
 	pushErr   error
 
-	existing []*corev1.RecordReferrer
-	walkErr  error
+	existing  []*corev1.RecordReferrer
+	walkErr   error
+	walkCalls int
 
-	deleteErr error
-	deleted   []string
+	deleteErr   error
+	deleted     []string
+	deleteCalls int
 }
 
 func (f *fakeReferrerStore) PushReferrer(_ context.Context, _ string, _ *corev1.RecordReferrer) (*corev1.ReferrerRef, error) {
@@ -37,12 +39,14 @@ func (f *fakeReferrerStore) PushReferrer(_ context.Context, _ string, _ *corev1.
 }
 
 func (f *fakeReferrerStore) WalkReferrers(_ context.Context, _ string, referrerType string, walkFn func(*corev1.RecordReferrer) error) error {
+	f.walkCalls++
+
 	if f.walkErr != nil {
 		return f.walkErr
 	}
 
 	for _, ref := range f.existing {
-		if referrerType != "" && ref.GetType() != referrerType {
+		if !sharesArtifactType(referrerType, ref.GetType()) {
 			continue
 		}
 
@@ -54,14 +58,42 @@ func (f *fakeReferrerStore) WalkReferrers(_ context.Context, _ string, referrerT
 	return nil
 }
 
+// sharesArtifactType reproduces the OCI store's type filter, which is coarser than it looks:
+// apiToOCIType maps every referrer type except signatures and public keys onto one shared media
+// type, so a walk filtered on scan reports also yields every other custom referrer. The fake
+// matches that, otherwise it would hide the task's need to check the type itself.
+func sharesArtifactType(want, got string) bool {
+	if want == "" || want == got {
+		return true
+	}
+
+	distinct := func(t string) bool {
+		return t == corev1.SignatureReferrerType || t == corev1.PublicKeyReferrerType
+	}
+
+	return !distinct(want) && !distinct(got)
+}
+
 func (f *fakeReferrerStore) DeleteReferrer(_ context.Context, _ string, referrerCID string, _ string) ([]string, error) {
 	if f.deleteErr != nil {
 		return nil, f.deleteErr
 	}
 
+	f.deleteCalls++
 	f.deleted = append(f.deleted, referrerCID)
 
 	return []string{referrerCID}, nil
+}
+
+func (f *fakeReferrerStore) DeleteReferrers(_ context.Context, _ string, referrerCIDs []string, _ string) ([]string, error) {
+	if f.deleteErr != nil {
+		return nil, f.deleteErr
+	}
+
+	f.deleteCalls++
+	f.deleted = append(f.deleted, referrerCIDs...)
+
+	return referrerCIDs, nil
 }
 
 // scanReferrer builds a stored scan report referrer as WalkReferrers would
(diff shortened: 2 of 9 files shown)


reconciler/tasks/indexer/manifest.go
```diff
@@ -29,11 +29,14 @@ type manifestReader interface {
 // isReferrerTag reports whether the manifest behind tag is a referrer artifact
 // (signature, public key, scan report, ...) instead of a record.
 //
-// Referrers are stored in the same repository as their subject record and are
-// tagged with their own CID, so a tag parsing as a CID does not make it a
-// record. Referrer manifests are identified by the OCI subject field that links
+// Referrers are stored in the same repository as their subject record. New ones are not tagged,
```
muscariello, line 32: this is obscure. New in what sense and earlier in what sense ?

tkircsi: This change, doesn't remove the already tagged referrers tags, but does not create tags for new ones.

tkircsi: It is not an existing data cleanup PR.

muscariello (approved):
LGTM

PR #2113, opened 2026-09-07
ci(release): sign release artifacts and backfill published releases

paralta:
The OpenSSF Scorecard Signed-Releases check scores 0 because we publish bare `dirctl` binaries with nothing attesting to their origin. This adds a provenance step so every future release ships a Sigstore-signed build attestation, plus a one-off workflow to retrofit signatures onto the releases the check currently inspects.

- The backfill workflow is temporary — it only fixes already-published releases and should be deleted once run.
- Backfilled releases get `cosign sign-blob` signatures rather than provenance, since a build attestation can't be produced retroactively.
- Backfilling is dispatch-triggered rather than run locally because keyless signing binds the certificate to whoever signs, so CI attributes the signatures to the repository instead of a maintainer.
- No detached signatures on new releases: provenance outranks a plain signature in the check's scoring.

### DIFF
--- a/.github/workflows/backfill-release-signatures.yaml
+++ b/.github/workflows/backfill-release-signatures.yaml
@@ -1,0 +1,41 @@
+# Copyright AGNTCY Contributors (https://github.com/agntcy)
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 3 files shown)


muscariello (approved):
LGTM

PR #2114, opened 2026-09-07
chore(deps): scan npm and python dependencies and fix reported vulnerabilities

paralta:
https://scorecard.dev/viewer/?uri=github.com/agntcy/dir Scorecard reports four vulnerabilities in the gateway UI and the MkDocs toolchain that Renovate never had a chance to catch — its enabled managers covered Go, containers, and CI, but not npm or Python. This enables both managers and bumps the affected packages.

The remaining Scorecard findings are Go advisories with no upstream fix published, so they're out of scope here.

### DIFF
--- a/docs/mkdocs/uv.lock
+++ b/docs/mkdocs/uv.lock
@@ -460,7 +460,7 @@ wheels = [
 
 [[package]]
 name = "mkdocs-material"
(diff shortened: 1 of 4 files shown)


muscariello (approved):
LGTM

PR #2117, opened 2026-09-08
fix(ui): remove the Verified catalog filter

conorbronsdon:
The catalog's Verified filter depends on publishers hosting domain-verification keys. As described in #2087, that requirement makes the filter return almost no results and confuses catalog users.

Remove the checkbox and its `verified=true` query clause. Trusted and Safe continue to work, and record verification metadata stays available to the rest of the UI.

Closes #2087.

Validation from `server/gateway/ui`:

- `npm test`: all 16 tests pass. The regression passes a legacy Verified selection alongside Trusted and Safe, then checks the exact query. It fails against the original implementation.
- `npm run check`: zero errors and one existing accessibility warning in `DetailModal.svelte`.
- `npm ci` and `npm run build`: pass. Generated static assets are included, and a second production build produces no changes to those assets.

Prepared with Codex assistance.

PR #2118, opened 2026-09-08
test(e2e): cover the remote extractor in Kind

### LINKED ISSUE #2091: Add Kind e2e coverage for the remote OASF extractor backend
### Problem Statement

Nothing verifies the remote extractor backend. #2073 wires an optional OASF-SDK subchart into the apiserver chart and derives `config.extractor.remote_addr` from the rendered Service, but no automated test exercises that path.

Two gaps combine to make the Kind environment inert:

- **CI never runs a Kind environment.** Every entry in the `reusable-test-e2e.yaml` matrix is a `local` variant — `client:local`, `local:local`, `network:local`, `daemon:external`, `integration-server:default`. `e2e:test:local:kind` exists in the test matrix but is never invoked.
- **The existing natural-language suite is host-assets-only.** `tests/e2e/local/15_nl_search_test.go` gates on a locally provisioned extractor manifest and drives `dirctl`, so it cannot cover a deployment where the extractor is a separate in-cluster service and there are no host-side assets. (It is also skipping unconditionally for an unrelated reason — see the linked bug.)

The consequence is that enabling the subchart in the Kind values proves nothing automatically. The integration has so far only been verified by hand.

### Proposed Solution

Add a Kind-based test path that calls the extractor-backed gateway endpoints without requiring host-side assets, so the derived Service address and `ExtractorService` registration are actually asserted:

- Add a Kind job to the e2e workflow matrix, or extend an existing one, so `e2e:test:local:kind` runs.
- Add a test that hits `POST /v1/extract` and `POST /v1/search` through the HTTP gateway against the deployed cluster, rather than driving `dirctl` against host-provisioned assets.
- Assert the failure mode that is easy to miss: without `OASF_SDK_EXTRACTOR_OASF_URL` the pod starts, passes its gRPC health check, and fails every call with `unknown service agntcy.oasfsdk.extractor.v1.ExtractorService`. A green health check is not evidence the extractor works.

### Notes

From image `v1.3.0` the embedding model ships inside the image, so a runner needs no access to `huggingface.co`. The server does still fetch the taxonomy from `OASF_SDK_EXTRACTOR_OASF_URL` on every start and takes roughly 40 seconds to become ready, which the job needs to tolerate.

### Acceptance criteria

- [ ] `e2e:test:local:kind` runs in CI
- [ ] A test calls the extractor-backed gateway endpoints against the Kind deployment with no host-provisioned extractor assets
- [ ] The test fails if `extractor.remote_addr` is wrong or `ExtractorService` is unregistered

conorbronsdon:
Adds a Kind job to the e2e matrix and exercises the deployed HTTP gateway's `/v1/extract` and `/v1/search` endpoints without restoring or provisioning extractor assets on the runner. The tests publish a local fixture, check relevant extracted taxonomy, and require natural-language search to return that fixture's exact CID.

The job installs the pinned Kind/Kubectl tools and locked Helm dependencies. Extraction has a bounded startup wait; a persistently wrong remote address or unregistered extractor service fails the test. The Kind naming fixture uses its fully qualified Service hostname so the existing naming tests can run in the newly enabled job.

Closes #2091.

Validation:

- Full local suite against Kind, with a fresh non-root runner and no host SDK assets: 220 passed, 0 failed, 4 existing skips.
- Wrong remote address: extraction test fails with HTTP 503 after its bounded retry window.
- Empty `OASF_SDK_EXTRACTOR_OASF_URL`: pod passes readiness, but the test fails; apiserver logs confirm `unknown service agntcy.oasfsdk.extractor.v1.ExtractorService`. Restoring configuration returns both new tests to green.
- Go test compilation, pinned golangci-lint 2.13.2, and `git diff --check` pass.

Local deployment validation reused server images from #2117's successful CI build (its changes are UI-only). This PR's CI builds its own revision.

Implemented with Codex assistance and independent model review.

### DIFF
--- a/.github/workflows/reusable-test-e2e.yaml
+++ b/.github/workflows/reusable-test-e2e.yaml
@@ -95,7 +95,7 @@ jobs:
 
       - name: Verify Kind has no host extractor assets
         if: matrix.task == 'e2e:test:local:kind'
-        run: test ! -e "$HOME/.agntcy/oasf-sdk/extractor/manifest.json"
+        run: test ! -e "$HOME/.agntcy/oasf-sdk"
 
       - name: Run end-to-end tests
         env:
--- a/.github/workflows/reusable-test-e2e.yaml
+++ b/.github/workflows/reusable-test-e2e.yaml
@@ -29,6 +29,8 @@ jobs:
             label: "Client"
           - task: e2e:test:local:local
             label: "Local"
+          - task: e2e:test:local:kind
+            label: "Local Kind"
           - task: e2e:test:network:local
             label: "Network"
           - task: e2e:test:daemon:external
@@ -61,6 +63,10 @@ jobs:
           cache-dependency-path: "**/*.sum"
           cache: true
 
+      - name: Install Kind test dependencies
+        if: matrix.task == 'e2e:test:local:kind'
+        run: task e2e:deps:kind
+
       - name: Download build artifacts
         uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
         with:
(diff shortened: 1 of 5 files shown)


ramizpolic (approved):
LGTM, thanks @conorbronsdon!

PR #2119, opened 2026-09-08
docs(proposal): record quality gate — policy engine for record admission and retention

### LINKED ISSUE #2128: [Epic]: Project-wide policy framework
### Definition

A content policy is a set of rules content should conform to.
It is enforced by automations, e.g. cron jobs.

### Implementation

As of now, content policies are implemented as Kubernetes CronJobs in the `dirctl` chart:

https://github.com/agntcy/dir/blob/ad1c172d9f11b020d682c3f86d8d77f2b55710c5/install/charts/dirctl/values.yaml#L94-L111

For each policy, the Helm chart generates a CronJob with a Bash script containing `dirctl` commands. For example, the prune job executes `dirctl search` and pipes the results into `dirctl delete`.

#### PROS:

- Decoupled from Directory. Jobs can be deployed without restarting Directory.

#### CONS:

- Limited functionality. The job is limited to the `dirctl search` command's capabilities. For example, it's not possible to prune old versions of records because `dirctl search` doesn't have a flag for that.
- Kubernetes only. Directory may be deployed in other environments (e.g. `dirctl daemon`) - those users currently don't have an official way to use content policies.

The supported actions are:
- sync
- prune
- publish
- unpublish

## Problem

The problem with the current implementation is **limited functionality** and that it's **Kubernetes only**.

### Limited functionality

For example, one functionality we would like to support is to prune old versions of records.

If there are two versions of a record (e.g. org.agntcy/directory):
- v1.1.0
- v1.0.0

The content policy should retain v1.1.0 and delete v1.0.0.

There's no built-in flag in `dirctl search` to detect old versions - nor should there be one. `dirctl search` is not meant for prune operations. So the current implementation is *limited in functionality* by `dirctl search`.

### Second-class citizen

Kubernetes CronJobs should not be the only automation that can enforce content policies.

Ideally, content policies should be enforced as soon as possible - as soon as the record is ingested, if possible.

Kubernetes CronJobs are a crutch - they prune records after they are already ingested by Directory.

Instead, Directory should treat content policies as **first-class citizens**, and enforce them as soon as possible.

For example, if there's a content policy to prune records with high severity scan results, Directory shouldn't wait for a cron job. It should prune the record as soon as it gets the scan results.

https://github.com/agntcy/dir/pull/2119 describes this in more detail.

### 1. Extend the Helm chart

The simplest solution is to extend the Helm chart with a Bash script to detect old versions.

#### PROS:

- Decoupled from Directory (still).

#### CONS:

- Bash scripts are harder to test, can't use the ORM and other abstractions we have in the Directory source code. I would much rather use Go for complex programs (e.g. a program for detecting old versions of records).
- As of now, the Bash scripts are dynamically generated by Helm. This isn't maintainable or readable. It's hard to extend. If we really want to stick with Bash, the CronJobs should be consolidated into a single Bash script
- Kubernetes only (still).
- Second-class citizen (still).

### 2. Migrate CronJobs to reconciler

This is the first step in the right direction. The content policy jobs should be handled by the reconciler. This solution would make policies part of the reconciler config.

#### PROS:

- Works everywhere, NOT Kubernetes only. For example, works with `dirctl daemon` too.
- It's better to use Go for this than Bash. It's more in-line with the project. We can test the code as part of the E2E test suite. We can use the ORM and other abstractions.

#### CONS:

- Not decoupled from Directory. To deploy a new job, reconciler needs a restart.
- Second-class citizen (still).

### 3. Content Policy Engine

This is the solution I had in mind and what I think we should ultimately strive for.

This solution:
- Adds a `dirctl policies` command. With this command, admins can configure policies for their Directory instance.
- The policies are stored in the database.
- The reconciler runs a job that enforces the policies (same as cron jobs).
- Directory can programmatically access these policies - enforcement is not limited to the reconciler job.

#### PROS:

- Works everywhere.
- Go instead of Bash.
- Decoupled. Admins can change policies without restarting Directory.
- First-class citizen content policies.

#### CONS:

- The reconciler only needs a restart if the reconciler job config is changed (e.g. enabled/disabled, modifying the schedule).

tkircsi:
## What this is

A **draft proposal**, not an implementation. It asks for a decision on where a
record quality gate belongs in the Directory process, and what its policy model
should look like.

Nothing here is committed to. The twelve numbered questions in
`open-questions.md` are the point of the review — each has a recommendation, and
each is genuinely arguable.

Also adds `proposals/` for design docs too large for a PR description. Kept out
of `docs/` deliberately: that tree is published with mkdocs and `docs-ci` /
`docs-deploy` trigger on `docs/**`.

## The problem

We want a configurable gate that decides which records a node is willing to
admit and keep, across push, import, sync and autosync. The two motivating
examples were:

1. only keep records whose scan says they are trusted
2. keep only 2 versions of the same record

## The three findings worth arguing about

**1. It has to be two gates, not one.** Neither example can be enforced at push
time. Trust is a *derived* fact that the signature, name and scan reconciler
tasks compute later (defaults 1m / 1h / 6h); at push the node does not even know
the record is signed, because `dirctl push --sign` sends the signature afterwards
as a separate referrer. And "keep 2 versions" is a property of a *set* grouped by
name, not of the record being admitted. So:

- **Gate A — admission (synchronous)**: decorate the existing `ingest.Ingestor`
  interface. Already an interface, constructed once, already documented as "a
  single, authoritative code path". Covers push, `dirctl import` and DHT
  autosync with no call-site changes. Can only see record metadata.
- **Gate B — enforcement (asynchronous)**: a new `reconciler/tasks/policy/`.
  Runs alongside the tasks that produce the facts, and evaluates against the
  search DB where `trusted` / `safe` / `verified` already have query builders.

**2. Reuse `RecordQuery` rather than adopting Rego or CEL.** `dirctl search`
already exposes a tested predicate vocabulary that covers exactly the facts we
care about. A policy becomes `match: {trusted: false}` + `action:`. Gate B then
compiles to one SQL query instead of loading records into memory; authors can
dry-run any predicate with `dirctl search <same flags>` before enabling it; and
the existing `dirctl` chart `policies:` collapse into a client of this model
instead of a parallel implementation. The cost is a flat AND of predicates — no
OR, no cross-field logic. `cel:` is the proposed escape hatch if we hit that.

**3. Destructive actions need CID tombstones.** Deleting a record that peers
still announce means autosync re-fetches it, the reconciler re-scans it, and
policy deletes it again — a tight loop at our 1m republish interval. `delete`
must also unpublish (note `StoreService.Delete` does not touch routing
announcements today), and a CID denylist must suppress the re-fetch. Content
addressing makes this precise and self-invalidating, and it is the *only* filter
the regsync path can apply, since that path's tag filter is already a CID list.

The sharpest sub-question there: a tombstone must **not** be created for
`trusted: false`, because that is absence of a fact rather than a negative
verdict, and it resolves on its own. Tombstoning it would permanently and
invisibly block records whose key merely had not propagated yet.

## Reading order

| File | Contents |
|---|---|
| `proposals/quality-gate/README.md` | placement in the process, phase model, engine choice, rollout |
| `proposals/quality-gate/policy-model.md` | schema, predicate vocabulary, actions, worked examples for both motivating cases |
| `proposals/quality-gate/tombstones.md` | the CID denylist, its limits, and staleness handling |
| `proposals/quality-gate/open-questions.md` | the twelve decisions |

If you only read one thing, read §3 of the main README (the fact-availability
table). Everything else follows from it.

## Two pre-existing issues this depends on

- `skill.Publish` writes to the store and DB directly instead of going through
  `ingest.Ingestor`, so it sits permanently outside Gate A.
- `StoreService.Delete` leaves routing announcements behind, which the `delete`
  action would inherit.

## Suggested rollout

Gate B in report-only mode first, with config-file policies. It tells us what
our nodes actually hold at zero blast radius and validates the vocabulary
against real data before anything is deleted. `PolicyService` and the API come
later, following the `SyncService` pattern — which is also the only way to get
hot-reload, since nothing in the repo watches config today.

### DIFF
--- a/proposals/README.md
+++ b/proposals/README.md
@@ -1,0 +1,27 @@
+# Proposals
+
+Design documents for changes too large to explain in a pull request
+description. A proposal captures the problem, the constraints discovered while
+investigating the code, the options considered, and the open decisions — so that
+the reasoning survives past the discussion that produced it.
+
+These are **not** user documentation. Published docs live in `docs/content` and
+are built with mkdocs; nothing here is part of the docs site.
+
+A proposal is a folder, `proposals/<topic>/`, containing at least a `README.md`
+that states its status in the first lines. Status is one of:
+
+- **draft** — under discussion, nothing implemented, subject to change
+- **accepted** — agreed direction; implementation may be in progress or partial
+- **implemented** — shipped; kept for the reasoning, may drift from the code
+- **rejected** / **superseded** — kept deliberately, with a note explaining why
+  and a pointer to what replaced it
+
+Proposals are merged rather than left on a branch: the decisions and the
+discarded alternatives are the valuable part, and they need somewhere durable to
+live. A proposal being merged means "this is a real record of a discussion", not
+"this is approved" — that is what the status line is for.
+
+| Proposal | Status |
+|---|---|
+| [`quality-gate`](quality-gate/) — policy engine for admission and retention of records | draft |
--- a/proposals/quality-gate/README.md
+++ b/proposals/quality-gate/README.md
@@ -1,0 +1,281 @@
+# Record quality gate
+
+Status: **draft** — under discussion, nothing implemented, subject to change.
+
+Recommendations below are starting positions for the discussion, not decisions.
+The twelve numbered questions in `open-questions.md` are the parts that most
+need other people's judgement.
+
+A configurable policy engine that decides which records a Directory node is
+willing to keep, and which ones it should get rid of. Applies to records
+arriving by push, import, sync and autosync.
+
+Contents:
+
+- `README.md` — this file: placement in the process, phase model, engine choice
+- `policy-model.md` — the policy schema, predicate vocabulary, actions, worked examples
+- `tombstones.md` — the CID denylist that makes destructive actions safe
+- `open-questions.md` — decisions we need to make before writing code
+
+---
+
+## 1. The two motivating examples, and why they don't belong in the same place
+
+The two examples we started from look similar but have opposite requirements.
+
+**"Only keep records whose scan says they're trusted."** Trust is not a property
+of the record bytes. It is a *derived fact* that a reconciler task computes later
+by fetching keys, checking signatures and running scanners. At the instant a
+record is pushed, the node knows nothing about its trust. So this rule cannot be
+a synchronous admission check unless we are willing to do live signature
+verification and a full security scan inline on the push path — which would turn
+a sub-second push into a multi-second-to-multi-minute operation and make push
+availability depend on external scanners.
+
+**"Keep only 2 versions of the same record."** This is not a property of a single
+record at all. It is a property of a *set* of records grouped by name. You cannot
+evaluate it by looking at the record being admitted; you evaluate it by sweeping
+the store.
+
+Both therefore land after the fact. But we still want a synchronous gate, because
+some rules genuinely are decidable at push time (naming conventions, required
+annotations, allowed skills/domains, size, schema version, source peer) and for
+those, rejecting the push with a clear error is far better UX than silently
+deleting the record 6 hours later.
+
+**Conclusion: the quality gate is two gates, not one.**
+
+---
+
+## 2. What already exists (and why none of it is this)
+
+| Thing | Where | What it does | Why it isn't a quality gate |
+|---|---|---|---|
+| OASF validation | `api/core/v1/record.go:179` (`ValidateWith`) | Schema validity + 4 MB size cap | Well-formedness only, not policy; not configurable |
+| Autosync peer allow-list | `server/routing/autosync/autosync.go:171` (`MaybeEnqueue`) | Ignores announcements from untrusted peers | Source-based only, no record predicates. Useful precedent though — it is admission control by another name |
+| dirctl chart `policies:` | `install/charts/dirctl/templates/_policies.tpl` | CronJobs running `dirctl search` → `sync`/`prune`/`publish`/`unpublish` | Client-side, outside the node, best-effort, no API, no audit trail. But its **match/action shape is the right model** — see §5 |
+| Casbin authz | `server/authz/` | Which SPIFFE trust domain may call which gRPC method | Access control over *callers*, not content |
+
+The chart policies are the closest existing thing and are essentially a
+prototype of Gate B implemented outside the server. Part of this work is moving
+that capability inside the node so it gets an API, a durable configuration, and
+an audit trail.
+
+---
+
+## 3. When is each fact knowable?
+
+This table is the single most important input to the design. Everything else
+follows from it.
+
+| Fact | Available at record push? | Populated by | Default interval |
+|---|---|---|---|
+| `name`, `version`, `schema_version`, `description`, `authors` | yes | `ingest.ImportRecord` → `db.AddRecord` | — |
+| skills, domains, modules, locators, annotations | yes | same | — |
+| `oasf_created_at`, record size, CID | yes | same | — |
+| source (peer ID / trust domain / RPC) | yes, from context | — | — |
+| `signed` | **no** — signature arrives as a *separate referrer push* after the record | `ingest.ImportReferrer` | — |
+| `trusted` (signature verified) | no | signature reconciler task | 1m |
+| `verified` (name ownership) | no | name reconciler task | 1h |
+| `safe`, `max_severity` (scan) | no¹ | scan reconciler task | 6h |
+| version count for a name | no (set property) | — | — |
+
+¹ Exception: a `ScanReport` **referrer** received over sync is written straight
+into `scan_reports` (`server/ingest/ingest.go:134-155`), so scan facts can arrive
+with a synced record. That's a peer's verdict, not ours — the policy model needs
+to be able to say whether a foreign verdict counts.
+
+Anything in the "no" rows is a Gate B rule. Full stop.
+
+---
+
+## 4. Where the gates go
+
+### Gate A — admission (synchronous, deny the write)
+
+**Hook: decorate the `ingest.Ingestor` interface** (`server/ingest/ingest.go:33`).
+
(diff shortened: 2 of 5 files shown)


Santoshkumarpuppala:
Read against `dir` main `2e661ac`. Two things sitting under §3's table.

§6 says `scan-status` already distinguishes absence from a verdict, "so the vocabulary supports it". That holds for scan — `RECORD_QUERY_TYPE_SCAN_STATUS` is a real query type (`record_query.proto:115`). It does not hold for the other two: `VERIFIED` (13) and `TRUSTED` (14) are booleans with no status counterpart, and `trusted` is the fact §6's grace-window paragraph is about.

The two families also fail in opposite directions when their producer isn't running (`server/database/gorm/record.go`):

- `trusted: false` → `NOT EXISTS (… signature_verifications … status = 'verified')` (`:602`)
- `verified: false` → `LEFT JOIN name_verifications … status IS NULL OR != 'verified'` (`:591-592`)
- `safe`, either polarity → gated on `EXISTS (… scan_reports … status IN ScannedStatuses)` (`:616-626`)

The first two match every record in the store; the third matches none. So `match: {trusted: false}, action: quarantine` doesn't eat records in a window — it eats the store. A `scan-severity` policy is a silent no-op whose report-only output reads `0 matched`, same as a clean store.

Not hypothetical. `reconciler/config/config.go` sets `signature.enabled = true` (`:196`), `name.enabled = false` (`:181`), `scan.enabled = false` (`:211`); `install/charts/dir/values.yaml` sets all three `true` (`:757`, `:762`, `:767`). Same policy and same records, opposite enforcement, decided by install method.

Three small changes: add a "produced by default?" column to §3, split binary vs chart; extend Q1's reasoning — "a policy that looks active and isn't" — from admission to an enforcement policy whose producer is disabled, refusing to load or exporting a `policy_inert` gauge; and separate "0 matched" from "0 had the fact at all" in report-only output.

The `trusted`/`verified` status query types look like the smaller proto addition, and would make §6's grace condition expressible for the fact it was written for.

tkircsi:
Thank you @Santoshkumarpuppala — this is a really valuable review, and exactly the kind of engagement I was hoping the open questions would draw out. I checked every claim against the code and they all hold.

Confirmed:

- `RECORD_QUERY_TYPE_VERIFIED` (13) and `TRUSTED` (14) are booleans with no status counterpart, while `SCAN_STATUS` (20) has one. I generalised a property of the scan family to trust, so §6 is wrong as written.
- The opposite failure directions are real. `trusted: false` → `NOT EXISTS (… status='verified')` and `verified: false` → `LEFT JOIN … IS NULL` both match everything when no producer rows exist; `safe` and `scan-severity` are gated on `EXISTS (scan_reports … status IN ScannedStatuses)` and match nothing.
- The default divergence is real, with one refinement: it is three config trees and the standalone reconciler is the outlier. `reconciler/config/config.go` has `name=false` / `scan=false`, while both `cli/cmd/daemon/daemon.config.yaml` and `install/charts/dir/values.yaml` enable all three.

One amplification of your point: `GetRecordsNeedingSignatureVerification` filters on `records.signed = true` (`server/database/gorm/signature.go:212`), so an unsigned record never receives a `signature_verifications` row at all. `trusted: false` therefore matches it permanently rather than transiently — the `minAge` grace window in §6 does not bound the damage, it only delays it by a day. So `policy-model.md` example 1 is not merely racy; it would quarantine every unsigned record in the store.

Your third suggestion is the one that concerns me most operationally. I proposed report-only Gate B as the zero-risk first phase, but a scan policy on a node with `scan.enabled=false` prints `0 matched`, indistinguishable from a clean store — so the phase I presented as safe can produce a false all-clear.

Would you be willing to take the document and open a PR with the corrections? Your three changes, plus — if you agree with the direction — the general form of them: derived-fact predicates want three-valued logic (true / false / unknown), and an enforcement policy should have to match a *positive verdict* rather than the absence of one. Your `TRUSTED_STATUS` / `VERIFIED_STATUS` query types look like the minimal concrete version, given the rows already carry `status`.

Please don't feel constrained by what today's code makes convenient. This is a proposal, so describing the process you think is correct and then the code changes needed to get there is more useful than working around the current shape. Happy to hand over the branch or to review whatever you open.

csirmazbendeguz:
Thanks @tkircsi !

I feel like the "tombstones" are out of scope for this issue? Surely, this affects all deletes not just the quality gate?

Santoshkumarpuppala:
Happy to take it — send the branch over and I'll open the PR.

Your amplification is right, and I checked it before agreeing rather than taking it on faith: `GetRecordsNeedingSignatureVerification` filters `records.signed = true` (`server/database/gorm/signature.go:212`), so an unsigned record never receives a `signature_verifications` row at all. That makes `trusted: false` permanent for those records rather than transient — and it means `minAge` cannot bound the damage, since a grace window only helps where a row is eventually coming.

Your refinement on the defaults is also right, and mine was incomplete. `cli/cmd/daemon/daemon.config.yaml:119-133` enables all three, same as the chart, so `reconciler/config/config.go` is the single outlier rather than a Go-versus-chart split.

On the general form: agreed, and it is the better framing than my three fixes. It buys two things. A predicate that can return unknown lets report-only separate "0 matched" from "0 had the fact at all". And requiring a positive verdict makes the failure direction uniform, so `trusted` and `safe` stop failing in opposite directions on the same store.

What I would put in the PR:

1. `TRUSTED_STATUS` and `VERIFIED_STATUS` query types mirroring `SCAN_STATUS` — both tables already carry `status`, so this is mostly plumbing.
2. §3's table gaining a produced-by-default column, split across the three config trees.
3. Q1's reasoning extended from admission to enforcement: refuse to load, or export a `policy_inert` gauge.
4. §6 rewritten so an enforcement policy matches a positive verdict, with the unsigned-record case as the worked example — it is the clearest one, because it never resolves.

I'll describe the process I think is correct first and then the code changes it implies, rather than fitting to what today's shape makes convenient.

PR #2125, opened 2026-09-08
feat(identity): record identity/ownership verifiable claims

ramizpolic:
## Summary

Adds a verifiable identity/ownership claim system for OASF records (identity.v1), purely additive — naming.v1 is left fully intact and untouched in this PR (removed in a follow-up stacked PR).

- **proto/api**: `IdentityClaim`/`OwnershipClaim` messages, `IdentityService` (`GetIdentityStatus`, `Resolve`), and the `api/identity/v1` Go package: claim construction, detached-JWS signing/verification (ES256/ES384/EdDSA/RS256), mandatory per-trust-domain SPIFFE trust bundles (no silent fallback), and OCI referrer marshaling. Adds `Record.GetIdentity/GetOwner/GetIdentityType/GetOwnerType` annotation accessors and `Record.GetDigest` (sha256:hex form, for OCI/ai-catalog.io interop).
- **server**: `server/identity` resolvers (dns/did/wellknown/spiffe) behind a common `Resolver` registry, the SSRF-safe `utils/safefetch` HTTP client they share, a unified `claims` DB table (record_cid + role), and eager claim verification wired into ingest.
- **client/cli/reconciler**: SDK methods (`SignAndAttachIdentityClaim`/`OwnershipClaim`, `GetIdentityStatus(ByName)`, `Resolve`), `dirctl identity claim/status/resolve`, identity/owner search flags, and a reconciler task that periodically re-verifies claims (catching key rotation, expiry, and revoked SPIFFE trust bundles).

## Design note

This PR is deliberately additive-only so it builds and passes tests standing alone: naming.v1's `NamingServiceClient`, its DB API (`NameVerificationDatabaseAPI`), CLI, and reconciler task are untouched. A couple of naming collisions between the old and new code were resolved by keeping both (e.g. an `expandNameWithProtocols` helper duplicated under a different name) rather than removing anything — naming.v1's actual removal happens in the follow-up stacked PR (`feat/identity-remove-legacy`).

## Verification

`go build && go vet && go test` pass across `api`, `server`, `client`, `cli`, `reconciler`, `tests`, `samples/ard-over-ads`.

PR #2152, opened 2026-09-14
feat(identity): add the ans scheme to claim verification

### LINKED ISSUE #2138: [Feature]: ans identity scheme: verify ans:// identity claims through DNS and the ANS transparency log
### Problem Statement

Name verification today has one method: the record's signing key must appear in a JWKS file at `https://<domain>/.well-known/jwks.json`. As #2087 notes, this "depends on record publishers actually hosting that well-known file at their domain, which essentially no publisher does today", so `dirctl search --verified` returns almost nothing.

Record identity is moving to identity.v1 in #2125: a record declares `agntcy.dir/identity`, the publisher pushes a detached-JWS `IdentityClaim` for it, and `server/identity` resolves the subject's key by URI scheme (`did:`, `https://`, `dns:`), with `spiffe://` verified against an embedded X.509-SVID and a configured trust bundle. #2126 then removes naming.v1. That is the right shape, and it is where this proposal now lives. The first version of this issue targeted naming.v1 and is withdrawn.

The self-hosted schemes in #2125 share the property #2087 describes for JWKS: each depends on the publisher hosting key material (a DID document, a JWKS file, a TXT record), and none carries revocation, history, or anything a third party can audit later. `spiffe://` has a real anchor, but the anchor is a trust bundle configured per deployment.

The Agent Name Service (ANS) is a population of agents whose identity material already exists in the shape identity.v1 wants. Every ANS registration produces:

- An `ans://v{MAJOR}.{MINOR}.{PATCH}.{agentHost}` name and a CA-issued identity certificate that carries the name as its URI SAN. Domain control is proven at registration.
- A DNS TXT record `_ans-badge.<agentHost>` with the value `v=ans-badge1; version=v1.0.0; url=https://<log>/v1/agents/<agentId>`, pointing at the agent's entry in an append-only transparency log. It is published under every ANS discovery profile with no extra step by the publisher.
- Log-signed statements about the agent: a status token carrying the current status (`ACTIVE`, `REVOKED`, `DEPRECATED`, `EXPIRED`, `WARNING`) and the `SHA256` fingerprints of the identity certificates currently valid for it, and a SCITT receipt for the logged registration event. Revocation is a logged event, not a deleted file.

Verifying an `ans://` identity claim needs one more resolver in the registry, not a new subsystem.

### Proposed Solution

Add an `ans` scheme to `server/identity` alongside `dns`, `https`, `did`, and `spiffe`. It plugs into #2125 as written.

**Subject.** `ans://v{MAJOR}.{MINOR}.{PATCH}.{agentHost}` as the record's `agntcy.dir/identity` annotation and the `IdentityClaim.subject`. `InferIdentityType` gains an `ans://` case; today the prefix falls through to `dns`.

**Claim.** The publisher signs the claim with the private key of the ANS identity certificate and includes the certificate in the JWS protected header as `x5c` (RFC 7515 §4.1.6). An `AnsSigner` next to `KeySigner` and `SpiffeSigner` does this and, like `SpiffeSigner`, checks the certificate's URI SAN against the claimed subject before signing. On the CLI it is the existing flag pair, accepted for `ans://` subjects:

```bash
dirctl identity claim --record <cid> --role identity \
  --subject ans://v1.0.0.agent.example.com --key identity.key --cert identity.crt
```

**Resolver.** A new `server/identity/ans` package implements `Resolver` with `Scheme() == "ans"`. `Verify` runs:

1. Parse the subject. Take the leaf certificate from `x5c`; require its URI SAN to name the same agent host and version, and the current time to fall within its validity period. No network call yet.
2. Resolve `_ans-badge.<agentHost>` TXT and select the value whose `version=` matches the subject's version. Accept the URL only if it is `https`, its host is on a configured allow-list of transparency logs, and its path is `/v1/agents/{agentId}`.
3. Fetch the agent's status token from the log and verify it against pinned root keys (or against the log's `/root-keys` over TLS, behind an explicit opt-in). Require the token's agent id and `ans://` name to match, a status of `ACTIVE`, `WARNING`, or `DEPRECATED`, and the certificate's fingerprint in the token's identity certificates.
4. Fetch and verify the log's SCITT receipt and require it to name the same agent and `ans://` name.
5. `VerifyJWS(jws, cert.PublicKey, payload)`, under the same key-type-derived algorithm rule as every other scheme.

Why the certificate travels with the claim: the log attests certificate fingerprints, the same shape as DANE `3 0 1`, not public keys, so a resolver cannot obtain the key from the log alone. `x5c` keeps `Resolver.Verify(ctx, subject, signature, payload)` and `verifyClaim` unchanged, and it is inert for every other scheme because `VerifyJWS` derives the algorithm from the resolved key and does not read headers. The alternative is to reuse `IdentityClaim.certificate` (field 5, today `spiffe://` only) and hand it to the resolver, which means adding a certificate parameter to `KeyResolver.Verify` for `dns`, `https`, and `did` to ignore. Either works for us.

**Configuration.** An `ans` block under `identity` in both the server and the reconciler, next to `spiffe_trust_domains`: `enabled`, `trusted_log_hosts`, `root_keys`, `allow_unpinned_root_keys`, `timeout`, `dns_server`, `ca_file`. Off by default; while off, an `ans://` subject fails with the registry's usual "no resolver registered". The resolver is added to both `NewRegistry` call sites, so ingest verifies eagerly and the identity task re-verifies on its interval, exactly as for the other schemes. A per-log circuit breaker inside the resolver keeps a black-holed log from stalling either loop.

**Result.** `ClaimStatus` as it is: `verified`, `subject`, `error_message`. Error strings are stage-prefixed and stable (`ans dns: …`, `ans badge-url: …`, `ans status-token: …`, `ans certificate: …`, `ans receipt: …`), never raw resolver output.

**Footprint in this repo.**

- `api/core/v1`: the `InferIdentityType` case. `api/identity/v1`: `AnsSigner`. Both additive.
- `server/identity/ans` (new) with its config package. One new dependency, `github.com/agentnameservice/ans-sdk-go` (MIT), for the badge-record parser, the log client, status token and receipt verification, and fingerprints.
- `server/server.go` and `reconciler/service/service.go`: one resolver each in `NewRegistry`. `server/identity/config` and the identity task config: the `ans` block, plus the daemon and Helm keys for it.
- `cli/cmd/identity`: `--cert` accepted for `ans://` subjects.
- Docs: a row in the scheme table of the trust-model doc, the CLI reference, and the verification reference.
- Nothing changes in the `Resolver` interface, `verifyClaim`, the `claims` table, `IdentityService`, or anything under naming.v1.

**Security.** The DNS pointer is untrusted; trust comes from the log's signatures over the status token and the receipt, checked against pinned root keys, and from the certificate binding. Log URLs must be `https`, on the allow-list, and shaped as `/v1/agents/{agentId}`; redirects are not followed. A claim whose certificate the log does not attest fails at step 3; a copied certificate fails at step 5, because the JWS is verified against that certificate's key. Revocation is a logged status that the identity task's next re-verification picks up.

**Delivery.** We maintain ANS. The resolver's trust path is implemented and tested against the ANS reference implementation, with golden fixtures captured from a running log and synthetic fixtures for every negative case, on top of naming.v1. We are porting it onto #2125's `Resolver` and will open the PR against `main` once #2125 lands, so the PR never carries #2125's diff. A Kind e2e with a local ANS stack, mirroring the dns-validation chart, follows separately.

### Alternatives Considered

- **Publish the identity key through an existing scheme (`did:web`, `https://` JWKS, `dns:` TXT).** Works today with no Directory change, but drops revocation, the logged history, and proof of domain control, and it reproduces #2087: it only works if publishers host key material.
- **Treat `ans://` like `spiffe://`: verify the embedded certificate against a configured bundle and stop there.** A certificate without an external anchor proves nothing about the subject's legitimacy, which is why #2125 requires a trust bundle for SPIFFE. The transparency log is that anchor for ANS, and it is shared rather than configured per deployment.
- **Publish a SubjectPublicKeyInfo hash in DNS on the ANS side so the certificate need not travel with the claim.** Avoids the `x5c` header, but discards the certificate binding (CA issuance, validity period) that is the point of the scheme, and diverges from the full-certificate fingerprints ANS already logs.

### Additional Context

- #2125 (identity.v1 and the resolver registry this plugs into) and #2126 (removal of naming.v1).
- ANS reference implementation (registration authority, transparency log, offline verifier): https://github.com/agentnameservice/ans. The log API this resolver consumes is `spec/api-spec-tl-v2.yaml` in that repo: `GET /v1/agents/{agentId}`, `GET /v1/agents/{agentId}/status-token`, `GET /v1/agents/{agentId}/receipt`, `GET /root-keys`.
- ANS Go SDK: https://github.com/agentnameservice/ans-sdk-go.
- DNS-AID: https://datatracker.ietf.org/doc/draft-mozleywilliams-dnsop-dnsaid/. ANS's default discovery profile emits SVCB records conformant to draft-02; those records are discovery data and are never consulted for trust.
- Related here: #638 (naming epic), #803 (introduced name verification), #833 (removed the DNS method), #2087 (Verified filter removal).

### Checklist

kperry-godaddy:
Closes #2138. Builds on #2125 (identity.v1) and is based on its branch, `feat/identity-add`, so the diff shows only this change; once #2125 merges I rebase onto `main`.

## What this adds

A fifth identity scheme, `ans`, next to `dns`, `https`, `did`, and `spiffe`. A record whose `agntcy.dir/identity` (or `agntcy.dir/owner`) is `ans://v{MAJOR}.{MINOR}.{PATCH}.{agentHost}` carries an `IdentityClaim` signed with the agent's Agent Name Service (ANS) identity key. The resolver verifies the claim through ANS rather than through anything the publisher self-hosts:

1. The identity certificate travels in the claim's JWS protected header as `x5c` (RFC 7515 §4.1.6). The resolver takes the leaf, requires a URI SAN naming the same agent and version, the current time within its validity, and a key type claims support.
2. The JWS is verified against that certificate's key before any network call. A mismatch is the registry's `(false, nil)`; a bad claim never costs DNS or log traffic.
3. `_ans-badge.<agentHost>` TXT is resolved; the badge URL must be `https`, on the configured allow-list of transparency logs, and shaped `/v1/agents/{agentId}`.
4. The log's status token is verified against pinned root keys (or, with an explicit opt-in, keys fetched from the log over TLS): same agent id, same `ans://` name structurally, a status of ACTIVE, WARNING, or DEPRECATED, and the certificate's fingerprint among the agent's attested identity certificates.
5. The log's SCITT receipt is verified and bound to the same agent and name.

The subject must be in canonical form (lowercase host, numeric version without leading zeros), so one attested identity cannot verify under several spellings.

## Hooks in identity.v1

- `InferIdentityType` returns `ans` for `ans://`; until now the prefix fell through to `dns`.
- `api/identity/v1`: `AnsSigner` (certificate in `x5c`; the claim's `certificate` field stays SPIFFE-only), `CertificateFromJWS` (bounded before decoding, documented as untrusted until the caller anchors it), `CheckSigningKey` (the one key allowlist signers and resolvers share), and `NewCertificateSignerFromFile`, which selects the SPIFFE or ANS signer from the certificate's URI SAN scheme so the CLI never switches on schemes. The key, size, and key-type checks every certificate-backed signer needs moved into a shared loader that the SPIFFE signer now uses too; both certificate signers report their expiry.
- `server/identity.NewDefaultRegistry` builds the resolver set for both binaries and appends the ANS resolver when its block is enabled.
- `Resolver.Verify`, `KeyResolver`, `verifyClaim`, the protos, and the `claims` table are unchanged.

The RFC 7515 chain-validation requirement is deliberately not applied to `x5c`: the log's fingerprint attestation is the trust decision, and the certificate is untrusted until it passes. If you prefer the claim's `certificate` field as the transport, the change is `verifyClaim`, the three resolver signatures, and the removal of `CertificateFromJWS`.

## Publisher side

```bash
dirctl identity claim --record <cid> --role identity \
  --subject ans://v1.0.0.agent.example.com --key identity.key --cert identity.crt
```

The command prints when the certificate, and with it the claim, stops verifying; after the registration authority renews the certificate the claim is pushed again. Two fixes to the command as it stood in #2125: `--key` and `--cert` were marked required together, so a plain `--key` claim failed flag validation; `--key` is now required on its own. With a certificate the signer follows the certificate's scheme.

## Operator side

`identity.ans` in both the server and the reconciler's identity task (`enabled`, `trusted_log_hosts`, `root_keys`, `allow_unpinned_root_keys`, `timeout`, `dns_server`, `ca_file`), off by default. While off, an `ans://` claim fails with `no resolver registered for subject scheme "ans"` (before this change the dns resolver failed it with a TXT lookup error; no persisted column carries the scheme, so nothing is re-indexed). Re-verification, and with it revocation pickup, needs `reconciler.identity.enabled: true`, which no shipped sample set; the reconciler samples now show it. The daemon warns at startup when the two blocks disagree or when the server has ANS on while the task is off.

Environment binding: no `identity.*` key was bound in either binary or the daemon, so `RECONCILER_IDENTITY_*` and `DIRECTORY_SERVER_IDENTITY_*` were inert unless the key appeared in the file. The reconciler now binds `identity.enabled`, `identity.interval`, and the ans keys, the server the ans keys, the daemon both prefixes; every binder reads the ans key list from the configuration package. The Helm reconciler Deployment gains `reconciler.extraVolumes`/`extraVolumeMounts` so a private log's CA can be mounted for `ca_file`; the daemon resolves a relative `ca_file` against its data directory.

Resolver behaviour under failure: a per-log circuit breaker (3 connection-level strikes, 2-minute cooldown, under the task's 5-minute default interval) keeps a dead log from stalling ingest or the task; a short outage costs one run of `failed` verdicts, the identity task's model for every scheme. Strikes accumulate across verifications and only a verification that completed every fetch against the log clears them, so a log with one hanging endpoint trips the breaker as a dead log does. Log clients and, in unpinned mode, root keys are cached per log origin; a key rotation is recovered within the same verification, and an unknown pinned key logs a WARN with the kid. Dependency failures log at WARN, per-claim verdicts at DEBUG, since the task re-verifies every claim every interval.

## Lint at the base

`server.New` and `registerTasks` failed the repository lint at #2125's head (maintainability index 19, cognitive complexity 32, nested block 6); the identity wiring of each moved into a helper, which clears both. Every other lint finding in a file this PR edits is fixed here (`api/identity/v1/{claim.go,sign.go,verify.go,sign_verify_test.go}`, `server/identity/registry.go`). Findings in files this PR does not edit are left as they are, so that the diff stays about one thing; with the repository's pinned linter at #2125's head they are: one in `api` (`identity/v1/referrer_test.go`), two in `reconciler` (the duplicate pair in `tasks/identity/task.go`), and 22 in `server` (`controller/{identity.go,identity_test.go,naming.go}`, `database/gorm/{identity_claims.go,identity_claims_test.go,record_exclude.go,record_remove_test.go}`, `types/database.go`, `ingest/{ingest.go,ingest_test.go}`, and the `did`, `dns`, `spiffe`, and `wellknown` resolver packages). `task lint` therefore stops at the server module until those are fixed; every package this PR edits lints clean on its own.

## Follow-ups

- In the identity task: keep the previous verdict on dependency failures and skip the upsert on context errors, a per-record deadline, a failure count and duration in the run summary, metrics.
- `expires_at` from the certificate for certificate-backed claims.
- Scheme constants instead of string literals; registry-owned scheme inference so a new resolver needs no `api/core/v1` edit; registry injection into the reconciler service so one composition root serves both binaries.
- ANS details (log, agent status) in `ClaimStatus`.
- A Kind e2e with a local ANS stack.
- A pull-mode relying-party verifier in the ANS SDK (`agentnameservice/ans-sdk-go`; the SDK's `AnsVerifier` verifies a connecting peer that presents its own SCITT headers) that shrinks this package to the allow-list, configuration, and result mapping.
- The dangling `#record-identity--ownership` anchors in `dir-component-records-validation.md`, and the stale `name:` sample blocks that #2126 removes.

## Testing

The pinned linter on every edited package (0 issues), `task test:unit` (83 packages), `task license`, codespell and markdown lint on the docs, `helm lint --with-subcharts` on both charts. Coverage: `server/identity/ans` 99.8%, `server/identity/ans/config` 100%, `server/identity` 100%. New tests: the signer and header helpers in `api/identity/v1` (table-driven, every key type, every rejection), the resolver package (per-stage tables through `Verify` with synthetic ES256 fixtures for every negative case and golden fixtures captured from the ANS reference implementation; the golden certificate's private key was never captured, so the golden test drives the attestation step and a signature-mismatch case), the configuration binders in all three loaders, the daemon drift warning and path resolution, the CLI loader and flag validation, the registry (dispatch by scheme, the `(false, nil)` verdict, error text kept intact, the unrouted scheme, the default set with the ans block off, on, and invalid), and the reconciler service with the identity task and its ans block. Manual end to end against the ANS demo stack is described in the trust-model doc.

Dependency: `github.com/agentnameservice/ans-sdk-go v0.1.17` (MIT) in the server module; the reconciler, cli, and tests modules carry it as an indirect requirement. `fxamacker/cbor/v2` becomes a direct server requirement through the fixture-minting test. `task deps:tidy` at #2125's head also promotes `sigstore/sigstore` to a direct requirement of the reconciler module and drops three stale indirect requirements from the tests module; both are pre-existing and included so the modules are tidy.


muscariello:
@kperry-godaddy I created a breakdown of this work to be split in different parts.  Can you review that and tell me what you think about it?

kperry-godaddy:
@muscariello - At first glance, it looks great.  I will dive into the details later today and this week. Thanks for taking the time to do this.

PR #2166, opened 2026-09-21
chore(github): add net.zenitheye/ to import records

zenithread-cpu:
Add the `net.zenitheye/` namespace to the scheduled MCP Registry import list.

ZenithEye Agent Commons is published in the official MCP Registry as:

`net.zenitheye/agent-commons` v2.0.0

I verified the current AGNTCY importer locally with `dirctl` v1.7.0. A dry run against the official MCP Registry:

- found exactly one matching ZenithEye record
- transformed it successfully to OASF 1.1
- produced zero failures
- made no public Directory changes

This allows the existing scheduled AGNTCY import workflow to ingest the ZenithEye namespace through the same mechanism already used for the other listed MCP Registry prefixes.


csirmazbendeguz:
Hi @zenithread-cpu !
Thanks, but we're not looking to import more records to our node.

